PR #5, opened 2015-04-08
refactored conceptlist metadata: extracted Conceptlist items from BibTeX...

xrotwang:
... into a tsv file


xrotwang:
@LinguList as far as I can see, there's no need anymore to keep the conceptlist metadata inside the references' BibTeX file, right? So I think it's more transparent and maintainable to keep it in a separate tsv file.

LinguList:
@xrotwang I agree on this point. While we were still in a phase of not knowing which columns might be added, it was probably useful to use bibtex (because of it's flexibility), but now, we can stick to this scheme.

PR #11, opened 2015-04-08
Removing many nans from the ALLEN-file, also removing many double-links

LinguList:
Note that the NANS in Allen-2008-500 were now all removed and new ids and definitions were added. This does, however, not mean that all links to OWIKI are also given, since some concepts, especially classifiers could not be identified. I added a questionmark to indicate missing links from OWIKI. The definitions are not particularly nice, especially for the classifiers, but they should serve their purpose, and we can also modify them later on. 

This leaves another set of ca. 70 nans in the HUBER-data and some other sets.

I also just corrected two missing items and added all relatiosn as "similar" (our default and required relation, we can discuss the name, but I like "similar" since it doesn't claim too much).


xrotwang:
As far as I'm concerned empty OMEGAWIKI columns indicate missing links to OmegaWiki alright :)
So I'd vote for removing the `?`, since this introduces just some non-specified informal markup.

LinguList:
Another important thing: I realized that our initial idea of adding relations to the links will actually lead to confusion and imposes redundancy: 

* concepticon lists all concepts for us (even apparently stupid stuff like animal/meat)
* we plan to add broader/narrower links between conceptlists and concepticon

The actual relations, however, are much better handled by adding links between the concepticon concepts, using the broader, narrower idea we had in mind! So for the first version, this would also make our goal much simpler: all we have to do is to get all lists down to defining unique concepts, no matter how strange it may be (we have lists that have "arm 1" and "arm 2" without further specification!). But if we then use "similar", "narrower", "broader" to interlilnk the concepticon links, we can handle this all in once, and do not have to bother about the individual lists.

A drawback is that we will have a certain amount of errors due to bad links at first, but in the future, we can refine these (and also hope on the community to provide additional support).

Another drawback is the representation: we would be forced to add another dimension (some transitive hull thing or whatever) to the algorithm that would allow to search for related concepts (via the relations) apart from "identity". 

For this reason, and since I think it is straightforward, I propose to remove possible relations from the individual concept lists and add an additional list with relations (maybe GML format, or JSON?) for the underlying semantic network.

cysouw:
This is a hard one, that deservers careful consideration, specifically between Mattis and me :-)

===
tl;dr

I don’t think that I like the idea of broader/narrower, because it just adds complications. 
And I think that some concepts might be left unlinked. Maybe we can use “NOT LINKED” as a special entry?

===

First: the goal of the concepticon is the linkage of lists. If a specific concept only occurs in one list (which is often the case with non-linked concepts), then they are not very interesting at the moment anyhow. 

Second: even when not linked to a CONCEPT SET, then the individual concepts still are there. They are simply not linked to anything. I still think that sometimes it makes sense to leave something simply unlinked (because it is such an idiosyncracy of some kind special list).

Third: our CONCEPT SETS are **not** supposed to define identities between concepts: they are supposed to make it easier to search across wordlists, and to combine wordlists from different sources with each other, so there is always some leeway in any links (or: every link is broader/narrower to some degree!) We already sometimes link concepts to multiple sets, which implies some kind of “narrower” already.

best
michael

LinguList:
I agree mostly, though we have a definite practical problem of linking stuff already at the moment. If we encounter lists like Leipzig-Jakarta, we have "arm/hand" which would be completely unlinked to all interesting "reflexes", like "hand" in Swadesh-100 or "arm" in Swadesh-200. So for practical purposes, we NEED the inter-conceptual layer.

Now, if we just add another file that can be ignore in most cases, we have the double advantage of

* not giving up the original agenda, while on the other hand
* answering to practicle purposes (combining leipzig-jakarta and swadesh-100, as many people have already done, everybody with slight differences)

So in contrast to the earlier idea of adding relations directly to the conceptlists, we now just add another file with a statement similar to our "see-also" column in the original version. So in my opinion, we don't loose anything here, while winning a lot.

cysouw:
Why not link “arm/hand” to both ARM and HAND?

LinguList:
Here, I have reserverations due to past experience: In the past, this lead to inconsistencies: whenever I was too lazy (or did not have enough time, since this takes time) to identify a specific omegawiki concept, I would add two links for one, like I would link "cold" and "frost" to COLD. Now, judging from the current state, this lead to errors which are avoidable if we require that all concepts are only linked to one entry (also for checking, since checking for unique relations is very general and will force every contributor to be explicit). The practice, which I would follow in the past was some "let's look at it later"-style. Now, I regret that, since I am now forced to go through the lists and have it re-checked, with some annoyhing inconsistencies (as I only found again today when looking closer at Allen's list). 

I am just afraid that at some point, our current practice will break down and we will have to do much re-editing when trying to refine it, and I have the hope that the unique-links will avoid greater problems in the future...

cysouw:
I don't get the problem, I think. Simply don't link something if you don't have the time to do so :-).

I don't like the addition of four different concepts for arrows (current many-to-one example from zgraggen). I don't think that such many-to-one links lead to similar problems. 

Proposal: we try to make as few as possible branching links. Maybe we can simply automatically extract exactly those that exist for checking. 

We can still make a list of semantic relations between our concept sets, but we can probably also do that automatically (clicks-like)

Best
M. 



-------
This Message is short because typing on an iPhone is not destined to lead to epic novels.

xrotwang:
I think it is desirable to have an injective mapping from concepts to concept sets, because I hope in the future, when people create word lists, they should list counterparts for our concept sets rather than for concepts from individual concept lists. It will also make it easier for downstream analysis tools, if all word lists can be transformed into lists referencing concept sets.

That said, for pragmatic reasons I think I could live with a special concept set called "Undefined", sort of analoguous to the language family "Unclassifiable" in Glottolog.

LinguList:
I just ran a check of the current status, and according to this check, we have about the following situation:

of 9647 concepts currently in our conceptlist collection, there are

* 189 splits, that is, concepts that link to more than one concept set, and
* 20 mergers, that is, concepts that link to the same concept set within one list

This gives good arguments for both positions: 

* the number of multi-links is not very large, given the large number of concepts, but
* it is also not difficult to correct some 200 entries to make everything unique

I see further practical advantages for unique links: if we encounter a large majority of lists which do not make an explicit distinction that we would've made initially, such as it is the case for "child" (descendant vs. young human) and some other concepts which are "important", one needs a certain expertise to make a correct linking, since not everybody will know that we have two concepts in our list, so we end up having links to either of the two concept sets (as we had in the past and still have). A further example is "nail (fingernail, etc.)", which is notoriously mis-glossed and really hard to tackle exactly: we shouldn't link "claw" with the concept sets FINGERNAIL and CLAW, but we can add a relation between FINGERNAIL and CLAW, thus conforming to what Tower of Babel now states as a common practice, namely, that CLAW is translated as FINGERNAIL in most practical wordlists. In the past, I was struggling terribly with these things, when mapping individual lists, and I really think that we can now right the wrongs by requiring uniqueness and adding additional relations...

LinguList:
Ohoh! We have another problem we need to tackle: Some lists, like ZGRAGGEN, don't have "unique IDS", since we used things like "page numbers" to create them. So in the end, the doubled lines ([two in ZGRAGGEN](https://github.com/clld/concepticon-data/blob/master/concepticondata/conceptlists/Zgraggen-1980-380.tsv#L20)), appear to be "mergers" although they are indeed to be kept distinct! I'll try to identify these cases ASAP and refine the IDs.

LinguList:
allright, it is even more complicated, since our source "Zgraggen1980" is actually not identical with the list we linked (there's only one arrow in the index!), and teh page-numbers also do not match up...

xrotwang:
From my point of view, renaming Defined Meanings to Concept sets does also help conceptually:

- Each concept is in one concept set.
- A concept set may have no definition - meaning it's "undefinable", and thus probably not useful for comparison.
- Concept sets with just one member concept may be either useful as candidates when creating a new concept list, or a hint that the member concept is not really common.

So now, rather than having a special UNDEFINABLE concept set, I'd vote for multiple concept sets without definitions.

LinguList:
Okay, last commit for this evening. I now refined the concept lists in such a way that I tried to remove multi-links in some exemplary lists and I added further concept sets. What is important in this context is that it really seems to be useful to re-check the data. I found many cases where it just makes more sense to add an additional concept (which is now easier, that we don't need to look for owiki-stuff all the time), and have a more precise mapping. 

I also added an experimental list of relations.

I propose the following right now: I'll keep refining the lists, adding relations, and adding new concept sets. If, in the end, it turns out to be too great a loss if we separate narrow concepts (like "arm/hand" from "arm" and "hand"), we can always get back to normal (as proposed by @cysouw) by simply re-defining the concepts on the basis of the proposed relations (or even re-winding our commits). There are still some 149 mult-links in the data, which I will check during the next days on a list-per-list-basis, depending on my mood. It is anyway good for cross-checking since I found quite a few cases which needed refinement, so I hope everybody agrees with this procedure (at least for testing purposes).

cysouw:
Just one short note, too much going on right now :-)

Please don't forget, that when two different people write "snow" in their lists, this does not mean that they mean the same thing. 

Could we use UNLINKED instead of UNDEFINED ? Most unlinked concepts are very well defineable!

Best
Michael



-------
This Message is short because typing on an iPhone is not destined to lead to epic novels.

LinguList:
I agree on this, unlinked is okay.

BTW: the problem of "snow" is exactly what I have been refining recently, if people write  "man/male", for example, they mean "male person" (as opposed to "male adult person", or "male young person"). So I added new concept sets here for accuracy. We have an additional problem, though: if compilers write "rain", without further specification, we don't know whether they mean "RAINING" or the "RAIN". Sometimes, their wordlist indicate the confusion (look at IELEX, where both nouns and verbs are used), sometimes, wordlists allow to reconstruct the compiler's intention. What is a fact here, is, that we have an additional level of "epistemology" here between the compilers and our interpretation, and we can't really avoid it. It's not only that the compilers don't mean the same, but also that we don't know what they mean (at least at times).

LinguList:
If we use the "unlinked" class, we could just assign it to the concept set "0", which is still free and also a special number. This would allow us just to remove all of the annoying NANS for the moment and ignore them for some time. It would also allow us to get rid of those cases where we have multiple links to the same concept set in one list. we would then link all these annoying cases (like ARROW-1 ARROW-2, etc.) to 0.

BTW: When dealing with lists like the one by Dixon, we need to either re-arrange the list, or link multiple times to the same concept set, since it seems really rather not useful to add man-1 and man-2, just because Dixon and Kroeber put it into different etymological classes (no semantic decision here!). 

But re-arranging the list is probably easier, or should we allow to link multiple concepts from the same list to one concept set, but disallow to link one concept to multiple concept sets?

xrotwang:
I would not allow linking multiple concepts from one list to the same concept set. When a concept list distinguishes between two concepts, we should provide concept sets for each, and let the future decide whether any of these turns out to be useful, in the sense of turning up in other concept lists. And again, as a reason, I would stress the Glottolog analogy: Glottolog aims to provide identifiers for any languoid a linguist may want to distinguish, even if we can't fit it in the language classification. Simply as a reference point, so that others can make it clear when they talk about the same thing.

Now it could be argued that the single concepts, e.g. http://concepticon.clld.org/values/Allen-2007-500-100 , can already serve this purpose. But I think this would be too confusing, so no one would understand the distinction and use the identifiers as intended. My main use case for the concepticon is minting [identifiers for concept sets which can be used in interchange formats](https://github.com/glottobank/cldf/blob/master/wordlist.md#the-data-file), after all.

cysouw:
I still don't see the problem, mattis! I think you are trying to be too precise, although precision is impossible with this mapping (because meaning is inherently fuzzy). 

I think we should be modest on our goals: we want to make a mapping that help combine wordlists from different sources. Now: wordlists have an inherent problem of fuzziness, which we won't solve. So, most wordlists include various words for a specific concept. When we link various concepts from a single list to one concept set, that just means that a few more words will end up together in one concept. Not a big deal I think (as that is the problem of the whole enterprise)

Eg, when I am interested in the word for arrow, I will probably want to compare it to all four different arrows from zgraggen. Only within the zgraggen data it is important to distinguish them. So for cross linking, all four can just be mapped to one set ARROW. 

I still think that we should go for one curated list of concept sets and just try to do our best to link individual concepts. 

And:

- idiosyncratic subdivisions that only occur in one list (like the four arrows) get a many to one linking
- idiosyncratic conflations get a one to many linking
- often the compilers of the wordlists themselves were confused, like with rain, or burn, or marry, etc. such (actually of course all!!!!) English terms are difficult, because they have multiple meanings. We cannot solve this problem, we can only try to *reduce the complexity*. 

So my plea: keep it simple, at he cost of some accuracy. 

Michael


-------
This Message is short because typing on an iPhone is not destined to lead to epic novels.

LinguList:
As a compromise, I would vote for:

* no fuzzy linking from concept to multiple concept sets
* allow multiple linking from within one concept list to multiple concept sets

This seems to reflect practical needs at its best, since we have cases of complete identity (as I just mentioned, since Dixon's list is not a conceptlist, but a cognate-set list!) in the lists, and disallowing fuzzy linking in the other direction facilitates representation and encourages linkers to give it a very good thought before they start random linking.

But in the end it is also a question of:

* who wants to use the lists **how** (we all three have different ideas here, I guess), and
* who is going to link the lists in practice (student assistants can only do part of the work, as my past experience shows, so it will go down to us experts, and to me in concrete), and
* what is the best compromise to keep everybody satisfied.

My problems and my precautions come from practical experience with list-linking: Since I am about to link many more lists in the future (every future database-project should link the lists, and I am involved in quite a few), I need a workflow that minimizes the error-rate and is as accurate as possible. 

The question is, Michael, whether you loose anything if I resolve the remaining cases (about 200 now) and add my little concept-network as an independent toy that can be seen as independent from the concepticon? If not, we could just give it this test-run now and see what future practice brings (especially interesting, when other people start to contribute linkings), given that one-to-one can easy become many-to-many, but not the other way round?

cysouw:
(as I said earlier, this is a hard one that deserves detailed discussion, so thank you for engaging ! It indeed boils down to the question what people want to do with the lists.)

I think with ‘fuzzy linking’ you mean “one to many”, isn’t it? That isn’t fuzzy at all IMHO :-).

So, to rephrase your proposal:

- we would force everybody to link each CONCEPT to a single CONCEPT SET (so no one-to-many)
- basically: try to link to the nearest sensible set, and if that really sound wrong, then make a new idiosyncratic set, or leave it unlinked. So if there is a concept RAIN, then make a decision to either link it to the noun or to the verb (trying to interpret the source).
- we allow for many-to-one linkage in cases in which concepts are very specific, often because of lexicalisations of a specific area (e.g. the four different kinds of arrows in Zgraggen).

@mattis: I think your ‘relation-network’ between the sets is interesting in any case! If you want to do that, that is simply a very good idea, so please yes, do that! It just adds something, as you say. I have nothing against that. As I said before: we can also try to induce such relations from the data, like in the clicks-project.

so yes, go ahead :-).

There are still various problems, but these cannot be solved easily by principle

As I see it, the main problem is, when in the future there is a decision to split a concept set, e.g. MARRY. Suppose at some point we decide to have two different concepts of MARRY (or four, which would semantically make really a lot of sense), then we will have to relink everything that is linked to MARRY. I see no way around this problem, also not when using a more complex structure.

Or even more problematic, if someone want to add a concept set, then it might be that various earlier links are better matched to this new concept (this is actually why I think the current problems arose most of the time).

best
michael




————————
Michael Cysouw

Auerfeldstrasse 20
D-81541 München

home: +49 (0) 89 4805 8805
mobile: +49 (0) 160 94 777 076
email: cysouw@mac.com
web: www.cysouw.de/home
————————

LinguList:
Now in cases as "MARRY", this nicely illustrates why it is easier to go for unique links, at least in one direction, since it allows us to NOT relink through all lists (provided the links are still acceptable):

* one list has "marry", we link it with the broadest version, which has a definition like "To go into marriage with somebody or to conduct the marriage ceremony for somebody."

Then, suppose a new list pops up, which re-defines "marry" as transitive, we then link it to:

* MARRY (SOMEONE): "to go into marriage with somebody, and note the relation: MARRY broaderThan MARRY (SOMEONE)

then, we encounter yet another list that shows "take a wife", then we would:

* MARRY (TAKE WIFE): to go into marriage with a woman", and the relation MARRY (TAKE WIFE) narrowerThan MARRY (SOMEONE)

In this approach, we only need ot add a link and a new concept. In the approach with multiple links, we need to re-define everything in all list, although it would make sense to initially start to link 

"marry somebody" in a Germanic wordlist with MARRY (TAKE WIFE) and MARRY (TAKE MAN)

So this is the point, which I see as the strongest argumetn for the relations.tsv that I have been starting to add:

**We do not necessarily assume compositionality of concepts, as we would if we allow for multiple links, and we can minimize the amount of changes made to the data, provided everything was encoded consistently in the beginning**

And marry is not the only case we have. There are many more (NAIL OF FOOT AND FINGER, NAIL OF FINGER, NAIL OF ANIMAL, NAIL OF HUMAN OR ANIMAL). If we go for multi-links (one concept to multiple concept sets), it will be very difficult to be consistent in the decisions here. Using unique links from concept to concept set, it will be much easier to handle it (and it is much easier to get back to the previous state in case we really find it impracticle in the future). 

With this last part of the commit, the main job is done, and not only have multilinks been removed, many entries have also been re-checked and refined. So it was definitely worth to got through it again.

The next step (to be done in the next week) is to try and resolve some of the multiple links (there are certainly also errors, as I found when checking for the links during this week), but more importantly, to re-check all links, and try to remove or reduce the number of NANs.

cysouw:
@LinguList : are you waiting for me to merge the branches? I just noticed that they are not yet merged. As I said earlier: please don't wait for me with merging - I might just hop in to discuss and work on it occasionally. I will now merge this.

PR #12, opened 2015-04-21
Removed all NANs from current list assembly and corrected many entries.

LinguList:
We now have exactly 50 lists in the concepticon and no obvious errors, as far as I can see. We also have an underlying network structure that can be resolved down to allow for multiple links just as in the versions before (e.g., arm/hand -> arm AND hand). Note the following:

* network structure is given as simple tsv-file in a directed graph. It shows different relations between the concepts, some of them being experimental, but four being important, namely "similar" (means concepts are very close), "sameas" (means concepts are actually identical, identity confirmed by Wikipedia or other definitions, and the concept with the higher ID can actually be removed, since they are not being used in any of the concept lists), "broader" (A broader B, A covers a larger semantic space than B), "narrower" (B covers a narrower semantic space than A).
* Many NANs were resolved, but I left some concepts unlinked (now indicated by empty fields rather than by keywords and the like)
* Most cases where one list links two "concepts" to the same conceptset have been resolved, some have been retained (e.g., "slow", "slowly", and some others)
* I tried to maintain consistency in many complicated cases ("warm" vs. "hot" vs. "warm/hot", also in the light of the translations in other languages), but there may still be cases where improvement is needed, and even erroneous links.


xrotwang:
Ok. Let's see whether I can get the clld app up and running this afternoon :)

LinguList:
I hope there are no bad-ass errors which I overlooked. The scripts I used have a checking mechanism, but I don't know whether it's fully exhaustive. Let's hope it works...

And regarding the concepticon.tsv: I forgot to mention that removed the WOLD links and teh "seealso", since these are no longer needed.

xrotwang:
Ok. Import was easy enough. I noticed some things, though: 

- Your usage of *broader* and *narrower* is in the opposite direction of the usage in [SKOS](http://www.w3.org/TR/skos-reference/#semantic-relations), i.e. yours derives from *SOURCE* *narrower than* *TARGET*, as opposed to *SOURCE* *has narrower concept* *TARGET*, as in SKOS. While this is probably ok, it does introduce some cognitive dissonance (at least for people with SKOS-like intuitions).
- *DRIP* appeared twice. I added a disambiguation number, but maybe something like *(transitive)* may be even better.
- I think I wouldn't add both directions of inverse relations to `conceptrelations.tsv`.

PR #14, opened 2015-05-15
Adding a couple of new concept lists and more concepts and the like.

LinguList:
I added a couple of more lists, some quite interesting (one used in sign language, then also the wordnet mapping, where I added further links to the Open multilingual wordnet project). 

All in all, we are getting somewhere with the lists, but refinement is, of course still in order.

One question remains for now: The Wordnet link by Lars Borin contains some meta-information given in comment lines, along with "how to cite" and the like. The question is: can we simply convert this to our format, as I just did, or should we retain the comments? We quote the source as requested, anyway, and I modified the source by adding more links. But I am still not sure whether this practice is correct...


LinguList:
Merging this now, since I think it is more or less OK. If it goes on in this paste, we may have some nice 70 lists assembled when launching the first official 1.0 release.

PR #20, opened 2015-08-10
Refined data and added more unlinked resources

LinguList:
The unlinked resources may be interesting, depending how quickly they can be added, we should decide which ones to map for the next release.


xrotwang:
`helpers/.temporary1.tsv` should probably be removed.

xrotwang:
It's almost scary how many of these lists there are :)

LinguList:
It definitely is! And this is only the top of the iceberg, given all the
lists which we found mentioned in the books but which we lack the scans
(as some in the current issue tracker)...

And if one then thinks of those lists which are simple collections of
data for specific languae families, or field work, and the like...

PR #25, opened 2015-09-25
some proposed corrections

xrotwang:


lrec/concepticon.tex
```diff
@@ -33,8 +33,8 @@
 We present an attempt to link the large amount of different concept lists (aka ``Swadesh lists'')
 which are used in the linguistic literature. This resource, the \emph{Concepticon}
 (\url{http://concepticon.clld.org}), links
-\conceptnumber concepts labels from \listnumber conceptlists to \setnumber concept sets. Each
-concept set is given a unique numerical identifier, a unique label, and a human-readable definition.
+\conceptnumber concept labels from \listnumber conceptlists to \setnumber concept sets. Each
```
xrotwang: I would rather stress that we assign URLs to each of the relevant resources in the Concepticon data model, thus integrating Concepticon in the Semantic Web.

PR #30, opened 2015-10-06
completed section about clldclient, added section about dictionaria

xrotwang:


LinguList:
Nice. I'll try and have the conclusion added during the next days.

Issue #31, opened 2015-10-16
Conceptlist from Grollemund et al. 2015


xrotwang:
In [Bantu expansion shows that habitat alters the route and pace of human dispersals](http://www.pnas.org/content/early/2015/09/09/1503793112.abstract) Grollemund et al. use wordlists for ~420 Bantu Languages containing words for 100 concepts, sampled from 159 concepts in 

> Hombert J-M, Van der Veen L, Medjo Mve P (2011) ALGAB, Atlas Linguistique du GABon (Laboratoire Dynamique du Langage, Lyon). 
> Available at http://www.ddl.ish-lyon.cnrs.fr/equipes/index.asp?Langue=FR&Equipe=8&Page=Action&ActionNum=48
> Accessed November 10, 2014.

LinguList:
Is already mapped, will be pushed with my next PR, or should I PR right now?

xrotwang:
Would be nice to have right now, since I'm working with the data.

PR #33, opened 2015-10-16
Added two more lists: grollemund (bantu) and the Uto-Aztekan

LinguList:


xrotwang:
merci. Should we add the 159 item list of the Atlas Linguistique du GABon (see #31 ) as well?

LinguList:
yes, I wasn't aware of it! Just will take until after Jena, but it's good for v-1.0, we may even be able to make it for 120 lists (I want to have number that gives zero for % 10 ;-)

PR #38, opened 2015-11-02
New structure proposed for conceptlists.tsv

LinguList:
Adding two more columns, one containing concept list types (currently 12 types, abbreviations see below), and one containing status on pdf resources. PDF and REFS is now separated, adressing the problem that we do not have pdfs for all sources and also do not necessarily want to have them.

Here is the list of characterizations I propose for concept lists (also added in the readme), currently adressed by an abbreviation, but we can spell it out verbosely:

* an: **annotated concept list**, a concept list that contains any level of annotation added to the concepts it contains. Typical example are those lists which additionally order the concepts into semantic fields, like "Bowern-208-207", or add colexification information, like "List-2015-1280". 
* av: **areal vocabulary list**, a concept list that was designed for the purpose of subgrouping in a specific area, like "Matisoff-1978-200", or "Norman-2003-40"
* bv: **basic vocabulary list**, the most typical case of a conceptlist.
* h: **historical concept list**, a list which we included for historical reasons, and where due to the historical character of the concept list, the reasons for the compilation are not necessarily clear
* hi: **HiHi vocabulary list**, following the terminology of [McMahon and McMahon (2005)](http://bibliography.lingpy.org?key=McMahon2005), this is the most stable part of a given basic vocabulary list.
* lo: **LoLo vocabulary list**, following the terminology of [McMahon and McMahon (2005)](http://bibliography.lingpy.org?key=McMahon2005), this is the least stable part of a given basic vocabulary list.
* ls: **Less stable sublist**, in contrast to the practice by [McMahon and McMahon (2005)](http://bibliography.lingpy.org?key=McMahon2005), wo extract two small lists from a baselist, which do not cover the whole baselist, less stable sublists are based on a complete division of a given wordlist in two parts. Paradigmatic examples are "Yakhhontov-1991-100", which is divided into "Yakhontov-1991-35" and "Yakhontov-1991-65". 
* ms: **more stable sublist**, the counterpart of a **less stable sublist**, that this the upper division of a given basic vocabulary list, which is supposed to be more stable than the lower division.
* q: **questionnaire**, a term which is reserved for all lists which have a larger amount of lexical items and have been created for specific purposes other than lexicocstatistical subgrouping or cognate detection. Questionnaires are often typical for a given area or language family.
* rank: **ranked concept list**, a list which can be ranked according to some meta-data that the list provides, like "Tadmor-2008-100", or "Starostin-2007-110".
* sp: **special purpose list**, a dummy category to include all lists which can currently not be sufficiently classified. 
* us: **ultra stable concept list**, a list that was compiled with the idea in mind to list concepts whose form-meaning relation is supposed to be very stable.

More than one type should be possible to add to a conceptlist, since some are multi-type. In these cases, comma-separation will be used.

I also modified the test-script by removing print-statements and adding them to errors and accounting for the new pdf-field to check only there for the explicit matching between pdf and source.


xrotwang:
Considering that these *types* describe various aspects of a concept list and that one concept list may have several such *types*, I'd rather call them *tags*.

LinguList:
Ah, I see, I agree. I'll just create and upload a revised list where the tags are also spelled out, so that it may be easier to read.

LinguList:
Okay, with the tag system, this is much nicer, since there's a bit more freedom also to tag the lists, which may be quite difficult. Note that the current tagging is not necessarily complete now, but just an initial attempt and proof of concept. I think it is a definitely useful feature, since often people also do not know for what purpose other people create their concept lists...

PR #39, opened 2015-11-03
fixed concept lists to allow import into concepticon app

xrotwang:
On import, concept lists are read as list of `namedtuple`s, thus, header names must be valid [python names](https://docs.python.org/2/reference/lexical_analysis.html#identifiers).


LinguList:
Ah, I see. So maybe we should add this already as a check when testing the data? Here, we could also add the test for "gloss" versus "English".

xrotwang:
@LinguList yes, I didn't add such tests, but will do so now.

PR #44, opened 2015-11-17
First test for the meta-data organization in the concepticon.

LinguList:
I added five more lists (including bowern's list, which was difficult, since it is very unspecific and contains even two glosses, one for foot/leg and one for foot alone!)

structure for metadata is now as follows:

* folder concept_set_meta: contains tsv-files with concepticon_id in first and some further ID in second column, potentially even more, but for now, we only have those two types in the examples
* file concept_set.tsv contains the information regarding the metadata. Here, I added various things, as one can see from the header, but one important thing is the link-category, where I used python-formating-syntax to specify how the element of the list can be used to create a direct link: e.g., for wiki, I have: http://en.wikipedia.org/wiki/{wiki_id}, and what I list in wiki_id in the column are the names of the wikipedia website, like "hand", "climate", etc.

I don't know whether it's useful in this form, so feedback is welcome, but for the workflow, it should be fine to work like this, I guess.


xrotwang:
Regarding the formatting of links, I'd recommend we use URI templates, following the recommendation in the [Metadata Vocabulary for Tabular Data](http://www.w3.org/TR/2015/PR-tabular-metadata-20151117/#dfn-uri-template-properties). There's a [python package](https://pypi.python.org/pypi/uritemplate) to work with URI templates.

xrotwang:
Actually, we might want to follow the Metadata Vocabulary for Tabular Data recommendations fully, i.e. specify metadata about the csv files in `concept_set_meta` by accompanying JSON-LD files. The link would then be expressed as [aboutUrl property](http://www.w3.org/TR/2015/PR-tabular-metadata-20151117/#cell-aboutUrl).

LinguList:
Yes, this looks much more consistent.

xrotwang:
The two files `concepticon.new.tsv` and `conceptlistsx.tsv` should not be commited to master, right?

LinguList:
@xrotwang , can I leave it to you to adjust the format for the two examples, with the specifications you mentioned? I don't really understand completely, how the json-specs should look...

xrotwang:
@LinguList yes, will do.

LinguList:
thanks!

PR #48, opened 2015-11-18
refactored concept_set_meta

xrotwang:
As long as the only metadata we collect for concept sets are identifiers/URLs of related reources, this way of describing the metadata may seem overly verbose. But I can imagine, that at some point the various metadata files will diverge, e.g. we may want to add the actual english definitions from OmegaWiki to the `omegawiki.tsv`; at this point, descriptions in JSON-LD will prove useful for their flexibility. At some point we may even see tool support, e.g. validating the foreign keys in these files.


LinguList:
I agree that it can get much more complex in the future, so verbosity is a good way to be prepared for all things that expect us out there in the dark and cloudy world of meaning, concepts, and semantics...

PR #49, opened 2015-11-19
Make sure concept set categories come from controlled lists

### LINKED ISSUE #45: Tests  the Concepticon
When working on the concepticon, more things come to mind, which we need tests for, and I suppose we summarize them here, to not forget about this.

* tests for concepticon.tsv should check for the number of categories, like person/thing, but also for the rough semantic field (animals, the body), etc., since spelling differences may create errors here, and the categories and fields are actually fixed

Of course, more tests are surely needed, but this is what I can think about at the moment.

xrotwang:
The current variety of terms in the categories SEMANTICFIELD and ONTOLOGICAL_CATEGORY
has been narrowed down by removing spelling variants, etc. The resulting sets are now
hard coded in `concepticondata/data.py` form where they are imported when running
the integirty checks for the data.

closes #45


LinguList:
Nice, many thanks!

PR #54, opened 2015-11-23
120 lists now, and some code adjustments, as well as adjustments in the concept relations

LinguList:
I realized I made errors in concept relations, confusing broader and narrower (usage is counter-intuitive for me, and I forgot to stick to it).

Now I went through all concept relations and cleaned them up.

I also added 5 more lists, so we have 20 now, although 4 are from the same publication (people are generous in creating concept lists).

Further adjustments to the code, since the code for mapping lists had some bugs after removing the omegawiki-stuff to the meta-data-collection.

At some point, I need to thoroughly understand the rationale behind the concept relations (as I said, ti's counterintuitive for me). But that's more my personal problem, since I just need to read the relevant posts where this is discussed.


LinguList:
I take the freedom to merge this right away, given that the tests run fine (I want to see the badges...).

xrotwang:
To see the badges you need to run [the `conceptliststats` command](https://github.com/clld/concepticon-data/blob/master/setup.py#L29) - this isn't done by travis - and then commit and push the updated README.

LinguList:
I just realized, and I saw them happily on my fork. I was a bit naive...

xrotwang:
I don't know if travis can be made to update files in the repos. But it would seem a bit weird if a tool that is meant to test a certain status of the repos can change this status.

LinguList:
yes, actually you're right. I really thought that travis was doing this.
It's so stupid I feel like little red-hat now, after she saw
grand-mothers teeth and was confronted with reality...

But it's cool with the badges anyway, and they are a very nice way to
check where we are with the data.

PR #58, opened 2015-12-02
added the concept lists for the datasets discussed in lexibank

LinguList:
they are suprisingly almost all very clear forks from blust's list, but they are still slightly different, since they may add concepts, or write them differently.

At some point, we need to discuss the policy of how to handle lists which are completely identical. I would treat the list by Dyen 1997 as that, since it is completely identical with Swadesh 1952, but one could of course discuss, whether it is useful to include a list that is to 99 percent identical but does not list any new meta-data (ranks, etc.)...


xrotwang:
I'd say if concept lists are identical, and the second authors do recognize this, then the second list could just be a piece of metadata for the first one. On the other hand, the same concept list being put forward by different authors does add information in itself.

LinguList:
Well, and then we have the situation where people SAY they use the list of X, but in fact, they don't do it. In the case of Lee's list of 211 items, this is the case: they misspell lightning into lightening, and they write ash instead of ashes, and they use 'at' two times for different words (which was a bit strange for me, but maybe some error in their conversion to the PDF format). For purposes of mapping the data afterwards (adjusting it for lexibank, or other use-cases), this is something that needs correction, but it is of course an extremely minor case.
Then there are also the cases where scholars REDUCE items from a given list, since they don't find them in their languages. These cases are tricky, since they again won't necessarily follow all concepts as they were defined in the list which is claimed. In the list of Semitic languages, for example, they explicitly write "fly (n.)", although fly always is the verb in all lists by Swadesh. So here we have one divergence, and 5 deletions of Swadesh's 100-item list (but probably more, I didn't compare all cases). 
I guess, the case of Lee-2011-211 is an extreme borderliner, since the lists are really almost completely identical. The case of Kitchen-2009-95 is more interesting, also to track the fuzziness that is introduced in semantics and the current glossing practice...

xrotwang:
Given all these cases, it seems as if it is never really the case, that two lists are in fact identical :)

LinguList:
yes, it's sad, but it's right. Unless the lists will explicitly link to the concepticon, because then, they can use whatever concept labels they want: they make clear which concepts they want to talk about in the end by mapping it.

PR #97, opened 2016-02-21
Fixing most of the remaining issues for Milestone release of data

LinguList:
This PR fixes most of the remaining issues for data release version 1.0. It's not yet perfect, but I think it is a good start. I also managed to add almost 700 babelnet mappings, which is useful for the illustration of metadata possibilities.


xrotwang:
Hey, good job.

LinguList:
Guess, work will never end on this ;-)
Only just an hour ago I found that instances of "you (sg)" were wrongly mapped with "you (pl)". And the complex of "flesh" and "meat" is also awful conceptually, but I think that in terms of data, it is okay for a first release. BTW, you may check the additional statistics at https://github.com/clld/concepticon-data/blob/master/concepticondata/README.md, which are interesting in terms of concept fuzziness and concept coverage. I added a function to "commands.py" to generate these.

PR #102, opened 2016-03-04
Final PR for V-1.0 Release

LinguList:
There is of course still a lot of things to be made, but at some point, one has to stop, so I'm handing in my final PR for release v-1.0.

Many things have still been changed. I corrected lists, like the one by Borin, where columns were wrongly arranged. I corrected multiple instances where "you" was linking to "you (plural)", although it was intended to mean "thou", and I refreshed the meta-data, now providing coverage also for the MRC psycholinguistic database.

I deleted the lrec-data and added the current draft as pdf in a folder "description" instead, since it makes more sense to not have latex-files for a conference stuff there in v1.0.

Finally, I introduced a new practice for metadata, namely, leaving slots empty where one cannot find a datum, as it happens at times with BabelNet, where our concept set definitions are too strict, so that one doesn't find a correspondence.


LinguList:
I took the liberty of merging this now, since checks have passed, and I need to submit these days. I'll try and create the v1.0 release right now.

PR #114, opened 2016-04-21
Correct line 151

blag:
Github would not render this file because line 151 has a different number of columns than the others. Appending a tab (column) should fix this.


LinguList:
good point, this also shows the spelling errors in that very line.

> In our representatin, these symbols are preseverd, but were placed in an additional column labelled SWADESH.

should become:

> In our representation, these symbols are preserved, but were placed in an additional column labelled SWADESH.

LinguList:
thanks for notifying!

blag:
No worries, thanks for all of your work - I'm using glottolog-data in another project of mine. :)

xrotwang:
@blag good to hear! We switched to a new repository to curate the glottolog data, btw.: clld/glottolog - also changing the layout of the data, for example the language data is now stored in ini-files in a folder hierarchy mirroring the language classification https://github.com/clld/glottolog/blob/master/languoids/README.md
We will introduce a couple of tools to manage the data in this repos over the summer, and describe in more detail how we envision collaborative curation of it.

blag:
Thanks for letting me know! I'm parsing [languoids/languages.yaml](/clld/glottolog-data/blob/master/languoids/languages.yaml) at the moment, but I'll keep an eye on that.

PR #124, opened 2016-05-06
Last commit for 1.0.1

LinguList:
not many things changed, just updated the statistics, added Kessler (need it for benchmark database, the list), added also Kuperman's et al. data on age of acquisition, after I was told that it could be used despite the no-derivates-CC-license, note that this data is automatically mapped, so it is also indicated in the meta-data description.


LinguList:
my apologies for using jabref for the references, but it was essential to have some tool that checks also the general bibtex consistency, and this actually even helped to detect an error in the file (yet it makes it difficult to comprehend the diffs)

PR #138, opened 2016-05-11
Errata

LinguList:
The following errata are modified here:

* bad links of "you" as concept label with "you (plural") as concept set (should be "thou" instead) in lists of Galucio and Wichmann
* Kassian had a line with missing tab-stops which is now corrected"
* identifier of "Galucio" was still Galicio due to a former misspelling of the name


xrotwang:
I think changing identifiers is somewhat problematic, because it will mean new URLs, thus breaking changes of the external API. The `clld` framework has [a mechanism](https://github.com/clld/clld/blob/master/clld/web/app.py#L301-L307) to deal with issues like this, but we'd need to keep track of changed identifiers here to make sure the replacements can be imported into the web app database. So I'd propose a JSON file `concepticondata/retired.json` with the following format:
```python
{
    "Concept": [
        {
            "id": "Galicio-215-1", 
            "replacement": "Galucio-215-1",
            "comment": "corrected typo"
        },
        ...
    ]
}
```

This could be extended in case we need to retire identifiers for other objects, like `Conceptlist`s. And setting `"replacement": null` could signal complete removal of an object.

LinguList:
Okay, this should be straightforward for me to do.

One question remains though: if we re-link, as I needed to do for those stupid and annoying cases of "you/thou", is this a regular change, or is this also something more major?

It is clear that it will need to be done, and the different versions of the concepticon-data will probably usually contain adjustments of this kind. This would mean that users who rely on a list would need to check with different versions. 

I'd argue that this is just unavoidable, and a normal case, or should we also use some way to handle it more explicitly?

xrotwang:
Ah, the error in Kassian is unfortunate, because [I already tried to fix it - but failed](https://github.com/clld/concepticon-data/commit/8bb9d65e8c8e3d1e934a234132cc2271d68f953c#diff-9f36b867bf4952e60c798d68eaad94c8). Unfortunately it leads to one unlinked concept: http://concepticon.clld.org/values/Kassian-2010-116-98

LinguList:
Yep, this was actually my error initially, since I did not align the columns in the right column. This file woas difficult to annotate due to the many different full-text columns...

xrotwang:
@LinguList Can you add the file listing the changes in identifiers or should I?

LinguList:
I'm just preparing this.

xrotwang:
@LinguList regarding re-linking: I'd say this is one of the kinds of changes which lead to a new minor version of the data. I.e. after merging this PR, we can't release 1.0.3 anymore, but have to go to 1.1.

Other changes in this category would be
- adding concept lists
- possibly also adding metadata

xrotwang:
So we may encounter situations where we want to fix minor bugs in the data, but have already committed changes requiring a new minor version. If we want to be able to still apply "hot fixes", we'd have to apply individual commits to the previous released data. Anyway, it would be possible, but somewhat complicated and shouldn't happen too often in particular if we release minor version every couple of months.

LinguList:
Okay, this sounds reasonable.

I guess anyway, that for users, it is more important to have their own data mapped to the concepticon, rather than having the concepticon showing their data. And if users rely on the concpeticon providing meta-data for their lists, and re-mapping happens, users will need to sync their data with new versions, I guess just as people will have to keep an eye on glottolog, if things like the classification or the coordinates of a language are adjusted, or even the mapping between iso and glottocode changes. 

But making new X.Y releases for this will help to make this clear.

xrotwang:
Yes, you can't have it both ways - always the most up-to-date data **and** no adjustments on your end.

xrotwang:
Hm. Adding `clld` as a dependency may have been too heavy-weight - the number of libs to install make the travis build somewhat flakey.

LinguList:
given that we have the requirement that stuff is inside ONE line in the
bibtex, a workaround with pure regular expressions (or even simpler, a
parser that takes each line, opens a stack when encountering an entry,
etc.) should also likewise be possible. This would be a very stupid
bib-tex-parser which requires some way of formatting, but could then
quickly check for the three requirements, like title, author/editor, and
year. I guess we need no more than 20 lines for that code, and it would
spare us one of the reasons, why CLLD is a dependency...

xrotwang:
Well, the solution I had in mind - and already described in RELEASING.md - is to use bibtool to normalize the bibtex file, before pushing. This may also alert us to other problems.

xrotwang:
I could reproduce the travis build error locally, though. It seems that the pyparsing package it tries to install actually is broken. So maybe we just have to wait for an update upstream ...

xrotwang:
The issue was already reported. Meanwhile, locking the version of pyparsing did the trick for us.

LinguList:
> Well, the solution I had in mind - and already described in RELEASING.md - is to use bibtool to normalize the bibtex file, before pushing. This may also alert us to other problems.

Sorry, I missed that somehow, but now I see. This is very nice, and I can even use those converters to adjust my biblatex dialect to BibTex.

PR #180, opened 2016-09-08
Aliases for concept lists

### LINKED ISSUE #154: Add common names for concept lists
Concept lists should be tagged with common names like "Leipzig-Jakarta", "WOLD" or "IDS".

xrotwang:
closes #154 - not all possible aliases have been filled in, though.


LinguList:
okay, I'll have a look and add them later (they have some for the lists of SIL related projects, etc.)

Good idea in general, as I know that it caused confusion among people who think in terms of "Swadesh list" and "Leipzig Jakarta", not knowing that they can be attributed to an author...

PR #187, opened 2016-09-15
added Aslian wordlist

evoling:


evoling:
I see what's wrong with my first attempt. I'll reopen once I think it's fixed.

PR #193, opened 2016-09-21
Adding Daniels 2016

SimonGreenhill:
One thing to note is that Don's numbering follows the Swadesh 100, but the missing items are left out. I've kept the numbering following the paper, which means that the numbers in the tsv file are not sequential.

Also, the addition to references.bib has a unicode character (LATIN SMALL LETTER BARRED i) in the language name. Wasn't sure if you were going for unicode or LaTeX.


LinguList:
> Also, the addition to references.bib has a unicode character (LATIN SMALL LETTER BARRED i) in the language name. Wasn't sure if you were going for unicode or LaTeX.

We go for Unicode in general, so it's all right with this!

concepticondata/conceptlists.tsv
```diff
@@ -167,3 +167,4 @@ Syrjaenen-2013-226	Syrjänen, Kaj and Honkola, Terhi and Korhonen, Kalle and Leh
...
 Dunn-2011-146	Dunn, Michael and Burenhult, Niclas and Kruspe, Nicole and Tufvesson, Sylvia and Becker, Neele 	2011		146	areal	English	Aslian	http://www.jbe-platform.com/content/journals/10.1075/dia.28.3.01dun	Dunn2011		This list was created by Niclas Burenhult and Nicole Kruspe in order to document the Aslian languages of the highlands of the Malay Peninsula. This is based on the Swadesh list, but also includes some relevant culture/technology words.	291-323	
+Daniels-2016-85	Daniels, Don	2016		85	basic	English	Magi,Aisi	http://dx.doi.org/10.1353/ol.2016.0004	Daniels2016	Daniels2016	This is a Swadesh 100 list used to document two languages of New Guinea	219-221
```
LinguList: I would suggest to re-name the description with additional link:

```text
This is a derived version of [Swadesh's list of 100 items (1955)](:ref:Swadesh-1955-100) used to document two languages of New Guinea.
```

Instead of labelling only as basic, I would suggest to label as "basic,areal", as I assume the missing concepts result from areal patterns in NG.

SimonGreenhill:
Ok, here you go

PR #195, opened 2016-09-23
Miedema and Welling 1985

SimonGreenhill:
This article contains a 40 item list and a Holle list for a number of languages in Irian Jaya. I have ignored the Holle list for now (~1500 items).


concepticondata/conceptlists.tsv
```diff
@@ -168,3 +168,5 @@ Wheeler-2015-102	Wheeler, W. C. and Whiteley, Peter M.	2015		102	basic	English	U
...
 Dunn-2011-146	Dunn, Michael and Burenhult, Niclas and Kruspe, Nicole and Tufvesson, Sylvia and Becker, Neele 	2011		146	areal	English	Aslian	http://www.jbe-platform.com/content/journals/10.1075/dia.28.3.01dun	Dunn2011		This list was created by Niclas Burenhult and Nicole Kruspe in order to document the Aslian languages of the highlands of the Malay Peninsula. This is based on the Swadesh list, but also includes some relevant culture/technology words.	291-323	
 Daniels-2016-85	Daniels, Don	2016		85	basic,areal	English	Magi,Aisi	http://dx.doi.org/10.1353/ol.2016.0004	Daniels2016	Daniels2016	This is a derived version of [Swadesh's list of 100 items (1955)](:ref:Swadesh-1955-100) used to document two languages of New Guinea.	219-221	
+Miedema-1985-	Miedema, J. and Welling, F. I.	1985		40	basic,areal	English	New Guinea		Miedema1985	Miedema1985	This is a list of 40 items 'based on Voorhoeve 1975' 	29-52
```
LinguList: Do we have the original list bo Voorhoeve? Sorry for being a pain in the neck, but can we retrieve this source, or can we reference it by providing a link? It would be good if people who read this could trace it back. It is possible [this source](http://glottolog.org/resource/reference/id/300404), so one might just add the glottolog id, and all is fine.

SimonGreenhill:
Hang on, sorry -- was going to reorder those numbers. Let me do that before committing.

concepticondata/conceptlists.tsv
SimonGreenhill: That will be the next list I add :)

SimonGreenhill: Do you want me to link it in the conceptlists.tsv?

LinguList: All right, super, than just make sure you reference this list either now, or before, e.g., like:

```text
This is a list of 40 items is based the [40-item list by Voorhoeve (1975)](:ref:Voorhoeve-1975-40).
```

That should be fine. Take your time for the rest, I won't merge until you're happy ;-)

concepticondata/conceptlists/Miedema-1985-40.tsv
```diff
@@ -0,0 +1,40 @@
+Miedema & Welling-1985-41-1	1	arm	1673	ARM
```
LinguList: Ah, sorry, there are missing headers:

```text
ID   NUMBER   ENGLISH   CONCEPTICON_ID   CONCEPTICON_GLOSS
```

Note also that our ID system has the convention to only take the first author, so it should be Miedema-1985-40-X in the ID field, as this needs to be identical with the name of the concept list.

SimonGreenhill:
next attempt!

PR #198, opened 2016-09-26
Dye et al 1968

SimonGreenhill:


SimonGreenhill:
Note: missing term for "sago thatch"

LinguList:
Is this something that occurs frequently, a sago thatch? We currently allow for very specified concepts to be missing (as you'll know), the policy being to add a new concept with definition if it is likely that it may occur in other lists.

SimonGreenhill:
sure, but I've got it two other lists -- Fleischmann and Turpeinen 1976, and Staalsen 1969

LinguList:
so let's add it. Increment the ID, define it's semantic field, make a short definition (maybe just the one from wikipedia), and that's it. You can actually use the same semantic field as for "thatch" (or not?):

```text
279	THATCH	The house	Natural roofing material such as straw, reeds, leaves, etc.	Person/Thing
```

So using this as a role model.

LinguList:
Hi @SimonGreenhill, so do you plan on adding the new concept set, or should we deal with it afterwards?

SimonGreenhill:
Not sure -- I think it's just thatching using sago leaves, so do you want to keep that specificity or collapse it into THATCH?

xrotwang:
I'd say collapse it - as long as we don't run into a list with "THATCH" **and** "THATCH using a sago leaf".

SimonGreenhill:
Nothing I've got has both THATCH and THATCH w/ sago, so I think THATCH is ok.

LinguList:
Well, I'd say one should be careful about those. Why so? We once had a data where words for "thin" were collapsed, and it turned out to mess up everything, because there were three words for "thin": thin tea, thin body, thin leaf. With sago thatch, it may well be a compound, sago + thatch, and this might lead people to annotate cognates, but they are not between thatches, but between sagos. If you don't mind, Simon, I'll merge and take over, also adding a relation between "thatch" and "sago thatch". With the relations, you can still compare over nodes, but need to follow our underlying structure.

SimonGreenhill:
All good, go for it 👍

LinguList:
okay, done, I directly pushed to upstream.

PR #207, opened 2016-10-04
refactored readmes

xrotwang:
- creation of markdown tables via tabulate package
- remove badges from concept lists table, because they load very slowly for bigger tables
- extended test suite


LinguList:
cool! I need to have a look at the tabulate-package, as I'm suffering a lot from markdown file creation.

xrotwang:
tabulate only helps with tables. Creating markdown programmatically is a bit weird, anyway, since one of the main goals of markdown is to be editable by hand :)
So I guess using python to create markdown will not be much different to using python to create HTML, for which there are tons of libraries available.

LinguList:
having nice tables will aldready be enough for me, but I would really have needed that in pre-markdown times, for tables in restructured text ;-)

PR #208, opened 2016-10-05
More documentation

xrotwang:
in particular how to use pyconcepticon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,22 @@
-# Concepticon-Data
+# CLLD Concepticon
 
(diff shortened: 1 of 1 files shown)


LinguList:
cool, thanks! As to myself, I am using pyconcepticon much more regularly by now, and it makes my life a lot easier. Writing a little wrapper for lingpy should be extremely straightforward by now, but it may even not be needed after all, as pyconcepticon is good enough as a standalone.

xrotwang:
I'll work on pyconcepticon today, with the goal of making it more consistent, e.g. having the commands of the cli use the API to access the data.

PR #209, opened 2016-10-05
Refactored API

xrotwang:
The API does now return more consistently dicts of mapping IDs to objects for all data types (Concepts, Concept Lists, Concept Sets). I also made sure all data access (in tests or the cli) goes through the API, thus can be used with any kind of clone or download of `concepticon-data`.

Note that this isn't backwards compatible.


concepticondata/conceptlists/README.md
```diff
@@ -81,6 +81,7 @@
 | [Leibniz-1768-128](Leibniz-1768-128.tsv) | 128 | 100 | 0 |
 | [Lin-2010-89](Lin-2010-89.tsv) | 89 | 100 | 0 |
 | [List-2014-1280](List-2014-1280.tsv) | 1280 | 100 | 2 |
+| [List-2015-250](List-2015-250.tsv) | 0 | 0 | 0 |
```
LinguList: I think we need to delete these lines from the conceptlists.tsv, as the data is not existent (list-2015-250, but also Snider-2004-1700 below). List-2015-250 can be directly deleted, as we still work on the data. Snider-2004-1700 was a long dream to have, and I left the description therefore, but the data is messy and difficult to add. So for the time being, we could delete them both.

LinguList:
Looks very nice, what I saw from the examples you give in the description.

xrotwang:
Yes, I really like the light-weight objects which are easily created using the `attrs` package. I think we can live with this API for quite some time.

PR #211, opened 2016-10-06
adding "overlap" as commandline command

LinguList:
this is practical. What I don't know how to solve is how to prevent the printed output in "tests".

I now also decided to put all missing items in Pallas as question marks. So I'd suppose, in the future, we'll use question marks to indicate missing translations.

### DIFF
--- a/pyconcepticon/tests/test_commands.py
+++ b/pyconcepticon/tests/test_commands.py
@@ -53,7 +53,8 @@ class Tests(WithTempDir):
     def test_overlap(self):
         from pyconcepticon.commands import overlap
         Args = namedtuple('Args', ['data', 'args'])
-        self.assertEqual(
-                overlap(Args(data='', args=['Swadesh-1955-100',
-            'Swadesh-1952-200'])),
-                87)
+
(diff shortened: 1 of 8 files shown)


xrotwang:
I think we should document this decision, i.e. "? means missing translation" right away.

pyconcepticon/tests/test_commands.py
```diff
@@ -49,3 +49,11 @@ def test_attributes(self):
 
         with capture(attributes, MagicMock(data=None)) as out:
             self.assertIn('Occurrences', out)
+
+    def test_overlap(self):
+        from pyconcepticon.commands import overlap
+        Args = namedtuple('Args', ['data', 'args'])
+        self.assertEqual(
```
LinguList: super, thanks!

LinguList:
or should we just use NAN? and where would you document, in any way?

xrotwang:
You are right, `NAN` may be better. I'd document this in CONTRIBUTING.md - at least until we find a better place.

LinguList:
we might think of putting what I just wrote in "contributing.md" into a faq.md, as contentwise, this might qualify as "frequently asked questions".

PR #212, opened 2016-10-06
Heuristically map concept lists to Concepticon

### LINKED ISSUE #210: Should we transfer lingpy's gloss-module to the pyconcepticon-api?
We use the function lingpy.meaning.glosses.parse_gloss and the function lingpy.meaning.glosses.compare_conceptlists exclusively in the concepticon. The question is, since these are basically standalones in lingpy, whether it would not better to directly include them in pyconcepticon? In this way, we could use maplist.py as a cli-command without loading lingpy, and lingpy could later provide a wrapper for pyconcepticon. In terms of responsibilities, this would make much more sense: lingpy is used for sequence comparison, pyconcepticon is used for concept comparison.

xrotwang:
The corresponding functionality from lingpy has been copied to
`pyconcepticon.glosses` and refactored. It can be called via API
as
```python
from pyconcepticon.api import Concepticon
api = Concepticon()
api.map('path/to/clist', 'output/path/to/mapping')
```
and via command line as
```
concepticon path/to/clist
concepticon path/to/clist > mapping.tsv
concepticon path/to/clist mapping.tsv
```

closes #210

### DIFF
--- a/pyconcepticon/api.py
+++ b/pyconcepticon/api.py
@@ -9,11 +9,13 @@ import attr
 from clldutils.path import Path
 from clldutils import jsonlib
 from clldutils.misc import cached_property
+from clldutils.dsv import UnicodeWriter
 
 from pyconcepticon import data
 from pyconcepticon.util import (
     REPOS_PATH, data_path, read_dicts, split, lowercase, to_dict, split_ids,
 )
+from pyconcepticon.glosses import concept_map
 
 
 class Concepticon(object):
@@ -86,6 +88,25 @@ class Concepticon(object):
             id=id_,
             meta=jsonlib.load(md_path),
             values=to_dict(read_dicts(values_path), key=itemgetter('CONCEPTICON_ID')))
+
+    def map(self, clist, out=None):
+        assert clist.exists()
+        from_ = []
+        for item in read_dicts(clist):
+            from_.append((item['ID'], item.get('GLOSS', item.get('ENGLISH'))))
+        to = [(cs.id, cs.gloss) for cs in self.conceptsets.values()]
+        cmap = concept_map([i[1] for i in from_], [i[1] for i in to])
+
+        with UnicodeWriter(out, delimiter='\t') as writer:
+            writer.writerow(['ID', 'GLOSS', 'CONCEPTICON_ID', 'CONCEPTICON_GLOSS'])
+            for i, (fid, fgloss) in enumerate(from_):
+                row = [fid, fgloss]
+                match = cmap.get(i)
+                row.extend(list(to[match[0]]) if match else ['', ''])
+                writer.writerow(row)
+
+        if out is None:
+            print(writer.read())
 
 
 def valid_key(instance, attribute, value):
--- a/pyconcepticon/cli.py
+++ b/pyconcepticon/cli.py
@@ -19,11 +19,11 @@ from clldutils.clilib import ArgumentParser
 from clldutils.path import Path
 
 import pyconcepticon
-from pyconcepticon.commands import link, stats, attributes
+from pyconcepticon.commands import link, stats, attributes, map_concepts
 
 
 def main():  # pragma: no cover
-    parser = ArgumentParser(__name__, link, stats, attributes)
+    parser = ArgumentParser(__name__, link, stats, attributes, map_concepts)
     parser.add_argument(
         '--data',
         help="path to concepticon-data",
--- a/pyconcepticon/commands.py
+++ b/pyconcepticon/commands.py
@@ -92,6 +92,11 @@ def attributes(args):
     print(tabulate(list(attrs.most_common()), headers=('Attribute', 'Occurrences')))
 
 
+def map_concepts(args):
+    api = Concepticon(args.data)
+    api.map(Path(args.args[0]), args.args[1] if len(args.args) > 1 else None)
+
+
 def readme(outdir, text):
     with outdir.joinpath('README.md').open('w', encoding='utf8') as fp:
         if isinstance(text, list):
--- a/pyconcepticon/glosses.py
+++ b/pyconcepticon/glosses.py
@@ -1,0 +1,205 @@
+# *-* coding: utf-8 *-*
+"""
+Module provides functions for the handling of concept glosses in linguistic datasets.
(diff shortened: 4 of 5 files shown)


xrotwang:
Note that currently the mapping functionality expects (as is convention for concepticon) columns `ID` and `GLOSS` or `ENGLISH` in the concept list to be mapped.

The resulting mapping file should be easy enough to combine with the unmapped concept list using e.g. [`csvjoin`](http://csvkit.readthedocs.io/en/0.9.1/scripts/csvjoin.html) and [`csvcut`](http://csvkit.readthedocs.io/en/0.9.1/scripts/csvcut.html) to prune the duplicate columns.

xrotwang:
Also note, that there is no support for multiple equally scored matches of individual concepts. Maybe they should just all be dumped into an additional `ALTERNATIVES` column?

LinguList:
cool, I'll test it soon, I'm really curious. As to this point:

> Also note, that there is no support for multiple equally scored matches of individual concepts. Maybe they should just all be dumped into an additional ALTERNATIVES column?

I was having troubles with the lack of this support in the past, as it would give me the less obvious match, like "stone (kernel)" instead of "stone". This annoyed me whenever I mapped a list, but I ignored it for most part, as I introduced the possibility to quickly add the corrected gloss and remap it, using another script from the `helper` folder. Ideally, I think, we could have an underlying ranking of all concepts that selects the most frequent one for mapping. This would immediately give the best match, and in cases of rare words, like stone/pit, one will then just need to recheck for the moment.

LinguList:
Okay, I checked it. First, I think the reduction of columns is a good idea, as they can be transferred back easily later. Second, the multitude of possible answers as based on the weights in the mapping is sometimes very useful, especially in larger lists. In my personal workflow it was always much easier to delete one line which would not apply than to search from an additional column and re-write content from there. But I guess this could be handled by introducing a flag that modifies the output.

In this context, I was just also thinking that it might be useful to check the quality of this automatic mapping procedure. If it turns out that it works well enough, we might apply it to data for lexibank that we then flag as "automatically mapped". And I imagine it might also make sense for dictionaria? But I have no real idea regarding the accuracy. Yet we have tons of data to actually test that.

LinguList:
Ah, almost forgot it: thanks a lot! Concepticon is, also with the other changes, getting really fun to work with.

xrotwang:
Good idea! Most frequent one should typically be what we want, right?
Am 06.10.2016 17:58 schrieb "Johann-Mattis List" <notifications@github.com>:

LinguList:
yes, I guess this is in at least 90% of the cases, if not more!

xrotwang:
I was also thinking of using something like [fuzzywuzzy](https://github.com/seatgeek/fuzzywuzzy#token-set-ratio) to get more fine-grained similarity measures, which may already be enough to break many ties when it comes to multiple matches. If this would get automatic mapping to - say - 99% for all existing concept lists, this would prove its usefulness, right?

LinguList:
definitely! This would be great. I think, if we find time, we should try
and have some "benchmark" to see how well the mapping works with our
existing lists. As this could also help to make people aware of
notorious cases like "thou" where the user just usually needs to check
again, while "stone" is rather obvious (unless one confuses it with
"stone (a kernel)".

xrotwang:
Ok. I did some experimenting, throwing frequency and fuzzywuzzy into the mix. Below is how far I got e.g. for Galucio-2015-90. Summary: 11 manual mappings could not be reproduced, but at least 2 seem to be actual errors: KINGFISHER and ELECTRIC EEL. Due to using fuzzywuzzy, which takes levenshtein distance into account, COCOA is mapped to COCA instead of COCOA BEAN - but this could possibly be corrected with some fine-tuning. I guess we are approaching the point of diminishing returns ... 
```
(concepticon)dlt5502178l:~/venvs/concepticon/concepticon-data$ concepticon test_mapping Galucio-2015-90~~~ mismapped: CAPUCHIN MONKEY (MACACO PREGO) -> CAP; should be CEBUS MONKEY
~~~ mismapped: COCOA -> COCA; should be COCOA BEAN
    alt: COCOA BEAN 2442
~~~ mismapped: CORN -> ACORN; should be MAIZE
    alt: CORNER 192
    alt: CORN FIELD 449
    alt: CORNFLOWER 1960
    alt: UNICORN 2601
~~~ mismapped: ELECTRIC EEL -> ELECTRIC EEL; should be CRICKET
~~~ mismapped: FLY -> FLY (MOVE THROUGH AIR); should be FLY (INSECT)
    alt: FLY (INSECT) 1504
    alt: BUTTERFLY 1791
    alt: SANDFLY 95
    alt: FIREFLY 206
    alt: FLYING FOX 235
    alt: DRAGONFLY 2352
~~~ mismapped: KINGFISHER -> KINGFISHER; should be GECKO
    alt: KING 1508
--- unmapped: Galucio-2015-90-43 LARGE ANT (TOCANDEIRA); should be ANT
--- unmapped: Galucio-2015-90-51 PACU FISH; should be FISH
    alt: FISH 227
~~~ mismapped: PEACH PALM (PUPUNHA) -> PEA; should be PEACH TREE
    alt: PEACH 2043
    alt: PEACH TREE 2358
~~~ mismapped: PINEAPPLE -> PINEAPPLE; should be SILKWORM
    alt: PINE 1740
    alt: APPLE 1320
    alt: PIN 64
~~~ mismapped: SWEET MANIOC -> SWEET; should be CASSAVA
    alt: MANIOC 927
    alt: MANIOC FLOUR 324
    alt: MANIOC BREAD 407
11
```

LinguList:
Okay, so around 90%, and this on a list with particular particular meanings. By adding noun-verb-information, one may also be able to avoid problems like fly (which is taken into account on the gloss-function from lingpy). 

And even more important: PINEAPPLE -> PINEAPPLE; should be SILKWORM is also an obvious error. (I already assigned them to #219 ). Actually not that bad as score, but it also means that automatic mapping should be taken with care and usually, post-checking would be required if one is serious about a dataset. On the other hand, for the lexibank-case, if a list is not mapped, we could use automatic mapping and indicate this, and similarly, I guess, this could be done in dictionaria.

PR #213, opened 2016-10-08
new commands for union and intersection, taking relations into account

LinguList:
This adds a new property, "relations" to the Concepticon. This is a dictionary with two additional methods:

```python
>>> Concepticon().relations.broader('gloss|id', search_depth=2)
>>> Concepticon().relations.narrower('gloss|id', search_depth=2)
```

These methods return broader and narrower concepts (as an iterate object) in the given search range.

Based on this method, union and intersection are now enhanced, and take the relations at a depth of 2 into account. This results in the EXACT comparability of traditional accounts when comparing the two swadesh lists:

```shell
$ concepticon union Swadesh-1955-100 Swadesh-1952-200
...
 23 *BURN                    [2102] BURNING (1, Swadesh-1952-200), BURN (SOMETHING) (1, Swadesh-1955-100)
...
146 *SKIN                    [763 ] SKIN (HUMAN) (1, Swadesh-1952-200)
...
206  YELLOW                  [1424] 
207  YOU                     [1213] 
```
This is an important result, as we want to be able to provide the same level of comparability as people intuitively feel that there should be. In the output, all concepts which are fuzzily matched are preceded by an asterisk, and the narrower concepts are listed in the rest of the line.

Note that it is not necessarily clear whether this method works on all concept lists, as the depth of comparison may well require different adjustments. I take the fact that it works on Swadesh, however, as a first indicator that the method is sufficient (and also sufficiently fast). 

Ideally, we would add a parameter allowing to define the search-depth, but I don't know how to do that in the given command-line settings.


LinguList:
Note: In the most recent addons, I adjusted the algorithm to also account for cases where one list has a broader concept and another actually splits the conceptual space. This is definitely useful and the technique seems to work when testing on Matisoff's list.

LinguList:
addon: the last commit refines the comparison algorithm. We still have a problem in correctly assigning cases where one concept is broader than another concept on the same list (HAIR and BODY HAIR). In these cases, one concept is ignored, which is not the wanted behaviour. But it is questionable, whether we actually want to have these concepts in concept lists anyway. There does not seem to be a fully perfect solution here. 

I also added a new function as_conceptlist, which allows to read conceptlists from a file also if they don't occur in the concepticon-data folder. This is useful for concept lsits comparison, using the api.

LinguList:
I'm still not 100% happy with this, but I take Swadesh-100 + Swadesh-200
= 207 as a first baseline. The problem are concepts that are broader
than other concepts when they come from the same list, like "hair" where
I tested on our Sino-Tibetan list. But obviously, our Concepticon is
missing the concept "hair (head)" here, as we would assume that a
concept list tries to split the semantic space. On the other hand: we
also know that this is not necessarily what happens, as people could
explicitly have the broader and the narrower variant. It's all a bit
complicated, but I hope the current procedure captures the major cases.

xrotwang:
I guess there's not much one can do - other than trying to figure out the most common use case, and provide all sorts of flags for fine-tuning for the non-standard cases.

LinguList:
what I did not do yet is checking that all concepts are actually
covered. I guess, having a simple representation "source-gloss :
linked-to" for all concepts in merged lists, where "linked to" is the
same list-concept-gloss if there is no indirect link. But later this
week or next week. What is important in this context is, that it is
possible to compute unions and intersections for two or more lists
including the network without using up to much computation time. This
seems to work with this solution, more or less...

PR #222, opened 2016-10-25
Updating Mapping Procedure

LinguList:
I propose an extension of the mapping procedure, which renders it multi-purpose:
- map a list to the concepticon
- map two lists

Api-wise, this means, we have a new parameter

``` shell
$ concepticon  --outfile=myfile map_concepts list1 [list2]
```

I also introduce a new mapping procedure, which only compares words with an identical main-gloss. This substantially enhances the speed, as the process of comparison is now mostly linear. The new mapping also produces a refined output which is  very easy for manual postediting, also provides the similarity scores, and shows the percentage of mapped items in the end. All these functions are useful for testing, for example, whether one wants to include a concept list to a dataset, but does not yet know whether coverage will be high enough,etc.

Output is now arranged as follows:

``` text
<<<
GLOSS CID  C_GLOSS      SIMILARITY
gloss   100   HAND              3
gloss   101   HANDSHOE     3
>>>
gloss    200   FOOT             2
#  100   200 0.5
```

To invoke the new behaviour, type:

``` shell
$ concepticon --map_type=2 list1 [list2]
```

This new procedure makes it much more convenient to add new lists, I guess, as ist also increase the accuracy. I could quickly add new lists, and I realized that especially the ranking using the frequencies, now stored in concepticondata/frequencies.tsv, and produced by evoking `concepticon stats` is quite useful as well.

I also modfied the link command slightly: now, if you only submit a gloss, it will add the concepticon-id, which was missing before, or only working if no IDs were in the file.


xrotwang:
Quite a bit of code :)
Will have a look later this week.

LinguList:
yeah, sorry, and also in different parts of the app. Yet the new mapping
is quite small, but in order to allow for the two versions, I needed to
screw around with things. I'm not against re-arranging this, if it is
difficult to maintain in this way. I'd opt for leaving the full-concept
comparison (map_concepts vs. map_concepts2) for the reason that the use
of fuzzywuzzy and the like would best profit from full comparison, while
the new fast implementation requires the logic behind the parse_gloss
function. But the names for the parameters etc. are very ad-hoc, to get
a first proof-of-concept running...

pyconcepticon/util.py
```diff
@@ -162,3 +162,11 @@ def write_conceptlist(clist, filename, header=False):
...
             if k not in ['splits', 'mergers', 'header']:
                 writer.writerow([v[h] for h in header])
+
+
+def load_frequencies(key='GLOSS'):
+    """
+    :returns: dictionary containing frequency information
+    """
+    return dict([(x[key], int(x['FREQUENCY'])) for x in read_dicts(
+        data_path('frequencies.tsv'))])
```
xrotwang: I think dict comprehensions are easier to read, i.e.
```python
return {x[key]: int(x['FREQUENCY']) for x in read_dicts(...)}
```

xrotwang:
I find it a bit weird that a command relies on a file (`frequencies.tsv`), which is only written when running an unrelated different command (`readme`). AFAICT the only reason to do this, is to speed up the concept mapping. I think this should be refactored to make the workflow more transparent; e.g. have an explicit command `compute_frequencies` or similar, which stores the frequencies in the user's config or cache dir and not in the repository.

LinguList:
I see. Here's a question on this: the cached_frequency() decorator that
you use for things, does it do something like this, that is: caching
that somwhere, or is it only during one session? (maybe a stupid question).

Now a second idea: given that the code for reading all concept lists and
calculating freqs is also small, we could just as well do it on the fly,
I suspect, I'll check right now how long it takes.

LinguList:
You are exactly right. I need to nail this in my head, as I keep
forgetting it.

LinguList:
I just tested by redefining frequencies with the cached_property() as a property of a concepticon.api object. It  requires freqs to be passed to the glosses, but this can also be ignored, and it seems to be reasonably small. For the moment, until there is an explicit cache integration, this should suffice, I guess.

LinguList:
BTW: including the frequencies makes a huge difference on the output. In most cases, it is just the most frequent concept that will be mapped, but for a clear mapping, I want to look at all possibilities, so having the dubious cases ordered is a huge relief. Ideally, we would even expand on the mapping procedure, including knowledge of known mappings, and the like, or fuzzy matches, but given that the number of people who will actually need this is small, this can be delayed for the future, in case lexibank expands so quickly that we'll want to have more powerful tools for automatic concept mapping.

LinguList:
I found a problem in the glossing procedure which prevented things like "arm or hand" in the concepticon to be properly included in mappings. I also adjusted concept comparison by adding three versions of glosses with multiple constituents, one normalized form of the structure "arm / hand" as a main-part of a gloss, as this is something we can frequently find in the data, and which may be missed otherwise. In this form, an entry "leg/foot" will be given three possibilities now:

```text
<<<			
foot	1301	FOOT	3
foot	2098	FOOT OR LEG	3
foot	2450	FOOT OR WALK	3
>>>	
```

This is useful, as one might otherwise easily miss the "foot or leg" in the concepticon. What should still be adjusted is the similarity scores.

I also modified the keywords:

```shell
$ concepticon --full_search list.tsv
```

will no use the full-search procedure, while ignoring it will yield the faster procedure which includes frequencies.

LinguList:
Okay, it's again getting bigger, but I had to work on missing lists, and maybe it's best to merge for the moment, before more lists get added and confuse this PR.

One essential point now is that we can even internationalize the concept mapping using the new strategy of having a base list of alias-concepts (a large tsv-file). When mapping, this file is compared with the rest, but the language can be any language, depeding on which the user selects (and which has been prepared). So far we have some 8 languages, automatically created from the currently mapped data. The results will of course differ greatly, depending on the coverage, but especially for the English cases I figured that it is already extremely promising, as we can now handle things like "eggpland" being linked to "aubergine"!

I also fixed many upcoming issues and added more concept sets.

LinguList:
Addon: To trigger multiple languages in concept mapping, write now:

```shell
$ concepticon --language=ru map_concepts yourlist.tsv
```
For "Russian" (the two-letter iso-codes).

xrotwang:
Yes, go ahead and merge. I'll get back to Concepticon next week.
Am 01.11.2016 20:11 schrieb "Johann-Mattis List" <notifications@github.com>:

LinguList:
thanks, and sorry for making such a big bulk of changes...

PR #237, opened 2016-11-03
made the code pep8 compliant

xrotwang:
Mostly re-formatting to make
```
flake8 --ignore=E711,E712,D100,D101,D103,D102,D301 --max-line-length=100 pyconcepticon
```
pass.

But I also refactored the `union` and `intersection` commands to share more (previously duplicated) code.


LinguList:
thanks, looks convincing, especially the reduction of redundancies (I was too much focused on results...)

PR #241, opened 2016-11-04
Enhanced API

xrotwang:
The API was fleshed out to support import of the concepticon data
into the clld app in a convenient way; in particular, I wanted to
make the import agnostic of any directory layout in concepticon-data,
i.e. there should not be any calls to `api.data_path` anymore.

### DIFF
--- a/pyconcepticon/api.py
+++ b/pyconcepticon/api.py
@@ -37,6 +37,14 @@ class Concepticon(object):
         """
         return data_path(*comps, **{'repos': self.repos})
 
+    @property
+    def bibfile(self):
+        return self.data_path('references', 'references.bib')
+
+    @cached_property()
+    def sources(self):
+        return jsonlib.load(self.data_path('sources', 'cdstar.json'))
+
     @cached_property()
     def bibliography(self):
         """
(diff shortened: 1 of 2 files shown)


LinguList:
I have one question regarding the bibfile, as I saw you added a variable. We assume AuthorYear-keys with Author capitalized, right? I was trying the bibtool, but it decapitalized the bibtex keys, after I edited with JabRef. But maybe I was just not using the command correctly...

xrotwang:
The way things are set up now, the bibtex key is used to relate references with source docs, i.e. the PDFs. So if we wanted to switch to lowercase bibtex keys, this would have to be changed in the sources catalog `sources/cdstar.json` as well.

xrotwang:
In any case, I think regarding references and sources our test suite is comprehensive enough to make a switch to lowercase possible. We wouldn't even break URLs, because they are already lowercase:
http://concepticon.clld.org/sources/beaufils2015
:)

LinguList:
I tend to prefer Capitalized bibtex keys, but only for historical reasons. It's just that I was a bit worried about the output of bibtools and left my commit without applying bibtools.

xrotwang:
If you can maintain the capitalized bibtex keys, all the better.

LinguList:
I'll figure this out ;-)

PR #250, opened 2016-11-17
adding an important list and integrating bibliography

LinguList:
as discussed in the context of lexibank, I am expanding the bibliography and adding all entries from the notes in conceptlists.tsv to the bibliography. Links to evobib are replaced by `[text](:bib:Key)` structures, and tests are modified to check whether unlinked references are in the data.
The body-part list of Snoek-2013-61 adds a long missing distinction of toenails vs. fingernails and alternative terms for body parts. I also modify our master-list for english terms which has now about 10000 unique glosses which are all variants of our about 2600 concept sets. Linking is heavily improved with this, and automatic mapping for lists like the one by Snoek reaches 93 percent with no corrections that I had to provide, apart from adding the missing concept sets.


LinguList:
I should add that I added the source Snoek2013.pdf as is to the folder, assuming and hoping that this is the way we currently proceed (or is there a way that I add the data to the repo directly?)

xrotwang:
Yes, that's the way. I'll take care of it later today.
Am 17.11.2016 14:15 schrieb "Johann-Mattis List" <notifications@github.com>:

xrotwang:
Smaller, more focussed pull requests would make reviewing easier :)

LinguList:
I guess I agree. The problem here was -- as always -- the question of whether I should make the PRs directly with the risk of having one every day, or wait until they accumulate. But I'll try to find a balance between the two in the future...

xrotwang:
Snoek2013.pdf is uploaded now: https://github.com/clld/concepticon-data/commit/59534bde5a839035eb81a6ef73fc71cf58057c93

LinguList:
excellent. Thanks!

PR #255, opened 2016-11-29
Mapping from concept sets to integer numbers

### LINKED ISSUE #254: "less than"/"greater than" relations
To characterize and use concept sets for numerals more effectively, it would be good to tag them - maybe even with a "type" annotation? And add "less than"/"greater than" relations.

xrotwang:
closes #254


LinguList:
Cool, I wonder how you did that so quickly.

xrotwang:
vim, mostly.

LinguList:
I knew it ;-)

PR #261, opened 2016-12-04
Adding missing lists and the like

LinguList:
This PR add new lists, which is why it looks big, but I think it's better to add 5 list at once than only one at a time.

New lists include body parts (very interesting, Wilkin's account on cross-linguistic semantic change), and Swadesh's translations into Russian. I also corrected Swadeshs 100-item list, as this list was wrongly represented: In fact, there is no real LIST of Swadesh of the 100 items, he only mentions that he adds 8 more to the 92 most stable of his 215 item master list (category "fun facts").


xrotwang:
Cool. Will upload the pdfs shortly.

PR #263, opened 2016-12-05
Add a lookup command for single words.

SimonGreenhill:
I wanted to query concepticon for matches for a word (rather than searching through a textfile, yawn). This PR adds the `lookup` command such that:

```
concepticon lookup ash
concepticon lookup "to run"
concepticon lookup ASHES hand "leg or foot"
```

prints output as follows:

```
GLOSS	CONCEPTICON_ID	CONCEPTICON_GLOSS
ASHES	646	ASH
hand	1277	HAND
leg or foot	2098	FOOT OR LEG
```

### DIFF
--- a/pyconcepticon/api.py
+++ b/pyconcepticon/api.py
@@ -200,47 +200,27 @@ class Concepticon(object):
         if out is None:
             print(writer.read().decode('utf-8'))
     
-    def lookup(self, entry, full_search=False, similarity_level=5, language='en'):
+    def lookup(self, entries, full_search=False, similarity_level=5, language='en'):
         to = self._get_map_for_language(language, None)
         if full_search:
             cmap = concept_map2(
-                entry,
+                entries,
                 [i[1] for i in to],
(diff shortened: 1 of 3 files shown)


LinguList:
Nice, this is a nice feature I'll also use gladly. So far, I always use the concepticon.tsv, but it's extremely annoying to search there.

pyconcepticon/api.py
```diff
@@ -115,30 +115,39 @@ def frequencies(self):
                 if concept.concepticon_id:
                     D[concept.concepticon_gloss] += 1
         return D
-
+    
+    def _get_map_for_language(self, language, otherlist=None):
+        self._to_mapping = getattr(self, "_to_mapping", {})
```
xrotwang: Could we move initialization of class instances into `__init__`? Code becomes hard to reason about (also for IDEs), when that property isn't met.

SimonGreenhill:
See this PR for the revised: https://github.com/clld/concepticon-data/pull/264

PR #265, opened 2016-12-06
speed up tests

SimonGreenhill:
This PR moves the tests requiring `Concepticon` into their own class, and makes them re-use a common `api` object to avoid the expensive set up times. It shaves about 4 seconds off the test run time(~25%). Thoughts?

### DIFF
--- a/pyconcepticon/tests/test_api.py
+++ b/pyconcepticon/tests/test_api.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals, print_function, division
 
 from clldutils.testing import capture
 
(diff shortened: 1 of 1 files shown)


xrotwang:
Since these tests are rather integration tests, i.e. testing the functionality **and** the data, they rely on the same data being accessed by the API anyway. So having them share read-only data across tests is ok.

SimonGreenhill:
yes, and they should all be idempotent anyway, right?

xrotwang:
Yes. Since most of the data is loaded in `@cached_property`s, nothing will change after the first access anyway.

PR #266, opened 2016-12-06
Adding Bromley 1967

SimonGreenhill:
Don't commit this yet. A few items are not linked and have ??? as their CONCEPTICON_ID and GLOSS. Checking these with @LinguList.


concepticondata/conceptlists.tsv
```diff
@@ -185,3 +185,7 @@ Dunn-2011-146	Dunn, Michael and Burenhult, Niclas and Kruspe, Nicole and Tufvess
 Miedema-1985-40	Miedema, J. and Welling, F. I.	1985		40	basic,areal	English	New Guinea		Miedema1985	Miedema1985	This is a list of 40 items based on the [40-item list by Voorhoeve (1975)](:ref:Voorhoeve-1975-40).	29-52	
 Voorhoeve-1975-40	Voorhoeve, C. L.	1975		40	basic,areal	English	New Guinea		Voorhoeve1975	Voorhoeve1975	A 40 word item list possibly derived from unpublished wordlists of J.C. Anceaux	94-129	
 Dye-1968-120	Dye, W. and Townsend, P. and Townsend, W.	1968		120	basic,areal	English	New Guinea		Dye1968	Dye1968	A 120 item list for a survey of the Sepik languages.	151-152	
+Bromley-1967-110	Bromley, M.	1967		110	basic,areal	English	New Guinea		Bromley1967	Bromley1967	A 110 item list possibly derived from Swadesh for a survey of Grand Valley Dani languages	305-307
```
LinguList: I'd prefer a link with Swadesh, so writing in the note field:

```text
A 110 item list, possibly derived from [Swadesh's 200-item list](:ref:Swadesh-1952-200) for a survey of Grand Valley Dani languages.
```

Also the dot at the end of the note is something I commonly use (for consistency).

If possible, providing the full normal name (M = Michael?), but only if given in source, as we agreed on this in the past (although I myself tend to leave it, as it may be annoying to look for people's names if they only have initials on their papers).

LinguList:
I just had a look at the list. In my impression, but for this, it would be useful to have a look at the source, and see what the author says, the elicitation of the past was used to trigger certain grammatically simple word forms (see also the usage of the plural / 2nd singular for all other verbs). This is also in Paul's list of Germanic, in the sound comparisons, as "drink!" is in German "trink!", while otherwise you had the infinitive. Ideally one could confirm this through the source, but I'd otherwise just make a bold decision (it can be changed, if we really get a list were they distinguish "drink!" from "drinks", etc.

concepticondata/conceptlists.tsv
LinguList: Sorry, but I just realized: it's a 111 item list. And it really seems it comes from Swadesh, but rather his 100 item list plus the 200 item list, as "breast" s new in the 100 item list. Once we make a cladistic analysis of Swadesh-lists in concepticon (with shared errors, and the like ;-), we will know who copied from whom...

LinguList: So to finish this NOTE, I'd suggest to reword like this:

```text
A 111 item list of so far unclear origin (definitely not directly derived from the [100](:ref:Swadesh-1955-100) or [200](:ref:Swadesh-1952-200) lists by Swadesh) for a survey of Grand Valley Dani languages.
```
It contains some strange new items, so we need the stemmatic analysis to be sure in the future.
```

SimonGreenhill:
1. Can't find the author's first name (think it's Myron, but could be wrong) and it's "M. Bromley" in the article.

2. Have gone through the PDF and revised some assignments. Can't see any mention of "died" or "I heard", so will leave these unlinked. Have also left "not know" unlinked.

PR #270, opened 2016-12-08
Fixes to lookup

SimonGreenhill:
I realised that `lookup` only returns one match instead of all matches. In fixing it, I've simplified the API to, rather than a dict of lists, return a generator of tuples (searchterm, concepticon gloss, concepticon id, similarity). 

This PR also adds a test or two.

There's an ugly `type(x) == int` check as it seems that the concept_map* commands can return either a bare integer or a list of integers.

### DIFF
--- a/pyconcepticon/tests/test_commands.py
+++ b/pyconcepticon/tests/test_commands.py
@@ -77,7 +77,7 @@ class Tests(WithTempDirMixin, TestWithFixture):
     
     def test_lookup(self):
         from pyconcepticon.commands import lookup
-        with capture(lookup, MagicMock(args=['sky'])) as out:
+        with capture(lookup, MagicMock(args=['sky'], language='en')) as out:
             self.assertIn('1732', out)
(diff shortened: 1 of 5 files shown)


pyconcepticon/tests/test_commands.py
```diff
@@ -74,3 +74,10 @@ def test_intersection(self):
                 intersection,
                 Args(data='', args=['Swadesh-1955-100', 'Swadesh-1952-200'])) as out:
             self.assertEqual(94, len(out.split('\n')))
+    
+    def test_lookup(self):
+        from pyconcepticon.commands import lookup
+        with capture(lookup, MagicMock(args=['sky'])) as out:
```
xrotwang: I think
```python
MagicMock(args=['sky'], language='en')
```
will fix the failing test.

SimonGreenhill: fixed!

PR #282, opened 2017-01-03
removing trailing whitespace in Birchall et al

SimonGreenhill:
Some of the english definitions in this list have trailing white space. I've removed them here as it was making my mapping for lexibank annoying. 

Perhaps a trailing whitespace check would help in the tests or on loading from the conceptlist?


LinguList:
Yep, definitely we should consider. Here, it was my error in not regarding it, as the mapping procedure automatically removes white-spaces initially and finally. A check for all concept lists is probably the best idea do deal with this.

PR #285, opened 2017-01-05
Update Tryon-1983-324.tsv

SimonGreenhill:
This is lower case in the source.

### DIFF
--- a/concepticondata/conceptlists/Tryon-1983-324.tsv
+++ b/concepticondata/conceptlists/Tryon-1983-324.tsv
@@ -176,7 +176,7 @@ Tryon-1983-324-174	174	my	Possessives	2090	MY
 Tryon-1983-324-175	175	your sg.	Possessives	732	YOUR (SINGULAR)
 Tryon-1983-324-176	176	his	Possessives	2092	HIS OR HER
 Tryon-1983-324-177	177	our pl. incl.	Possessives	780	OUR
-Tryon-1983-324-178	178	our pl. Excl.	Possessives	780	OUR
+Tryon-1983-324-178	178	our pl. excl.	Possessives	780	OUR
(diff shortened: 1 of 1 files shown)


concepticondata/conceptlists/Tryon-1983-324.tsv
```diff
@@ -176,7 +176,7 @@ Tryon-1983-324-174	174	my	Possessives	2090	MY
 Tryon-1983-324-175	175	your sg.	Possessives	732	YOUR (SINGULAR)
 Tryon-1983-324-176	176	his	Possessives	2092	HIS OR HER
 Tryon-1983-324-177	177	our pl. incl.	Possessives	780	OUR
-Tryon-1983-324-178	178	our pl. Excl.	Possessives	780	OUR
+Tryon-1983-324-178	178	our pl. excl.	Possessives	780	OUR
```
LinguList: Ups... I guess we should be careful and already now add another concept set: 

* OUR
* OUR (INCLUSIVE)
* OUR (EXCLUSIVE)

It's annoying, but it would be more consistent.

concepticondata/conceptlists/Tryon-1983-324.tsv
```diff
@@ -176,7 +176,7 @@ Tryon-1983-324-174	174	my	Possessives	2090	MY
 Tryon-1983-324-175	175	your sg.	Possessives	732	YOUR (SINGULAR)
 Tryon-1983-324-176	176	his	Possessives	2092	HIS OR HER
```
LinguList: Just realized that HIS OR HER is probably worth to be split into:

* HIS OR HER
* HIS
* HER

given that we have only his here.

LinguList:
I just made an issue from my comments. They are independent of this PR:

* https://github.com/clld/concepticon-data/issues/286

PR #291, opened 2017-01-25
Refactored handling of aliases

### LINKED ISSUE #216: We have the concept set "betel nut" twice
There's

```
217     BETELNUT        Agriculture and vegetation      The seed of the areca palm (Areca catechu).
```

and

```
2369    BETEL NUT       Food and drink   The seed of the areca palm (Mama'an in Meranau common version).
```

### LINKED ISSUE #217: Check linking status of duplicate concept sets
Quite a few concept sets are related via `sameas` relations. It should be checked whether in all these cases only one of a set of duplicate concept sets is actually linked from any concept list. If this is the case, the duplicates can be regarded as synonyms, but should be clearly marked as such.

xrotwang:
- concepticon.tsv now has a column REPLACEMENT_ID which - if not-empty - links
  to a conceptset ID of a concept set which supersedes the current one.
- The API has been updated to make this column accessible.
- A check for links to superseded concept sets has been added to the tests.
- The mapping helper file (for english) has been updated, redirecting aliases of superseded
  concept sets.

fixes #217 
fixes #216


LinguList:
nice! Many thanks!

PR #296, opened 2017-02-12
Stokhof 1975

SimonGreenhill:
This adds a few irritating combined concepts (MAN OR MALE etc). 

Note, I haven't added "the snake sees me" and "I see the snake" as I wasn't sure how to handle these.


LinguList:
If concepts in a questionnaire get too "phrasy", we should avoid them pragmatically, as this is the job of the "glossicon" for interlinear glosses to deal with them. Unless it's really one big compound, I'd also not expect that such concept sets could help for lexical comparison, but if people manage to work on the actual data and identify the word for "snake" in that phrase, they could make a derived list...

LinguList:
and I'm glad we have man or male now, as quite a few existing lists have this problem. question is: does the list really leave the question open, whether it's s property or a person? In most lists I found so far, there was enough evidence externally to link to "male person", which is itself broader than "man", as it does not specify the age (and "man" is usually "adult male").

concepticondata/concepticon.tsv
```diff
@@ -2672,3 +2672,8 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2671	PENIS SHEATH	Clothing and grooming	Sheath to cover a penis	Person/Thing	
 2672	HEADWATERS	The physical world	The source of a river or stream	Person/Thing	
 2673	BIRD OF PARADISE	Animals	A bird of the family Paradisaeidae.	Person/Thing	
+2674	ORAL CAVITY	The body	Cavity that is limited by the lips in front, the cheeks on the side, the soft and hard palate at the top and the floor of the mouth below.	Person/Thing	
+2675	MAN OR MALE	Kinship	Relating to the male sex of humans.	Property
```
LinguList: Yes, the longer I think about it, given that we have the broader gloss "male person", I'd suggest to not introduce this concept here, but to assume, as in the other cases, that the original concept lists does not mean "adult male person or of male gender property", but rather "a male person or an adult male", in which case linking to "male person" should handle it without introducing more concept relations.

SimonGreenhill: ok, I'll remove those two (MAN OR MALE and WOMAN OR FEMALE). No information in Stokhof about what they actually are.

LinguList: So it's as always with semantics: a shaky business, and more some kind of "Fingerspitzengefühl". I never knew what people mean by "man/male", but I guess we have so far assembled quite a few concept lists which have exactly those glosses. So linking them all consistently to the "male person" is at least consequent, and if we understand later what it actually means, we'll be able to change it quickly.

PR #297, opened 2017-02-13
Define concept of spearthrower

SimonGreenhill:
Saw that this was not defined in a few lists, and I'll need it for some New Guinea languages. 

Also, I think the glossing of spearthrower = soldier in the list by Barry Alpher and David Nash is wrong, they're probably recording [Woomera](https://en.wikipedia.org/wiki/Woomera_(spear-thrower))


LinguList:
Great! Thanks a lot. This explains the weird gloss, and I learned something more about combat.

PR #303, opened 2017-03-20
Clouse 1993

SimonGreenhill:
Note that there's a grammatical thing not defined:

```
Clouse-1993-71-17	17	adj. suffix		
```

What would you like me to do with it?


LinguList:
I recommend to ignore, but if these things keep accumulating, we'd need
a policy at some point. Think of the "concept" "genetive marker", for
example, which is useful for many languages and language families to be
compared (they have it in Hindi, Chinese, Japanese, etc.), but so far we
have not found a list which would actually list this. I'd say, if we
encounter a second list in which the "adj. suffix" is listed as an
explicit questionnaire item, we may start thinking about it, but should
leave it unlinked as long as we do not know whether this is just
peculiar for one language or cross-linguistically valid...

LinguList:
Did you want to submit the PDF and forgot to push it?

LinguList:
Sorry, just saw that the PDF is there. I don't really understand why it throws this error.

LinguList:
Or, no, I think I know: As @xrotwang now re-arranged our PDF-hosting, the PDF should be transferred to some server whose name I forgot, so @xrotwang: should we merge (the rest is all excellent), or do we need to handle the pdf in another way before?

xrotwang:
I'll take care of the pdf. Just merge.
Am 20.03.2017 17:18 schrieb "Johann-Mattis List" <notifications@github.com>:

xrotwang:
I get the following errors:
```
ERROR:conceptrelations:447: invalid TARGET: HAIR OR FUR
ERROR:conceptrelations:447: invalid TARGET_GLOSS: narrower
ERROR:Clouse-1993-71:66: invalid value for concepticon_gloss: FAECES
ERROR:Clouse-1993-71: deprecated concept set 1397 linked for Clouse-1993-71-25
```
are these known?

xrotwang:
Ah, the first two are due to an ill-formatted line in conceptrelations. Will fix.

LinguList:
yes, and the latter should be FAECES (EXCREMENT), not only FAECES, as this lead to confusion.

xrotwang:
Ok. will fix as well.

LinguList:
And instead of 1397 DULL, there should be 379 BLUNT (this is the sameas relation, and we take the smaller one).

xrotwang:
Ok. Sources are uploaded and errors fixed: https://github.com/clld/concepticon-data/commit/0f2e32ddbf979ac1aa91c94bddb68e92fa306b1f

SimonGreenhill:
Sorry for the problems guys :/

xrotwang:
Problems with tests aren't really much of a problem :)

PR #308, opened 2017-04-13
update concepticon, add cihui

LinguList:
Finally, I managed to add this resource, which contains more than a 100 unmapped chars, but which is nevertheless really valuable, also for lexibank, as this will give us some 700 linked concepts for 18 dialects in a cross-semantic cognate collection.


LinguList:
I'll be so frank as to just merge this, as I trust that most mappings are okay enough. This was one of the hardest sets ever to be mapped, and there is no optimal solution, but at least I think I now cover some 80% of all 905 concepts...

PR #311, opened 2017-04-20
adding Voorhoeve 1971

SimonGreenhill:


LinguList:
I yesterday heard there are languages spoken in areas where they don't have mosquitos, forests, and fruits. All the normal things one would choose for some Swadesh list. But how can we actually talk of "basic vocabulary" if we can imagine environments where it's not needed? What are the words all languages must have? ... anyway, thanks for adding.

PR #327, opened 2017-05-14
Foley 86

SimonGreenhill:
Adds definitions for "FEMALE PERSON OR MOTHER", "TOMORROW OR YESTERDAY", "SAGO GRUB" and "LEAF SPINE". 

I did not define a term for "sago refuse" as I haven't seen this in other datasets.


LinguList:
I think, the error lies in the gloss for "faeces", but cannnot look it up right now, maybe I changed to something different, and maybe this is not yet in the mapping algo (although it should be). Otherwise another cool list, and we should look soon into the network of relations, as it is growing and will show interesting things,,,

SimonGreenhill:
("fixing faeces" is a commit message I never thought I'd ever write)

PR #328, opened 2017-05-15
exposing test command in API

SimonGreenhill:
This small PR makes the command `concepticon test` run the tests in test_data. This means that data integrity can be checked locally instead of waiting for travis.

### DIFF
--- a/pyconcepticon/commands.py
+++ b/pyconcepticon/commands.py
@@ -485,3 +485,9 @@ def check(args):
     for clist in clists:
         _get_missing(api, clist)
         _get_mergers(api, clist)
+
+
+@command()
+def test(args):
+    from pyconcepticon.tests.test_data import test as _test
+    _test()


LinguList:
Nice idea!

xrotwang:
I think this is good because it makes for a generally more homogeneous/integrated experience, but wanted to point out that running `nosetests` should have made it possible already to run the integrity checks locally. I just realized, though, that the new sub-command runs the integrity checks without actually requiring `nose` or `mock` to be installed, so: good!

LinguList:
Yes, I was amazed how quick it was. I was too lazy to test if it captures all errors but suppose it does? I used to use nose in the past, but it always is this annoying 21, 22, 23 feeling of waiting the 5 seconds to run all tests...

SimonGreenhill:
The PR is definitely just a hack i.e. I'd intended the use case to be "check my just added wordlist". Perhaps a better idea would be to extend the "check" command to "check wordlist.tsv"?

xrotwang:
Possibly more fine-grained control over what tests are run may be good. But considering there are relations between different parts of the data and a test run still taking only seconds (?), I'd keep it as is for some time to come.

LinguList:
Well, I tell you, I just tested with a wrong linking, and it is so much more quicker than the nosetests:

```shell
$ time nosetests
real	0m12.171s
user	0m12.076s
sys	0m0.088s
```

vs.

```shell
$ time concepticon test
real	0m1.521s
user	0m1.492s
sys	0m0.024s
```

xrotwang:
Ok, that's good. And it's what I mean: As quick as it is now, we can live with running the whole integrity test from `concepticon test`.

PR #350, opened 2017-06-27
update northeuralex, add data to concepticon

LinguList:
I changed some further instances, also refining one concept set which still remained with a comma. There are borderline-cases, like "SINK" which is now split inot two concepts, one meaning German "versinken" one "sinken". But in this way, we do justice to NELEX. 

I started using Merriam-Webster to create definitions. This is why I also added it as metadata. I also wrote a custom-script to pull wikipedia definitions. If data keeps increasing in terms of new concept sets at a fast pace, we may consider adding something like this to the Concepticon API. But I don't expect similar bursts in the nearer future.


xrotwang:
@LinguList are you going to fix this: https://travis-ci.org/clld/concepticon-data/builds/247657871#L535

LinguList:
sure, didn't see it as it runs without errors with me

LinguList:
ah, I see it, forgot to push concepticon.tsv.

LinguList:
We need to make sure to report back to the nelex people. They could by now refine the concepticon-mappings, and would ideally provide us with the urls, so we can link from concepticon. I'll also add them to the list of contributors.

LinguList:
okay, pushed this as well, so they are now also listed as contributors.

xrotwang:
Still fails, though: https://travis-ci.org/clld/concepticon-data/builds/247663641#L535

LinguList:
grrr, on it, one second.

LinguList:
strange that ```Concepticon test``` does not report this issue...

xrotwang:
Hm. That is strange. We should look into this. maybe the test is looking at files not under version control?

PR #351, opened 2017-06-27
closes #331

### LINKED ISSUE #331: make cli-command to re-adjust the mapping background-data
currently, the data is stored in pyconcepticon/data, as `map-{0}.tsv`, with `{0}` being replaced by language-two-letter code (`map-en.tsv`, `map-de.tsv`, etc). The data is produced using a highly idiosyncratic script. This should be added to the `cli.py` (of course in enhanced form) and then one should be able to update the data for a particular language (or even for all languages), by typing some concepticon-command in the command line.

xrotwang:


LinguList:
looks fine to me, excellent work, that we have this now as part of the mapping procedure!

Issue #356, opened 2017-07-03
Invalid/non-standard concept identifiers


xrotwang:
We have four released concept lists with invalid concept identifiers. The concept identifiers in these cases do not have the concept list ID as prefix, but are mostly missing the list-length component of the ID, or just a dash.
I see two options to deal with this:
- Just fix the IDs, leaving appropriate commit logs as only trace.
- Fix the IDs, but store a mapping of revoked IDs as part of the data.

For concept set IDs, I'd definitely go for the second option. For concept IDs, I'm inclined to go with the simpler option 1, and add a test to not make this happen again.

The relevant concept lists are:
- http://concepticon.clld.org/contributions/Syrjaenen-2013-226
- http://concepticon.clld.org/contributions/Zheng-2007-86
- http://concepticon.clld.org/contributions/Luqman-2010-65
- http://concepticon.clld.org/contributions/Borin-2012-40

PR #358, opened 2017-07-03
Describe non-standard concept list columns via JSON metadata

### LINKED ISSUE #100: columns in concept lists and regular attributes
We have a nice zipfian distribution with the columns with metadata in all concept lists (I paste it below).

I started tagging columns for regularity, currently here:
- https://github.com/clld/concepticon-data/blob/master/concepticondata/data.py#L60

Here are some major types:
- RANK: There are important things, like "rank", which should always be an integer, not float, since rank assumes that we fully resolve the rank of a wordlist (my next PR will correct issues on this).
- LANGUAGES: Furthermore, we have languages, which should be identified as such. I tagged them as "language" in the data.py file above
- SUBLIST: Furthermore, we have sublists, which means that they contain links to other lists, or just an indicator to which it belongs. This is not very systematic at the moment (thaks to the sources).
-  CATEGORY: is actually nice, since it refers to "semantic field" and the like.
- URL: But also XXX_URL (!) refers to links, and this is already automatically handled.

Now, problems are cases like:
- SWADESH_1955_100, etc. where the people link to another concept list by assignign an asterisk or some other symbol or a + and a -, but ideally, we would indicate already when naming the column that this refers to another concept list, so we could write "CONCEPT_LIST_SWADESH_1955_100" or is this an overkill?
- columns like "SCORE", "FREQUENCY", etc. are sometimes composed (AGE_SCORE, SPANISH_FREQUENCY"), sometimes uncomposed, and they usually refer to numerical data (float or int). Is there a way to formalize it or would it also be an overkill?

The general question is: can we regularize some of these aspects without being forced to add json-descriptors?

xrotwang:
This PR closes #100 in the way discussed there, i.e. by adding metadata describing (or at least providing a place to describe) non-standard data in concept lists.

Note that we also introduce a metadata description for the standard columns (see conceptlists/default-metadata.json), which also dictates a format for concept IDs. This format in turn requires fixing #356, which is rolled into this PR.


xrotwang:
I used some heuristics to derive proper datatypes from column names, but additional corrections by hand will be needed. But I think we can defer this editing to the moment when the datum under consideration is actually put to use in some sort of analysis.

PR #361, opened 2017-08-04
Fixing some errors and re-creating the mapping-data

LinguList:
found some bugs when testing, thanks to @laiyunfan, now fixed them all (hopefully)


LinguList:
I will quickly merge this, since @laiyunfan is about to submit another list.

PR #388, opened 2017-11-20
addding cobl list for concepticon

LinguList:
this is also to introduce @chrzyki and @tresoldi to some of our practice when submitting to concepticon. You may especially want to pay attention to the metadata, which helps us to control for the content and makes it easy to read in the concept list via the csvw spec.


concepticondata/conceptlists.tsv
```diff
@@ -1,4 +1,5 @@
 ID	AUTHOR	YEAR	LIST_SUFFIX	ITEMS	TAGS	SOURCE_LANGUAGE	TARGET_LANGUAGE	URL	REFS	PDF	NOTE	PAGES	ALIAS
+Heggarty-2017-200	Heggarty, Paul and Anderson, Cormac and Scarborough, Matthew	2017		200	basic	English	Indo-European languages	http://cobl.clld.org	Heggarty2017		This is the basic list of 200 items which were selected for the Cognates in Basic Lexicon project. The compilers selected the items following various different criteria, such as ease of elicitation, non-fuzziness of the meaning, representation in Indo-European languages, and more. The numbers are at times higher than 200. This results from the earlier selection which containted more concepts which were no longer used in the official version.
```
xrotwang: I'm not sure, whether the domain `cobl.info` will be used for the clld app. If so, we'll have to make sure to keep `cobl.clld.org` as alias.

concepticondata/conceptlists/Heggarty-2017-200.tsv
```diff
@@ -0,0 +1,201 @@
+ID	NUMBER	URL	ENGLISH	DESCRIPTION	CONCEPTICON_ID	CONCEPTICON_GLOSS	EXAMPLE_CONTEXT
+Heggarty-2017-200-172	172	http://cobl.clld.org/parameters/172	thin	'Thin' of an object, not a person.  Opposite of 'thick', not of 'fat'.	2249	THIN (OF SHAPE OF OBJECT)	This piece of wood is thin, and that one is thick.
```
xrotwang: For most other concept lists, the ID column is composed of the prefix and consecutive integers rather than the local numbers. Maybe we should use this practice here as well, in particular because the local numbers are not under our control.

concepticondata/conceptlists.tsv
LinguList: So should I add cobl.info now, or should I rather leave it as is?

concepticondata/conceptlists/Heggarty-2017-200.tsv
LinguList: Okay, will refine.

concepticondata/conceptlists.tsv
xrotwang: I guess leave as is. Was rather a note-to-self.

PR #392, opened 2017-11-22
Suggested changes needed for Robinson's data on Alor-Pantar.

tresoldi:
Preparing for Robinson (2012) data. The suggested concepts for inclusion are:

- BRIDE PRICE, which seems a general enough concept and which probably features in many lists compiled with anthropology in mind (however, neither this or DOWRY are in Concepticon...)
- BARN OR SHED, the other concepts are either too generic or already used in Robinson's list; I didn't add a concept relation because was not sure which one would be; GARDEN HOUSE does not match the concept for the Alor-Pantar culture
- CONJUNCTIVITIS, is a concept like found in all languages (could very well be in an expanded Swadesh list)
- COOKED CORN, I was not sure about this, but Concepticon already has COOKED RICE and COOKED CORN is likely to come up in many Native American languages
- GRAVEL, this would probably need a relationship, perhaps with ROCK?
- SUGAR PALM, LONTAR PALM, MORINGA TREE, are very important cultural concepts of Southeast Asia and Concepticon seems to be very open to specific plant names
- STARFISH, I was actually surprised this one didn't come up until now
- SCARED, there is no property of "being scared", at least as far I could tell

Among the concepts in Robinson that I did not map but might be open to inclusion are RAISED PLATFORM and TO WAKE SOMEBODY UP (transitive, Concepticon only has an intransitive one of WAKING UP). Other unmapped concepts are very specific cultural items, numbers (123, 100,000), terms relating to lices and one action ("to bathe a child", different from "to bathe" -- it is not clear, but maybe Robinson's "to bathe" is reflexive and "to bathe a child" is a plain transitive?)


concepticondata/conceptlists.tsv
```diff
@@ -1,4 +1,5 @@
 ID	AUTHOR	YEAR	LIST_SUFFIX	ITEMS	TAGS	SOURCE_LANGUAGE	TARGET_LANGUAGE	URL	REFS	PDF	NOTE	PAGES	ALIAS
+Robinson-2012-397	Robinson, Laura C and Holton, Gary	2012		397	basic	English	Alor-Pantar languages	http://booksandjournals.brillonline.com/content/journals/10.1163/22105832-20120201	Robinson2012		The authors inform that the list is built upon the about 400 items of Holton (2012), but this source was unavailable (and apparently the dataset itself was never published).
```
LinguList: one little point: we have the ambition to only quote papers which are referenced (if you find counter-examples, blame me, I'll make them up), so you should

```
The authors inform that the list is built upon the about 400 items of [Holton (2012)](:bib:Holton2012), but this source did not submit the dataset.
```
And put the :bib: into the references. I also changed the wording slightly, as the paper is from the same year, and we could say: with this source, the dataset was published in some sense.

tresoldi:
Oh, ok. Very good practice, btw. :)

LinguList:
Excellent, I was not sure whether it was :bib: but the tests say: yes!

LinguList:
You get the drive, I'd say, nice work! I'll have to prepare the next datasets soon...

LinguList:
one last remark: TO WAKE SOMEBODY UP, didn't read carefully before: we might at some point want to check whether what we have already in there is all plain transitive. Maybe it's another case for "burn" "burning" "burn (something)"? I think, ideally we should file an issue to not forget this. I'll give you writing access, so you can flag this issue as "concept linking problems".

LinguList:
sorry, I don't have the perms: @xrotwang, could you please set up @tresoldi for write access, so that he can also label issues and add them to milestones? I'd say, this is not immediate, so no need for milestone 1.1.

xrotwang:
invite has been sent by github

LinguList:
Thanks, I filed too many issues recently, so that I'm getting a bit reluctant of filing another one (at least not today).

PR #395, opened 2017-11-25
 Corrections and additions to Zgraggen-1980, for the zgraggenmadang dataset.

tresoldi:
I have tried to balance corrections and addictions with as few changes as possible; some of the additions (such as RIPE/GREEN COCONUT) might be a bit too specific, but they are needed as the concepts were already there (but it did not go as far as adding the various arrow types as hyponims).

One thing I did not touch, as it involves some theoretical discussion and maybe a more advanced understanding of Papuan languages, are the grammatical concepts (pronouns, possession, negation, etc.) -- they are not in Concepticon and were only added to the Z'graggen list.


LinguList:
Thanks a lot, I am very happy with this!

tresoldi:
I'm glad you like it. The grammatical mappings still need work for this to be "finished" (all those singular/dual/plural, inclusive/exclusive, possessives, obliques, etc. are a bit of a pain, but should probably be included in the general concepts -- typologists would be trilled).

LinguList:
I suppose you make an issue and flag it with a new milestone 1.2, as for version 1.1, we won't have the time.

Issue #389, opened 2017-11-20
Concept mapping for gloss languages other than English 


IndianaTones:
Apologies, but I cannot find the correct code for how to map using a glossing language other than English (in this case French). 

I usually use:

$ concepticon map_concepts INFILE.tsv > OUTFILE.tsv

so not sure where to indicate the difference.

xrotwang:
I guess `map_concepts` could use a docstring:
```
$ concepticon help map_concepts
None
```

xrotwang:
```
$ concepticon map_concepts --language fr  INFILE.tsv > OUTFILE.tsv
```
should do the trick.

xrotwang:
Please leave this issue open until the command is documented.

LinguList:
Documentation has been so far prepared for the tutorial: 

* https://github.com/digling/edictor-tutorial/blob/master/Historical%20Language%20Comparison%20with%20LingPy%20and%20EDICTOR.ipynb

Search for 2.5: "Enhancing your data".  This could probably be copy-pasted.

IndianaTones:
Thank you - the solution is a bit tricky as I had to label the glosses column as ENGLISH (even though it is obviously French).

The string which works is:

$ concepticon --language=fr map_concepts Concepts-Fulfulde.tsv

as warned, the output isn't terrific- 430/3343, 13%, but better than nothing!

LinguList:
Oh, that is annoying, but no need to label it to "ENGLISH", "GLOSS" will also work, but this is a real bug, we should fix that.

IndianaTones:
Just wanted to update that the glosses it found were really accurate. I got a total of 414 solid entries from this so thank you!!!

LinguList:
wow, this shows that it is a useful function. We hope wit the next release we'll have it working more properly. And with new data to be added (more lists in French, etc.), it will hopepfully even work better on different languages...

Issue #382, opened 2017-11-07
refine concept mapping algorithm and handling


LinguList:
- [ ] modify "map_concepts" to "map" or "link" in cli.py (underscore annoys too much)
- [ ] check description for scores and add them to some tutorial
- [ ] refine mapping for some inconsistent cases (see below)
- [ ] don't convert all to lowercase, keep case for an additional layer of similarity (if all is identical)

The following is unexpected:
```sh
$ concepticon lookup aubergine
GLOSS	CONCEPTICON_ID	CONCEPTICON_GLOSS	SIMILARITY
aubergine	1146	AUBERGINE	4

$ Concepticon lookup "aubergine (noun)"
GLOSS	CONCEPTICON_ID	CONCEPTICON_GLOSS	SIMILARITY
aubergine (noun)	1146	AUBERGINE	3
```
Our rule says: if there is no pos-information, penalize this, but top score is only obtained upon identity:

```sh
$ Concepticon lookup "THE AUBERGINE"
GLOSS	CONCEPTICON_ID	CONCEPTICON_GLOSS	SIMILARITY
THE AUBERGINE	1146	AUBERGINE	1
```

There needs to be a better logic for the scores, and we should have a convincing scoring scheme...

LinguList:
I just figured that the calculation of the self.frequencies of Concepticon is taking an extremely long time, since it is reading every list, which is hampering our automatic lookup. I would suggest to either store frequencies explicitly in a text-file, maybe in pyconcepticon/data/ and then recompute it once in a while, or to drop it completely (although frequencies are useful).

LinguList:
We can argue that the pysem library offers a more consistent mapping now. We would only need to add cmd line functionality.

PR #397, opened 2017-11-26
Including a (preliminary) conceptlist for Sun 1991 data (sunzmyyc on lexibank)

tresoldi:
A lot of work into this one, and I am still full of doubts. The concept list with automatic mappings had its good share of problems, in part due to the complex English glosses of the original; in particular, some of the concepts are only used for one language. The process left me with many ideas on better (semi-)automatic alignment, we should discuss this later.

As I know nothing of TB languages, my only resources to infer concept mappings were Google and the Wiktionary, so please excuse the many mistakes I certainly did. Some English glosses with possible Concepticon matches were left blank when I could not assert without reasonable doubt the correspondences (like in verb/noun distinctions, harder with TB languages, or the many cases of homonyms). Please note that, in some cases, the problems might be in the source itself, and someone with good knowledge of the data might think of generating a new dataset after a good review.

While I tried to be conservative in terms of new concepts, there were many that seemed good candidates in terms of future work on Eastern Asian languages.


LinguList:
Interesting, are these really 1255 concepts? I lived a long time under the impression we were dealing only with some 1000 concepts. I think I should double-check this thoroughly, before we go on. I'm only afraid I won't have full time doing so tomorrow, so it may have to wait until Tuesday. What we can also do is: @tresoldi submits this list now, and I myself refine later, when I find time. Given that we can reference concept lists in concepticon through lexibank, all changes that I might make would immediately affect lexibank as well. What do you think, @xrotwang and @SimonGreenhill ?

tresoldi:
There are 1255 unique labels, which I guess map to around 900 concepts or less -- some are clearly the same (like `arm (upper)` to `arm`, all the months with the animal name, `again;again` to `again`, and so on) and there are many specific ones (like `mat woven from bamboo leaves`) with a single entry for a single language. I can provide you my list of supposed "homoconcepts", but it is what you get from filtering by Concepticon_gloss (I organized the file this way as I am not completely sure about all cases).

As for the PR, feel free to delay it, I can always use a `concepts.csv` and remap to Concepticon's Sun-1991 list when finished.

LinguList:
Actually, the unique labels are due to the OCR, as the SUN-1991 list have their own identifiers. I'll have a look later and decide what to do, as we'd definitely use their numbers as our identifiers, and if I'm not mistaken, they should have them ocred along with the rest (apart from the Chinese glosses) in STEDT. So for me, the concept gloss is the Number in Sun, not the unique labels, which are due to errors in STEDT or additinal annotations in Sun's data, which are also not consistent.

tresoldi:
You mean the source id in `Sun-1991.tsv`? There are 948 of them, but they don't map completely to the 1004 in the automatic mapping`concepts.mapped.csv` (which, however, match the first part of the `srcid` field of the data, `ZMYYC.csv`). Still, the data handling for remapping should not be a problem, the hardest part is to actually check the data. :)

LinguList:
Well, as I said before, even if we won't have Segments, this will be useful for our plans with CLICS, and there is an even bigger list which I dream to link later (1800 concepts!), so I took advantage of your involvement to get this running (it was egoistic, but for a good purpose, as these data need to be linked, they could be so great for further analyses). So if you work on the languages.csv in the meantime, and I will try to devote some time to look into this so I can give some response by Tuesday, I hope this works. Also don't work on the segmentation until having talked to me first, as some things can be done differently and more efficiently there, and if it all turns out to require another 40 hours or so, we'll just say the work is done with this initial mapping and the languages, as lexibank supports different degrees of quality.

I did not look at your Sun-1991-XXX.tsv file yet, but if you preserved the srcid or however that is called in the STEDT data, I should be fine. 

Let's wait what @xrotwang and @SimonGreenhill think of merging this now, and having me refine it later, but alternatively, if I find the time, we might have some answers on Tuesday.

tresoldi:
I was actually wondering if it would end up being used for CLICS, as the patterns (including the combination of morphemes that join different histories, as I read in some paper of yours) were clear even without knowing the languages. As for other lists, I'll write down some of my ideas on improvements to automatic alignment later (nothing fancy, just trivial NLP stuff like sequence similarity, stopwords and the like).

Just one thing: the list I submitted is **not** linked to the `srcid` field from STEDT, which as I said is around 200 entries less -- the same `srcid` can map to more than one label, and one or more labels can map to one concept (one thing that would really help would be a mapping to a language more easily googlable, like Mandarin or Cantonese). I can easily do a map of which STEDT ids link to which concepts (that it, according to this preliminary linking of mine), if you want. Meanwhile I'll go on with the languages.

LinguList:
> automatic alignment 

what do you mean by this exactly? We're talking about concepts, right, do you mean concept linking/mapping?

And as to the srcid: let me please have another look into it, we can still re-use the mappings you made, but as we have fixed identifiers in the book, they will be there, even if STEDT has hidden them, and this is also the version that I had when linking this the first time.

tresoldi:
Yes, here I was using "automatic alignment" for "automatic concept mapping"; it was confusing both because it is a bit of a Portuguese interference ("alinhar") and because in our context we use "alignment" for "segment alignment". Sorry!

LinguList:
Okay, I just went through the data and refined the original concept list:

[Sun-1991-1004.tsv.txt](https://github.com/clld/concepticon-data/files/1505908/Sun-1991-1004.tsv.txt)

What you did not know, but it's my fault, as I didn't tell you, is that the SRCID is built from the gloss-number in the book vs. the number that they use for the language, divided by a dot.

I ran the following code on ZMYYC.csv:
```python
In [28]: wl = Wordlist('ZMYYC.csv', row='gloss', col='language')
In [28]: concepts = defaultdict(set)       
In [28]: for i, gloss, sid in wl.iter_rows('concept', 'srcid'):
    ...:     cid, lng = sid.split('.') if '.' in sid else ('', '')
    ...:     print(cid)                                                                       
    ...:     if cid: concepts[cid].add(gloss)  
In [27]: with open('Sun-1991-1004.tsv', 'w') as f:
    ...:     f.write('ID\tNUMBER\tGLOSS\tFURTHER_GLOSSES\n')
    ...:     for i, (key, cc) in enumerate(sorted(concepts.items(), key=lambda x: int(x[0]))):
    ...:         f.write('{0}\t{1}\t{2}\t{3}\n'.format(
    ...:         i+1, key, cc.pop(), '; '.join(cc)))
```

This yields 1004, but you can see where the variation comes from. I suppose there was some typing errors or the like. I suppose, it is best to add your concepticon_ids and glosses to this file, as from there, it will be easier to go on. I assume, "further_glosses" will need to be added as a new item in the specific metadata for this file, but it is probably useful to retain it, as it may point to either inconsistencies in the original book, or in STEDT.

tresoldi:
The "concept_idx dot lang_idx" was actually quite clear, my fault was that at first (I should have checked tha data more carefully before proceeding) I thought the difference in number of concepts was due to this TB data being a subset of some larger Sino-Tibetan dataset (as if Sun had his/her own concepticon). I should have realized what the problem was when things like mismatched parentheses in glosses started popping up...

I have an idea on how to solve this: I will remap the number/id fields in the conceptlist using the indexes from Sun data and the most frequent English gloss for each concept, so we get to the 1004, and will fix the variation in the source first with hardcoded exception in lexibank (just until the list is ready for concepticon), and than correctly with a local `concepts.csv`. I'll try to do it today (a bit busier than usual, but it should be ready for you for tomorrow).

LinguList:
Sounds good, what I did so far with this kind of data was:

1 make the list with 1004 items (as I did, arbitrarily selecting one concept gloss, but frequency is a good decision)
2 when adding to lexibank and a gloss shows variation, I'd add a "concept_in_source" (or "gloss_in_source") to the forms.csv for that entry. This makes it most transparent, right? 

Another point, when adding to lexibank, please make sure to keep the first ID in the stedt datafile, as this can then be linked to STEDT via an URL. We'd really not want to loose that info, right?

tresoldi:
Updating almost everything we discussed. After all the cleaning and solving some problems in STEDT data, the number of concepts is actually down to 984. These are the ones found in the source, but it is clear that there are some duplicates there too (at least in the sense of concept for Concepticon).

The `NUMBER` field in the concept list (and, consequently, the last part of the ID one) now matches the first part of the `srcid` field in `ZMYCC.csv`. The `ENGLISH` field reports the most common gloss for each concept; the list of alias, which will be included in lexibank processing, is here:

[sun-1991-alias.txt](https://github.com/clld/concepticon-data/files/1506871/sun-1991-alias.txt)

LinguList:
Okay, so the list still has officially 1004 concepts, right, as in the book, but only for some 900+x there are valid entries? Since in my parse from today in the morning, I could find a gloss for all 1004 items, although there are some which are semantically close (but given their own ID in Sun 1991).

LinguList:
Sorry to tell you about another unspoken convention in Concepticon: if people claim they have 1004 concepts, we name their list Author-Year-1004, even if it turns out that the data does not have the concepts. But in fact, since STEDT lists data for all 1004 items, as my test showed, I wonder what happened with the missing ones? I'll check them in the book quickly.

LinguList:
Okay, sorry, but it turns out that there's something wrong here.

The first concept missing in the list is 31, which is "earth", according to my analysis that I uploaded above. Earth also occurs as concept 17, but here we have two different Chinese glosses:

Number | English (STEDT) | Sun 1991 | Translation (Mattis)
--- | --- | --- | ---
17 | earth | 天地 | "heaven and earth": the world
31 | earth | 土（干土) | "earth (dry earth)"

So I am afraid you'll have to follow my earlier strategy and use the numbers as the basis when hashing the data. All variants and aliases for the concepts should then only be assigned to one individual number, where they occur. Once this is done and linked, we can first merge, and I'll later see that we resolve the things through my project by manually checking with the book. Luckily, these are only 30 cases, but who knows how fuzzy our concept mappings may become in light of those translations provided by STEDT. We should also mention that in the conceptlists.tsv.

tresoldi:
Sorry, I am not being clear, hard day here. So, the file has 1255 different glosses, with 280 of them being aliases for other glosses. There 1004 different concept numbers (i.e., the first part of `srcid`), but many glosses map to two or more numbers (first example at hand, `bud` maps to both 226 and 788); the reason is not clear, many times a language has only one gloss->number, but sometimes two.

If you consider only unique glosses (e.g., only one `bud`, using the concept number with most occurrences in the Concepticon conceptlist), and assume that multiple words in the same language with the same gloss to be synonyms, then you only have 984 concepts. Maybe I am misunderstanding Concepticon design and should keep all the different numbers? Or for this dataset the concept number is actually more important and should take precedence over glosses? In that case, should I edit the glosses?

LinguList:
On the other point of concept mapping and the algorithm: I recommend to look at #382 and #389, as these are issues where we complain about the concept mapping algorithm. I am not sure whether you have read the actual algorithm in pyconcepticon/glosses.py, but what I think is smart about the algorithm is that we do not make an all-to-all comparison, but instead use all of the data we have assembled already, reduce it to its core components, and create a dictionary. This makes the algorithm theoretically linear in terms of the time it needs, and we can just apply the same principle to any other language for which we have linked data in Concepticon. What is bad is the similarity scores, as they follow a buggy logic. If you find time to look into that, or even have an idea how to improve it, this would be cool.

tresoldi:
Just commited a new version, now with 1004 concepts. There are some missing concepts from the new version that could be candidates for inclusion in Concepticon (like earth as planet, none of the available quite makes it, as some of the glosses clearly have it as "opposed to heaven"), but I decided not to work on that until we are sure that the list is correct (I hope it is now, sorry for all the versions).

As the for concept mapping algorithm, I do have some ideas, might take a look later and implement an alternative (can't be sure it will be worth unless I sit down and write the code). But let's finish this challenging concept list first. ;)

tresoldi:
A silly and quick test (no time for more serious coding): https://github.com/glottobank/lexibank-data/issues/344

LinguList:
Concept mapping is currently not our highest priority, but it could become, if more lists are being added, although for lists of 100 items, we usually have no real problems. But have a look at my comments in glottobank/lexibank-data#344, as I think it is important to keep in mind what we already have, and that concept mapping is way beyond string similarity.

concepticondata/concepticon.tsv
```diff
@@ -3034,3 +3034,31 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
 3040	DIFFERENT-SEX YOUNGER SIBLING	Kinship	A person of the different sex who is born after another person who has the same parents.	Person/Thing	
+3041	PATERNAL AUNT (FATHER'S OLDER SISTER)	Kinship	The older sister of one's father.	Person/Thing	
+3042	MATERNAL AUNT (MOTHER'S OLDER SISTER)	Kinship	The older sister of one's mother.	Person/Thing	
+3043	MATERNAL AUNT (MOTHER'S YOUNGER SISTER)	Kinship	The younger sister of one's mother.	Person/Thing	
+3044	BOVINE	Animals	Member of the Bovinae subfamily, mostly pointing to bulls, cows and oxes.	Person/Thing	
+3045	COPPER (COLOR)	Sense perception	Having a red to yellow metallic color that resembles the metal copper.	Property	
+3046	FODDER	Food and drink	Feed for livestock, especially chopped hay or straw.	Person/Thing	
+3047	HOT SEASON	Time	Period of the year associated with hot or warm weather.	Person/Thing	
+3048	MOUTH HARP	Modern world	Lamellophone instrument consisting of a tongue or reed attached to a frame, placed in the performer's mouth and plucked; Jew's harp, jaw harp.	Person/Thing	
+3049	OIL (FROM ANIMALS)	Food and drink	Liquid hydrophobic substance from animals which is used for cooking.	Person/Thing
```
LinguList: Oil from animals should be FAT (FROM ANIMALS), 1282, because you usually have this in a non-liquid form, and in the Chinese language for glossing, they don't distinguish between OIL and FAT, so please link to 1282

concepticondata/concepticon.tsv
```diff
@@ -3034,3 +3034,31 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3053	TIGHT	Basic actions and technology	Being not loose.	Property	
+3054	BEAR FRUIT	Agriculture and vegetation	To produce or yield fruits.	Action/Process	
+3055	SWINDLE	Law	To deceive or deprive by trickery and fraud; to cheat.	Action/Process	
+3056	ANIMAL (CLASSIFIER)	Quantity	Classifier for animals.	Classifier	
+3057	FLAIL	Basic actions and technology	A manual device used for beating or striking with, consisting of a free-swinging stick attached to a handle.	Person/Thing	
+3058	GRAZE	Animals	To feed on grasses and herbage in a field or on pastureland.	Action/Process	
+3059	MALARIA	The body	Infectious disease characterized by chills, fevers, and sweating, caused by a protozoan of the genus Plasmodium and transmitted by infected Anopheles mosquitos.	Person/Thing	
+3060	PHEASANT	Animals	Bird of the family Phasianidae.	Person/Thing	
+3061	PHLEGM	The body	Mucus secreted by the respiratory tract, particularly during respiratory infections, usually thick and sticky.	Person/Thing	
+3062	PIGEON	Animals	Bird of the family Columbidae.	Person/Thing
```
LinguList: we have DOVE there, 1853, so please link to this one (similar to eggpland / aubergine)

concepticondata/conceptlists.tsv
```diff
@@ -1,4 +1,5 @@
 ID	AUTHOR	YEAR	LIST_SUFFIX	ITEMS	TAGS	SOURCE_LANGUAGE	TARGET_LANGUAGE	URL	REFS	PDF	NOTE	PAGES	ALIAS
+Sun-1991-1004	Sun, Hongkai et al.	1991		1004	basic	English	Tibeto-Burman languages	http://stedt.berkeley.edu/~stedt-cgi/rootcanal.pl	Sun1991		This is a list of concepts generated from the entries listed in the STEDT database; there are many concepts overlapping, some of which are only used for a single doculect.
```
LinguList: I propose to rewrite the note as:

"This concept list originally served as a questionnaire for a large-scale investigation on Tibeto-Burman languages. The original questionnaire was in Chinese with no English translation. Later the [STEDT](http://stedt.berkeley.edu) project ([Matisoff 2015](:bib:Matisoff2015)) digitized the data, translating Chinese concept labels to English, but not listing the original Chinese forms. As a result, some concept labels are identical, although they are different in the Chinese version. We are still trying to add the original Chinese concept labels to this resource, but for the moment, we only link the STEDT version, occasionally adding Chinese concept labels, where we figure they are important to distinguish the original meaning."

I also suggest to not label this as "basic" but rather as "questionnonnaire" as we did with Buck and other lists, it is surely not a "basic" list.

The reference for Matisoff, I suggest to add:
```
@Book{Matisoff2015,
  Title                               = {{T}he {S}ino-{T}ibetan {E}tymological {D}ictionary and {T}hesaurus project},
  Address                             = {Berkeley},
  Author                              = {Matisoff, James A.},
  Publisher                           = {University of California},
  Year                                = {2015},

  Owner                               = {mattis},
  Timestamp                           = {2016.09.08}
}

```

concepticondata/concepticon.tsv
```diff
@@ -3034,3 +3034,31 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3038	SAME-SEX YOUNGER SIBLING	Kinship	A person of the same sex who is born after another person who has the same parents.	Person/Thing	
 3039	DIFFERENT-SEX OLDER SIBLING	Kinship	A person of the different sex who is born before another person who has the same parents.	Person/Thing	
 3040	DIFFERENT-SEX YOUNGER SIBLING	Kinship	A person of the different sex who is born after another person who has the same parents.	Person/Thing	
+3041	PATERNAL AUNT (FATHER'S OLDER SISTER)	Kinship	The older sister of one's father.	Person/Thing	
+3042	MATERNAL AUNT (MOTHER'S OLDER SISTER)	Kinship	The older sister of one's mother.	Person/Thing
```
LinguList: sorry to bother, but we'd need concept relations here:

3042 MATERNAL AUNT (MOTHER'S OLDER SISTER) broader 2693 MATERNAL AUNT (MOTHER'S SISTER)

concepticondata/concepticon.tsv
```diff
@@ -3034,3 +3034,31 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3038	SAME-SEX YOUNGER SIBLING	Kinship	A person of the same sex who is born after another person who has the same parents.	Person/Thing	
 3039	DIFFERENT-SEX OLDER SIBLING	Kinship	A person of the different sex who is born before another person who has the same parents.	Person/Thing	
 3040	DIFFERENT-SEX YOUNGER SIBLING	Kinship	A person of the different sex who is born after another person who has the same parents.	Person/Thing	
+3041	PATERNAL AUNT (FATHER'S OLDER SISTER)	Kinship	The older sister of one's father.	Person/Thing
```
LinguList: Concept relations:

3041	PATERNAL AUNT (FATHER'S OLDER SISTER)	broader 2691 PATERNAL AUNT (FATHER'S SISTER)

concepticondata/concepticon.tsv
```diff
@@ -3034,3 +3034,31 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3038	SAME-SEX YOUNGER SIBLING	Kinship	A person of the same sex who is born after another person who has the same parents.	Person/Thing	
 3039	DIFFERENT-SEX OLDER SIBLING	Kinship	A person of the different sex who is born before another person who has the same parents.	Person/Thing	
 3040	DIFFERENT-SEX YOUNGER SIBLING	Kinship	A person of the different sex who is born after another person who has the same parents.	Person/Thing	
+3041	PATERNAL AUNT (FATHER'S OLDER SISTER)	Kinship	The older sister of one's father.	Person/Thing	
+3042	MATERNAL AUNT (MOTHER'S OLDER SISTER)	Kinship	The older sister of one's mother.	Person/Thing	
+3043	MATERNAL AUNT (MOTHER'S YOUNGER SISTER)	Kinship	The younger sister of one's mother.	Person/Thing
```
LinguList: concept relation:

3043 MATERNAL AUNT (MOTHER'S YOUNGER SISTER) broader 2693 MATERNAL AUNT (MOTHER'S SISTER)

concepticondata/concepticon.tsv
tresoldi: Oh, that was actually a left-over from the "same concept with different glosses", it is not used in Sun's list anymore.

LinguList:
Thanks. I just received the Chinese list from Doug Cooper. I will probably update this later on, but you already did an amazing job here.

tresoldi:
Thanks! :)

2017-11-28 10:10 GMT-02:00 Johann-Mattis List <notifications@github.com>:

PR #407, opened 2017-11-28
Added (original) Chinese glosses to Sun-1991-1004

tresoldi:


LinguList:
Excellent, many thanks! We should also not forget to thank Doug Cooper in our contributions.md

PR #410, opened 2017-11-30
Reviewing Sun's list after Chinese glosses.

tresoldi:
I've reviewed all concepts now that we have the original Chinese glosses. I've corrected many mappings and I'm proposing quite a few new concepts which seemed general enough to be worth inclusion (and should come up again in new lists). It took a while but was actually fun, even taught me some tidbits of Chinese.

Among those that I considered including but didn't:

- 亩 (963) might be added as HECTARE
- 青稞 (189) might be added as HIGHLAND BARLEY
- it was not clear to me, but maybe 编 (655) would justify a general "ORGANIZE OR PUT IN ORDER" concept, instead of BRAID?
- would 裂开 (759) justify a new "SPLIT OPEN" concept? It seems that none of the various "CUT" is specific enough
- would 庄稼 (185) justify a new "CROP(S)" concept (i.e., kind of plant)? Or it is more in the sense of (also missing from Concepticon) "YIELD" (i.e., amount produced)?
- would 剁 (635) justify a new "MINCE" concept? as for 裂开, it seems the none of the "CUT"s is specific enough
- 旋转 (763) might be added as "ROTATE OR SPIN (SOMETHING)"


LinguList:
> 亩 (963) might be added as HECTARE

this is a very specific measure, less than a HECTARE, so I'd suggest, if adding at all "MU (CHINESE MEASURE FOR AREA)" or similar.

>  青稞 (189) might be added as HIGHLAND BARLEY

yes, I think it's good!

>    it was not clear to me, but maybe 编 (655) would justify a general "ORGANIZE OR PUT IN ORDER" concept, instead of BRAID?

No, as the word in brackets means explicitly "plait" (Zopf in German), so its: to braid a plait (my english is bad in these terms). The verbal reading.


>    would 裂开 (759) justify a new "SPLIT OPEN" concept? It seems that none of the various "CUT" is specific enough

Yes, I agree, it is close to "aufbrechen" in German, so could also mean "cut open".  CUT-terms are a nightmare, we needed to announce a PhD position for this if we had the means

>    would 庄稼 (185) justify a new "CROP(S)" concept (i.e., kind of plant)? Or it is more in the sense of (also missing from Concepticon) "YIELD" (i.e., amount produced)?

CROPS seems good!

>    would 剁 (635) justify a new "MINCE" concept? as for 裂开, it seems the none of the "CUT"s is specific enough

yes, that sounds actually very nice, but I'd make sure it's a verbal reading, so MINCE (SOMETHING)

>    旋转 (763) might be added as "ROTATE OR SPIN (SOMETHING)"

it is more like a reflexive reading, at least judging on the examples I looked into so far. Maybbe @laiyunfan could help us here.

tresoldi:
>>     亩 (963) might be added as HECTARE
> 
> this is a very specific measure, less than a HECTARE, so I'd suggest, if adding at all "MU (CHINESE MEASURE FOR AREA)" or similar.

I'd say it is too specific, but at the same time it is likely to come up again and we already have analog units such as catty 斤(三斤).

>> would 裂开 (759) justify a new "SPLIT OPEN" concept? It seems that none of the various "CUT" is specific enough

> Yes, I agree, it is close to "aufbrechen" in German, so could also mean "cut open". CUT-terms are a nightmare, we needed to announce a PhD position for this if we had the means

I've added both SPLIT OPEN and MINCE (SOMETHING), but I didn't add relations as I was not sure (to be pedant, they are not hyponyms but troponyms). On the 'cut' terms, I was actually thinking of reviewing the troponyms of "cut" starting from wordnet (after these dataset are finished, of course!), I'll let you know.

I think I'll start the work on Sun's dataset after that, it should be stable enough (but please review it, Chinese speakers!).

BTW, this will be in conflict with @chrzyki work due to Concept IDs, sorry. I don't see allocating a concept space as a solution, maybe the best thing is really to solve conflicts as they come up.

LinguList:
as @chrzyki wanted to go through this with me anyway, we first merge this one and then look for a solution how to quickly merge the other list. Basically, one can do this with a little helper script.

LinguList:
We'll work on this later in our project on ST languages, so no worries if there are some problems now (I am also not extremely profitient in all of these terms).

PR #414, opened 2017-12-08
Updated to the glosses used in the thesis; mapped concepts already av…

tresoldi:
…ailable; added new concept. For work on https://github.com/glottobank/lexibank-data/issues/350


concepticondata/conceptlists/SatterthwaitePhillips-2011-423.tsv
```diff
@@ -23,118 +23,118 @@ SatterthwaitePhillips-2011-423-21	665	BEE	21	171	bee	蜜蜂	29
...
+SatterthwaitePhillips-2011-423-40	1394	BONE	40	174	bone	骨头	126
 SatterthwaitePhillips-2011-423-41	994	BOW	41	175	bow	弓	165
 SatterthwaitePhillips-2011-423-42	1530	BOWL	42	175	bowl	碗	147
 SatterthwaitePhillips-2011-423-43	1366	BOY	43	175	boy	男孩 [B = 男青年]	253
-SatterthwaitePhillips-2011-423-44	1583	BRAIN	44	175	brain>>	脑子 [B= 脑髓]	119
+SatterthwaitePhillips-2011-423-44	1583	BRAIN	44	175	brain	脑子 [B= 脑髓]	119
 SatterthwaitePhillips-2011-423-45	1531	BRANCH	45	175	branch	树枝	53
-SatterthwaitePhillips-2011-423-46	680	BREAK (DESTROY OR GET DESTROYED)	46	176	break	弄坏 [B = pohuai]	317
-SatterthwaitePhillips-2011-423-47	2558	BREAK (CLEAVE)	47	176	break （branch）	断（小枝）	361
+SatterthwaitePhillips-2011-423-46	680	BREAK (DESTROY OR GET DESTROYED)	46	176	break (damage)	弄坏 [B = pohuai]	317
```
LinguList: I just realized that this is actually an instance of BREAK (SOMETHING) or even better DESTROY (SOMETHING). The Chinese word *could* be used intransitively, but usually, they'd add a passive marker to it, so this is really the action of destroying something. Following the English gloss BREAK (SOMETHING) should suffice.

concepticondata/conceptlists/SatterthwaitePhillips-2011-423.tsv
```diff
@@ -12,7 +12,7 @@ SatterthwaitePhillips-2011-423-10	1927	BAMBOO	10	168	bamboo	竹子	60
 SatterthwaitePhillips-2011-423-11	868	BANANA	11	169	banana	香蕉；芭蕉 [B = 香蕉]	138
 SatterthwaitePhillips-2011-423-12	1204	BARK	12	169	bark	树皮	55
 SatterthwaitePhillips-2011-423-13	932	BARLEY	13	169	barley	大麦	349
-SatterthwaitePhillips-2011-423-14	1793	BAT	14	169	bat>>	蝙蝠	44
+SatterthwaitePhillips-2011-423-14	1793	BAT	14	169	bat	蝙蝠	44
```
LinguList: You delete asterisks and >> characters, but note that the concepticon gives an explicit explanation for that in our conceptlists.tsv:

> This is an expanded meaning list combining parts of the lists proposed by [Matisoff 1987](:ref:Matisoff-1978-200) and the well known lists of [Swadesh (1952](:ref:Swadesh-1952-200) and [1955)](:ref:Swadesh-1955-100). The author was so friendly to provide us with an excel spreadsheet containing the data. Note that although the author mentions in his thesis on page 44 that a total of 410 words (meanings) were used to collect the data, we find as many as 423 words, including some duplicate entries, but also some entries which are obviously near synonyms but were probably split into different semantic slots by the author (who himself marks seven additional entries as not belonging to the original concept). Entries marked with an asterisk seem to reflect items which also occur on traditional concept lists and were given in the spreadsheet version of the data. Entries ending on the percentage character are those marked as later added by the author in the original source. Why some entries are marked with two greater-then characters in the spreadsheet data of the author is not entirely clear to us, but we kept it in order to provide all information that is available.

I think we should keep this version that reflects both spreadsheet and the data in the dissertation, as we are in the comfortable situation of having both, and it is properly explained.

concepticondata/conceptlists/SatterthwaitePhillips-2011-423.tsv
tresoldi: I'll note that, I was going to review the concept list later.

concepticondata/conceptlists/SatterthwaitePhillips-2011-423.tsv
tresoldi: Oh, now I see the problem (and my fault for not reading the note on conceptlists.tsv, sorry).

Doing a kind of archaeology of the data, my hypothesis was that the Excel files were either an intermediate stage or, more likely, a "branch" of the data converted from some intermediate representation (likely a CSV with X-SAMPA notation, which explains some artifacts); the data actually used in the study and provided at the end of the document itself (which is what is open and available) was built on this same root, with corrections, including the removal of languages with no entries (the author probably intended to include them but was unable to), and the combination of multiple sources/columns for each language into a single one. This also explains the difference in the number of concepts (I would bet I can tell which ones they are, because the PDF ends up preserving the order of edition of the file, and they likely are the 12 that were problematic to parse, especially because their Hanzi are not Unicode characters but pasted images -- amazing how much you can tell on an author operation ;) ).

I will revert the changes later, sorry for that. This brings the question on what to do:

- Should I add a new conceptlist?
- Should I use this conceptlist as it is (with the minor corrections) and treat asterisks and etc. locally, with my own concepts.csv?
- Should I add a new column to this conceptlist, reflecting the "clean" version of the gloss?

tresoldi:
Minor corrections, changes and review; the original ENGLISH column (the header of the Excel spreadsheet) was renamed (see discussion on lexibank issue for dataset satterwaitephillpstb, https://github.com/glottobank/lexibank-data/issues/350 )

PR #415, opened 2017-12-11
Added Gampe-2017-48. Fixed Berry to Strawberry in Luniewska-2016-299.

chrzyki:


chrzyki:
Will fix this if/when @tresoldi's request is merged.

concepticondata/conceptlists.tsv
```diff
@@ -211,4 +211,5 @@ Voorhoeve-1971-125	Voorhoeve, C.L.	1971		125	basic,areal	English	New Guinea		Voo
...
+Gampe-2017-48	Gampe, Anja AND Kurthen, Ira AND Daum, Moritz M.	2017		48	acquisition	English	Global	http://journals.sagepub.com/doi/full/10.1177/0142723717736450	Gampe2017
```
LinguList: Wait, sorry for not mentioning earlier, but we'll need a little story here, explaining, for example, that Gampe's list is partially building on Luniewska, if I'm not mistaken.

LinguList: Yes, one more thing: this should also include some explanation of the new columns in the data. I am not sure, would have to re-read the paper, but I think it'll be some reaction times for the mean_concept_vocab? Maybe, changing the name of the column to be more telling, might also not be a bad idea.

chrzyki: Alright, will write notes & rename columns.

LinguList:
I guess, as long as those are explained in the conceptlists.tsv, it should be okay (so far, this was my practice: if new columns are introduced, explain their role in the conceptlists.tsv).

tresoldi:
I took the liberty of solving the conflicts, as I am going to commit some minor changes to SatterthwaitePhillips-2011-423 and it might not allow fast-forwarding anymore. Tests are passing (I checked before doing so), sorry if it was too bold of a move on my side!

chrzyki:
Ah, awesome! Thank you very much & no problem at all! Will talk about the columns today with @LinguList and then merge this.

chrzyki:
Thanks for the help @tresoldi and @LinguList. Merging this.

PR #418, opened 2018-01-02
fixing the new clldutils/csvw-dependencies

LinguList:


LinguList:
I'll just merge this right away, so we can go on with this later.

PR #425, opened 2018-01-10
Documentation Improvements

chrzyki:
Just a little something to improve the documentation of commands that are directly available to end-users.

### DIFF
--- a/pyconcepticon/commands.py
+++ b/pyconcepticon/commands.py
@@ -102,8 +102,11 @@ def link(args):
 
 @command()
 def mergers(args):
-    """Return all merged concepts in a concept list (checks for local concept
+    """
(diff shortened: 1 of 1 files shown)


chrzyki:
Ah, alright! Will change this then! Thanks for pointing this out.

xflr6:
Here are some examples (I prefer the former markup variant):
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

There is also [pep-257](https://www.python.org/dev/peps/pep-0257/)

LinguList:
My favorit is numpy, as we use it also for lingpy.

chrzyki:
And another attempt.

PR #439, opened 2018-03-01
tests ported to pytest

xrotwang:
Please check out. Most things should have become simpler - if anything.


LinguList:
looks fine to me, but I don't understand why the tests think there should be a repo site-packages/concepticondata. Is that due to the placement of the concepticondata folder?

PR #451, opened 2018-03-07
added editor info

xrotwang:
I think it makes sense to also list the editors. This would allow grabbing this data as well when initializing the web app. It also makes tracking editorship over time more transparent.


LinguList:
First I was confused by the 2013, but you're right: we started with the app already by then... Feel free to merge.

PR #452, opened 2018-03-07
adding Davies and Comrie 1985

SimonGreenhill:
A linguistic survey of Upper Yuat (another TNG.org list).

Some questions:
1. This list has verbs as "he bites" which I've mapped to "BITE". Is this ok?

2. "he gives it to me"  is mapped to "GIVE".  Should this be given its own new concept (give transitive?)

3. "axe (stone)" -> I've put this in AXE but perhaps could be a new concept set (STONE AXE)?

PR #453, opened 2018-03-07
Add some more tests to get coverage up.

SimonGreenhill:
Sorry -- it looks like this has merged in the changes from #452 as well.


LinguList:
Yes, ideally, you maybe introduce the improved note (if you follow my suggestion) here, resolve the conflicts, and we merge?

I'd propose this for the Comrie list:

> A 116 item wordlist survey of the Upper Yuat region, based on a variant of survey lists by the summer institutes of linguistics (currently not in Concepticon yet).

Issue #436, opened 2018-02-28
Kibrik-2012-122: Early acquired verbs


LinguList:
This list is interesting, as it contains verbs which are early acquired by children. Based on an earlier psycholinguistic study. It is also one of the rare cases where typological studies in contrast to linguistic studies are emphasized.

* https://www.degruyter.com/view/j/ling.2012.50.issue-3/ling-2012-0017/ling-2012-0017.xml

chrzyki:
Prepared a first draft of this list, just something to get started: 
[Kibrik-2012-122.tsv.zip](https://github.com/clld/concepticon-data/files/1766700/Kibrik-2012-122.tsv.zip)

Also, bibliographic information:

```
@article{kibrik2012,
  title={Toward a typology of verbal lexical systems: A case study in Northern Athabaskan},
  author={Kibrik, Andrej A.},
  journal={Linguistics},
  volume={50},
  number={3},
  pages={495--532},
  year={2012},
  doi={10.1515/ling-2012-0017},
  url={https://doi.org/10.1515/ling-2012-0017}
}
```

chrzyki:
Might as well take this, then. :) Any lists/references to keep in mind for this one?

LinguList:
I haven't had time to really read this paper. What I think is important is to trace a bit the history (they take it from age-of-acquisition, do we find the original list, etc?). We want to re-use the data for a study on verbal derivation, and I'd suggest you add @laiyunfan as a reviewer of the mappings, as this list is in general interest of the CALC people...

chrzyki:
Question: They have 'leave' in possession/change of state/exchange and possession of objects. I take it this is the Russian оставить, what would be good equivalent in Concepticon (if already available)?

LinguList:
So German "da lassen". We don't have it, I'm afraid...

chrzyki:
Thanks!

chrzyki:
@LinguList Have added a couple of comments in the commit in the branch that is going to be PRed when ready. :)

Thanks for your help!

https://github.com/clld/concepticon-data/commit/646aab3bd8758b58d1324f723a60c9fa770c14b6

LinguList:
My next PR will add Russian words. I list here the missing definitions in concepticon that should/could be added:

1. switch on (turn on) (anmachen, also das Licht)
2. switch off (turn off) (ausmachen, also das Licht)
3. leave (something for someone) (im Sinne von "dalassen" oder "übriglassen")
[...]

PR #458, opened 2018-03-23
Kibrik-2012-122 work in progress

chrzyki:
A PR for a better work-in-progress documentation/workflow. See also discussion here: https://github.com/clld/concepticon-data/commit/646aab3bd8758b58d1324f723a60c9fa770c14b6

#436


concepticondata/conceptlists.tsv
```diff
@@ -215,5 +215,7 @@ Luniewska-2016-299	Łuniewska, Magdalena AND Haman, Ewa AND Armon-Lotem, Sharon
...
-Davies-1985-116	Davies, John AND Comrie, Bernard	1985		116	basic,areal	English	New Guinea	http://sealang.net/archives/pl/pdf/PL-A63.275.pdf	Davies1985		A 116 item wordlist survey of the Upper Yuat region, based on a variant of survey lists by the Summer Institute of Linguistics (currently not in Concepticon yet).
+Davies-1985-116	Davies, John AND Comrie, Bernard	1985		116	basic,areal	English	New Guinea	http://sealang.net/archives/pl/pdf/PL-A63.275.pdf	Davies1985		A 116 item wordlist survey of the Upper Yuat region, based on a variant of survey lists by the Summer Institute of Linguistics (currently not in Concepticon yet).		
 Bengtson-2017-12	Bengtson, John D	2017		12	basic	English	Global		Bengtson2017		Twelve basic word meanings from the [Swadesh list of 100 items](:ref:Swadesh-1955-100).		
+Kibrik-2012-122	Kibrik, Andrej D.	2012		122	basic	English	Global		Kibrik2012		A list of 122 verbs (based on Tomasello 1992) used for comparing verbal lexical meaning construction across (primarily) Koyukon, Russian, and English.
```
LinguList: Tomasello 1992 should either not be mentioned, or included in the references and linked via ```[xxx](ref)```, and we should file an issue that we get hold of this as well, as it may lighten up problems...

concepticondata/conceptlists/Kibrik-2012-122.tsv
```diff
@@ -0,0 +1,123 @@
...
+102	Kibrik-2012-122-102	swing	Motion	Activities	Activities not involving objects	2871	SWING (MOVEMENT)
+103	Kibrik-2012-122-103	see	Perception	Activities	Activities not involving objects	1409	SEE
+104	Kibrik-2012-122-104	look	Perception	Activities	Activities not involving objects	1819	LOOK
+105	Kibrik-2012-122-105	watch	Perception	Activities	Activities not involving objects	1650	WATCH
+106	Kibrik-2012-122-106	taste	Perception	Activities	Activities not involving objects	21	TASTE (SOMETHING)
+107	Kibrik-2012-122-107	smell	Perception	Activities	Activities not involving objects	1586	SMELL (PERCEIVE)
+108	Kibrik-2012-122-108	listen	Perception	Activities	Activities not involving objects	1608	LISTEN
+109	Kibrik-2012-122-109	feel	Perception	Activities	Activities not involving objects	2559	FEEL (TACTUALLY)
+110	Kibrik-2012-122-110	hurt	Physical impact	Activities	Activities not involving objects	2411	HURT (SOMEBODY)
+111	Kibrik-2012-122-111	hurt	Physical condition	Activities	Activities not involving objects	1129	PAINFUL
```
LinguList: painful is obviously not right here, as it's not an action...

chrzyki: Ok, I was a little bit confused by this. The article says hurt INTR with болеть. I read this like у меня́ боли́т зуб or something similar. Would HURT (SENSE PAIN)/2561 then be better suited?

LinguList: Leave unlinked for now. Without the Russian entries, it is just extremely difficult to read the glosses. We may also have to dig into the earlier source from the 1990ies to see what they originally meant. We should try and get this book via Fernleihe.

concepticondata/conceptlists.tsv
```diff
@@ -217,5 +217,5 @@ Auracher-2017-16	Auracher, Jan	2017		16	acquisition	English	Global	https://www.n
...
 Davies-1985-116	Davies, John AND Comrie, Bernard	1985		116	basic,areal	English	New Guinea	http://sealang.net/archives/pl/pdf/PL-A63.275.pdf	Davies1985		A 116 item wordlist survey of the Upper Yuat region, based on a variant of survey lists by the Summer Institute of Linguistics (currently not in Concepticon yet).		
 Bengtson-2017-12	Bengtson, John D	2017		12	basic	English	Global		Bengtson2017		Twelve basic word meanings from the [Swadesh list of 100 items](:ref:Swadesh-1955-100).		
-Kibrik-2012-122	Kibrik, Andrej D.	2012		122	basic	English	Global		Kibrik2012		A list of 122 verbs (based on Tomasello 1992) used for comparing verbal lexical meaning construction across (primarily) Koyukon, Russian, and English.		
+Kibrik-2012-122	Kibrik, Andrej D.	2012		122	basic	English	Global		Kibrik2012		A list of 122 verbs used for comparing verbal lexical meaning construction across (primarily) Koyukon, Russian, and English.
```
LinguList: Given that we'll definitely NEED Tomasello to further dig intto the origin of htis list, I'd actually prefer to have it included already as a reference, so that it is easier to add it later to concepticon (do we already have an issue for Tomasello?)

chrzyki: Ok, will add. I've got the list here with me (the appendix of the book is something like 90 pages though, would have to check where exactly the  verbs come from).

LinguList: once this is done, feel free to merge, I'll add the Russian concepts later then. We'll then also need a strategy to add Tomasello...

PR #471, opened 2018-04-18
adding Kolipakam et al.

SimonGreenhill:
There's no obvious page numbers or image of the wordlist as it's contained in the supplementary files in:

* drav_cov_est_ucln_yule.xml - BEAST 2 xml file.
* SI_robustness_cognate_coding.zip - a zip file with many things.


LinguList:
Yes, and as far as I can tell: the mapping was already done, right?

PR #472, opened 2018-04-19
Replace instanceof relations with narrower

### LINKED ISSUE #468: instanceof relation is strange
The `instanceof` relation is used in four places, eg. https://github.com/clld/concepticon-data/blob/master/concepticondata/conceptrelations.tsv#L276

However, the concept MID-AUTUMN FESTIVAL is just a very specific sub-concept of FESTIVAL, and the same holds for the other concepts related through `instanceof`, so I think you should either use `narrower` or define those terms.

My naïve understanding of `instanceof` in the absence of defined semantics is that it applies to eg. the relation between FOUR and NUMBER, or maybe to WORLD → PLACE (POSITION), but overall very rarely in the database, because generally a concept is a class, not an instance.

Anaphory:
In the general use in ontologies, the `instanceof` relation is between objects and their classes. This is not really a relationship that applies to any two concepts, compare #468.

Fixes #468.


LinguList:
sorry, we can't have this, I am afraid, as it will map "Autom Festival" to "Festival" if we compare the two things, and a special KIND of Festival should NOT be confused with a festival in our algorithm that makes the fuzzy mapping via union and set! So either change the relation, or remove the relation, or leave it as is. But never define an exemplar of something as something narrower than the category. This goes against our policy of union and set comparisons. Sorry, should've noted that earlier...

tresoldi:
Maybe we should use this one as an example in the documentation? I always find it better to justify what might look strange with concrete examples (think about the "dull/stupid" example -- I used it on Tuesday and people readily realized why Concepticon was needed).Em 19 de abr de 2018 6:05 PM, Johann-Mattis List <notifications@github.com> escreveu:sorry, we can't have this, I am afraid, as it will map "Autom Festival" to "Festival" if we compare the two things, and a special KIND of Festival should NOT be confused with a festival in our algorithm that makes the fuzzy mapping via union and set! So either change the relation, or remove the relation, or leave it as is. But never define an exemplar of something as something narrower than the category. This goes against our policy of union and set comparisons. Sorry, should've noted that earlier...

LinguList:
Yes, it needs to be there, you are right, @tresoldi !

Anaphory:
In that case, I think you need to fix a few other cases of this and explain when something becomes so much `narrower` that it's `instanceof`.

1181	PALM TREE	narrower	2372	SAGO PALM
2456	LARGE WILD HERBIVORE	narrower	1936	DEER
2461	WOLF-LIKE ANIMAL	narrower	2460	DINGO
2138	LEGENDARY CREATURE	narrower	1037	ELF
(and many other connections on this level)

This may be connected to my suggestion of ‘cultural replacements’. It looks like you have the following four relationships:

 1. Item from one word list where another word list felt the need to be semantically more precise, because  two concepts co-lexified in the gloss language are not co-lexified in target languages, so it got noticed. (This is very apparent in most XXX OR YYY concepts and probably holds for most kinship terms, but also holds eg. for our FEMALE SARONG.) – These items should be merged-and-marked when comparing concept lists, as `narrower` is now.
 2. Item from one word list does have a functional equivalent in a culture another word list is targeting, which is however different in significant features so that it's not just a narrowing of concept. I assume this holds for WOLF/DINGO/WILD DOG under WOLF-LIKE ANIMAL which I found above. From our word list, it's the case eg. for LOCUST/RICE EAR BUG (under *FIELD CROP PEST INSECT?) – These items should be merged-and-marked to the overarching concept like `narrower` does now, but also if two lists contain cultural equivalent siblings, eg. merging DINGO and WILD DOG under WOLF-LIKE ANIMAL.
 3. Very specific narrowings of broad general concepts, which might all be met in the same culture. Like the ELF, FAIRY → LEGENDARY CREATURE, AUTUMN FESTIVAL, DRAGON BOAT FESTIVAL → FESTIVAL and SAGO PALM → PALM TREE mappings. – These should be marked as connected (in particular given that they might not be in many English-based ontologies you might infer other relations from), but not merged when comparing word lists.
 4. Items that are ontologically somewhat narrower, but where there is no reason to assume confusion. You probably don't have too many of these in the database, because they are not really relevant for this application and better represented in external ontologies you link to. The examples I found could arguably be instances of cases 1 to 3 (eg. WALK → GO, HEADBAND → HEADGEAR, DEER → LARGE HERBIVORE). – These may be marked as connected (although that can be offloaded to automatic inference from elsewhere)

I would keep `narrower` for category 4 and change it to something like `disambiguates` for category 1. I would fix all examples of category 3 to `instanceof` and make a note of what it means in a readme file somewhere, or think of a different gloss. I would use something like `cultural-realization-of` (maybe something less clunky) for category 2 and make it behave like category 1 for now.

LinguList:
please don't check only the glosses, but also the reflexes. In fact: we introduced the dingo-thing because they give "wolf/dingo" or things like that, so to keep their fuzzy over-arching concept, we introduced the other categories. similar with mouse or rat -> muroid (mouse or rat), where we don't have exemplar relationships. Sago palm is probably something in need of correction, you're right, but again: it depends on definitions.

I don't have the time nor the ability to go into all details about ontologies. When linking to wordnet, this is anyway covered, there you have ontological categories and linkins far beyond waht we can do. Broader Narrower serves initially only one purpose: accounting for "arm/hand" vs. "arm", vs. "hand", and I think we should leave it at that. It was based on @xrotwang's suggestion, who had additional reasons for using the terms, as they are common in ontologies, etc.

As to all the rest: if we discuss this in total now, I'd need to have a) personel or people willing to work on this from beginning to end (going through all of the 3000 concept sets), and b) complete compliance with our pragmatic and admittedly at times very simple demands (e.g., that 200 + 100 = 207 in Swadesh terms, to account for people's intuition).

So please don't understand me wrong: I deeply appreciate the enthusiasm, but I am afraid we are beginning a discussion here that we can never bring to the end. Maybe deleting all my non-narrower/broader relations is even better (correcting for exemplar cases like "sago palm"), as it will make clear that we don't even try to go deeper (because we lack the capacities). If there is the willingness to make this a more proper ontology, we need time commitment, or apply for funding, etc., to make this happen, as nobody of us will have enough time to work on this full-time, and we'll need at least one person full-time for setting this up for one to two years (also reading existing literature, etc.), to make this all happen.

PR #477, opened 2018-04-20
Birchall fix

SimonGreenhill:


xrotwang:
We did have a discussion on whether or not we want to allow changing identifiers (the `Birchall-2016-126` in this case) - although I can't find it right now (@LinguList do you know?).
I think we wanted to keep a list of errata, basically to create redirects in the web app, right?

xrotwang:
Ah, we have this: https://github.com/clld/concepticon-data/blob/master/concepticondata/retired.json
So possibly this list should also include other identifier categories, like `ConceptList`.

LinguList:
yes, will need to run through that thingy of retired. I also wonder how it's possible to have 1 merger now that the only merger we had before (mother and "mother ") is now deleted?

xrotwang:
Where do you see the merger thing? Did you re-create the READMEs?

concepticondata/conceptlists/README.md
```diff
@@ -17,7 +17,7 @@
 | [Bennet-1976-100](Bennet-1976-100.tsv) | 100 | 100 | 0 |
 | [Bennet-1976-50a](Bennet-1976-50a.tsv) | 50 | 100 | 0 |
 | [Bennet-1976-50b](Bennet-1976-50b.tsv) | 50 | 100 | 0 |
-| [Birchall-2016-126](Birchall-2016-126.tsv) | 126 | 100 | 1 |
+| [Birchall-2016-125](Birchall-2016-125.tsv) | 125 | 100 | 1 |
```
LinguList: @xrotwang, the merger is in this line, but it should not be there, as the former merger was MOTHER.

SimonGreenhill:
sorry -- I hand-edited README.md, and didn't update that column

xrotwang:
Ok, time to write the tutorial :)

SimonGreenhill:
Ok, added to retired.json and reran `concepticon stats` (the latter is not showing up as a diff, as the hand-changes I made earlier capture that).

concepticondata/retired.json
```diff
@@ -639,6 +639,11 @@
       "id": "Galicio-2015-28-28",
       "comment": "corrected typo",
       "replacement": "Galucio-2015-28-28"
+    },
+    {
```
LinguList: You'll have to add replacments for each of the entries, I'm afraid, as the 2016-125 is now the new ID :(

SimonGreenhill:
Ok, I'm adding an entry for each renumbering as:

```JSON
{
    "id": "Birchall-2016-126-29",
    "comment": "removed duplicate item in wordlist",
    "replacement": "Birchall-2016-125-28"
},
```

...but what should I do for the actual deleted item? (Birchall-2016-126-19): 

```JSON
{
    "id": "Birchall-2016-126-19",
    "comment": "removed duplicate item in wordlist",
    "replacement": none  # OR? 
},

```

LinguList:
since it's duplicate, why not link to the source of the duplicate, i.e., its identical entry?

LinguList:
In fact, @xrotwang uses these to create a forward-link to the new entry if you call for the olden one. So most straightforward approach would be to link this to the new unified entry now, I guess.

SimonGreenhill:
how's this?

SimonGreenhill:
Phew!

LinguList:
Just discussed with @MacyL, who said she was going "through hell" with the Huang-1992-1820 (!!!) mapping: we need T-shirts with a print on them, saying "no matter what you link now" (front) "you'll never walk alone: Concepticon" (back).

MacyL:
From someone is "going through hell".... 
front
back

PR #494, opened 2018-08-08
Update

Schweikhard:
I added some new concepts (mainly hyperconcepts) and corrected some links and some relations. Will continue with these kinds of edits.


concepticondata/conceptrelations.tsv
```diff
@@ -452,12 +451,12 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
 2672	AUNT OR UNCLE	narrower	1327	UNCLE
 2683	EAGLE OR HAWK	narrower	1905	EAGLE
 2683	EAGLE OR HAWK	narrower	307	HAWK
-2686	CRUSTACEANS (PRAWNS OR SHRIMP)	narrower	238	SHRIMP
-2686	CRUSTACEANS (PRAWNS OR SHRIMP)	narrower	2349	PRAWNS
+2686	CRUSTACEANS (PRAWNS OR SHRIMP)	instanceof	238	SHRIMP
```
LinguList: "instanceof" is not handled by our internal algorithms. So please revert to original relation. this guarantees that we can assemble words for SHRIMP and PRAWNS, even if they are different. This is crucial for our concept list comparison algorithms and should be kept.

Same applies to goat/sheep. Watermelon is fine.

concepticondata/conceptrelations.tsv
```diff
@@ -530,3 +529,8 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
 3024	BRIDE PRICE	similar	3168	DOWRY
 3110	CLAP	similar	3154	SLAP
 3171	RINGWORM	similar	3173	TINEA
+2678	HAIR OR FUR	narrower	189	HAIR (BODY)
+3182	MARRIAGE OR WEDDING	narrower	1259	WEDDING
+3182	MARRIAGE OR WEDDING	narrower	2432	MARRIAGE
+3183	BREATHE OR BREATH	narrower	1407	BREATHE
+3183	BREATHE OR BREATH	narrower	870	BREATH
```
LinguList: I am missing the new relations for "life/living" and "dead".

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,9 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3179	CORAL ROCK	The physical world	Limestone created from the skeleton of reef-forming stony corals (Scleractinia), displaying their growth patterns.	Person/Thing	
 3180	SHOOT (WITH SLINGSHOT)	Warfare and hunting	Use a rubber-based projectile weapon to propel projectiles against a target.	Action/Process	
 3181	MUSTARD	Food and drink	It can be a yellow sauce with a strong taste or a plant which is used to make mustard sauce.	Person/Thing	
+3182	MARRIAGE OR WEDDING	Kinship	The ceremony of marrying someone, or the state of being married	Person/Thing	
+3183	BREATHE OR BREATH	The body	The act of breathing or the air that goes into one's lungs	Other	
+3184	BE ALIVE OR LIFE	The body	The act of being alive or the life itself	Other	
+3185	DIE OR BE DEAD OR KILL	The body	To cease to live, or to be not alive, or to cause the end of a life	Other
```
LinguList: we have the practice of ordering "or" glosses alphabetically. So "be dead or die or kill" might be more approppriate. If you add this, don't forget to add the relation to "die" proper.

concepticondata/conceptrelations.tsv
Schweikhard: You are right, I am going to add those.

concepticondata/concepticon.tsv
Schweikhard: Good to know. I assumed the ordering was by word class (first verbs then nouns).

concepticondata/conceptrelations.tsv
Schweikhard: So what exactly are the rules for when to use which relation? "Narrower" when there are only 2 or 3 possible hyperonyms, "instance of" when there is a basically unlimited amount?

concepticondata/concepticon.tsv
Schweikhard: Also, then the already existing concept of "die or be dead" needs to be renamed as well.

LinguList: We describe only ONE relation (narrower/broader) for the time being. Ths is the one we mention in our paper. The other relations were experimental and added by me for testing. But one could expand tham ad infinitum, and we had other things to do, so they are not really developed...

LinguList: sorry: this answer was meant to answer your comment above ...

Schweikhard: NP. Regarding the ordering, it's about 20 hyperconcepts that are not in alphabetic order and need to be renamed then.
In some few cases we might to retain the non-alphabetic order, though (e.g. he or she or it).

LinguList: Yes, there are some historical cases. I just started some time ago, when adding new concepts, to adopt for this procedure, as "foot or leg" etc. is then easier to remember, as we often have "arm or hand" vs. "leg or foot" etd. in all orders. esthetically, I prefer "die or be dead", as this is more common. we could just ignore this here. 

But I'd be inclined to ask you to add "die or be dead" as a normal concept set first, and then make the triple-one, as this is easier for future mapping as well (we have to re-map a lot of lists still, where this is not consequent).

I leave the decision of ordering to you.

Schweikhard:
I did the changes requested and some more (didn't do most of the alphabetic renamings yet).

 I also added a new hyperconcept "mighty or powerful or strong". Since one of its hyperonyms is in the category "the body" another in "emotions and values" and the first (which we don't have yet) would be in "social and political relations", what category do I put it in? Leaving the category empty doesn't result in any integrity test problems.

Also, some of the translations in Key are incoherent, e.g. thatch [english]; toiture de chaume [french]; palha, colmo, sapé [portuguese];  соломенная крыша [russian]; paja (para techumbre) [spanish]
some of which refer to a roof made of plant material, some to plant material used for making a roof.
I would therefore link it to the hyperconcept of "roof or thatch" (currently "thatch or roof"), or should such be left at the hyperonym?

Also, seems like LibreOffice caused some formatting issues, gonna fix those.

Schweikhard:
English "thatch" has both meanings, though. There are quite some such cases where the English term is ambiguous and the translations thereby became incoherent.

The empty lines and quotation marks should be fixed (with the quotes it was a mistake in the saving of the tsv-files with LibreOffice, so fixed it by simply saving the content with different settings or with Gedit)

concepticondata/concepticon.tsv
```diff
@@ -31,7 +31,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 30	DISPEL	Motion	Force to go away.	Action/Process	
 31	PAN	Food and drink	Cooking utensil relatively flat, having a handle, used to cook food such as meat, vegetables or eggs, usually in oil or butter.	Person/Thing	
 32	EARWAX	The body	A yellowish, waxlike secretion from certain glands in the external auditory canal.	Person/Thing	
-33	NETTLE	Agriculture and vegetation	The common name for between 30-45 species of flowering plants with a cosmopolitan though mainly temperate distribution. They are mostly herbaceous perennial plants, but some are annual and a few are shrubby.	Person/Thing	
+33	NETTLE	Agriculture and vegetation	Any of numerous plants having stinging hairs that cause skin irritation on contact (especially of the genus Urtica or family Urticaceae) (Source: BabelNet)	Person/Thing
```
LinguList: We often do not really use that "source". But we link to the metadata instead. Have a look at concept_set_meta, where you find a folder with wikipedia and babelnet, where you could add the id for this. The former translatoin comes probably from wikipedia.

concepticondata/concepticon.tsv
```diff
@@ -3181,4 +3181,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3186	BE DEAD OR DEATH OR DIE	The body	To be not alive or the end of a life or to cease to live.	Other	
 3187	DROWNED	The body	To have perished by suffocation in water.	Property	
 3188	MIGHTY OR POWERFUL OR STRONG		Holding much (political, social, ...) power or being done using a huge amount of energy or having great physical strength.	Property	
-
+3189	STAB TO DEATH	The body	To kill by stabbing with a knife.	Action/Process	
+3190	BROKEN OR SHATTERED	Basic actions and technology	Not capable of working or broken into pieces.	Property
```
LinguList: I don't really understand this definition. I also propose to replace the glosses for broken and shattered by "broken (into pieces)" and "broken". Shattered is not the nromal word one uses often, so the gloss is confusing.

concepticondata/concepticon.tsv
Schweikhard: I see - many definitions are sourced like I did, that's why I assumed it to be the standard method. I assume I only need to add a source if my definition comes from a source not in that folder?

concepticondata/concepticon.tsv
Schweikhard: I wasn't sure about calling it SHATTERED. But some entries on "BROKEN" refer to "broken into pieces", some to "not working", one clearly to both, the rest are ambiguous and the definition is "not working", so I decided to rather move the "broken into pieces" concepts to a new concept. We could instead also change the current definition of "BROKEN" to "broken into pieces" and instead add a new concept "NOT WORKING" for that meaning.

concepticondata/concepticon.tsv
LinguList: They have shortcuts, like [WK:float](en.wikipedia.de/wiki/float), but your source doesn't have the ID. If you add [B:ID] this is also fine, maybe even better, and later, we can link them.

LinguList:
I'm also basically fine with adding "thatch/roof", but it is not clear from the definitions that you gave. If it's for the translations: note that we have to stop at some point.

LinguList:
I'll close this now. We had a long discussion. Please resubmit a new PR after I closed this, so we can see more clearly which changes we had. Afterwards, please @tresoldi, if you find time, could you also throw an eye on this? If not, let us know and we wait until next week after our talks...

PR #495, opened 2018-08-09
Updates

Schweikhard:
New pull request for making the changes easier visible.

In reference to our discussion, one further point about the "thatch/roof": The hyperconcept already exists, so it would be only about moving the link there and not about adding a new concept. What I suggested was moving the ambiguous/incoherent cases on Key's multilingual elicitation list to hyperconcepts. The fact that also the German data reflects these ambiguities is only an additional argument, not the main point. Since you're fine with it in the case of thatch/roof, I'll also do so with the other such cases, but I'll first focus on other updates.


LinguList:
> In reference to our discussion, one further point about the "thatch/roof": The hyperconcept already exists, so it would be only about moving the link there and not about adding a new concept. What I suggested was moving the ambiguous/incoherent cases on Key's multilingual elicitation list to hyperconcepts. The fact that also the German data reflects these ambiguities is only an additional argument, not the main point. Since you're fine with it in the case of thatch/roof, I'll also do so with the other such cases, but I'll first focus on other updates.

I'm fine with that. 

If you go on working in more things, maybe check as well the "live" and "die" mappings. In Chinese questionnaires it definitely is "live or living" or similar, as it is an adjectival reading that is used equivalently to our nouns. If you wan to work further in piece before, you may also close this PR, as we prefer complete ones, where all discussions are solved, and then, once you think you reached a point of completion, PR again.

LinguList:
Ah, and specific questions may then be handled via issues, which you may mark as "question" in the title or so.

Schweikhard:
Okay! "live" has no links, and both "die" and "be alive" I checked before and they look fine to me, except for cases I fixed already.

LinguList:
Nice. Please make, however, sure, to submit a new PR, once you feel you reached a consistent stage. If it's too long, it'll be very dificult to review, but if it's not complete, we may forget certain things, etc.

PR #496, opened 2018-08-13
Update

Schweikhard:
I've implemented all updates that I could change with the information I have. There are some left but I will first have to ask questions or discuss them, so I'll add github issues for them.


Schweikhard:
While going through my remaining questions I found some further things to change which I had previously been unsure about but which I now with more experience decided to continue with on my own. I'll update this pull request (or make a new one) when I'm done.
I'll also add the remaining questions as issues.

Schweikhard:
Finished for now.

concepticondata/conceptrelations.tsv
```diff
@@ -530,3 +529,80 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
...
+3226	FASTEN OR SQUEEZE	narrower	1094	FASTEN
+1587	SMELL (STINK)	narrower	3229	STINK
+3228	BALL OR SPHERE	narrower	3227	SPHERE
+3228	BALL OR SPHERE	narrower	1845	BALL
+1088	SINK (DESCEND)	narrower	2908	SINK (DISAPPEAR IN WATER)
+3231	DEITY	narrower	1944	GOD
+588	BROKEN	narrower	3190	BROKEN (NOT WORKING)
+588	BROKEN	narrower	3191	BROKEN (SHATTERED)
+1352	KNIFE	narrower	3233	KITCHEN KNIFE
+580	FUR	narrower	2678	HAIR OR FUR
```
LinguList: Important: this has been swapped:

HAIR OR FUR "the narrower concept is" FUR

right?

concepticondata/conceptlists/List-2014-1280.tsv
```diff
@@ -517,7 +517,7 @@ List-2014-1280-515	1623	SPEAK	515	speak, talk	tell story	20	http://clics.lingpy.
 List-2014-1280-516	1866	STUTTER	516	stutter, stammer	stutter, stammer	238	http://clics.lingpy.org/browse.php?community=cluster_238_stutter, stammer	0	http://clics.lingpy.org/browse.php?community=network_stutter, stammer_0&view=part	227	11	11	1161
 List-2014-1280-517	1458	SAY	517	say	tell story	20	http://clics.lingpy.org/browse.php?community=cluster_20_tell story	27	http://clics.lingpy.org/browse.php?community=network_say_27&view=part	308	73	129	37
 List-2014-1280-518	1711	TELL	518	tell story	tell story	20	http://clics.lingpy.org/browse.php?community=cluster_20_tell story	28	http://clics.lingpy.org/browse.php?community=network_tell story_28&view=part	251	44	97	124
-List-2014-1280-519	1115	SPEECH	519	speech (make a)	tell story	20	http://clics.lingpy.org/browse.php?community=cluster_20_tell story	19	http://clics.lingpy.org/browse.php?community=network_speech (make a)_19&view=part	218	43	93	142
+List-2014-1280-519	3209	MAKE A SPEECH	519	speech (make a)	tell story	20	http://clics.lingpy.org/browse.php?community=cluster_20_tell story	19	http://clics.lingpy.org/browse.php?community=network_speech (make a)_19&view=part	218	43	93	142
```
LinguList: did you check the reflexes? I mean: what does KEY have? You know, the "(make a)" can well be a gloss exemplifying what the noun is, it is crucial to check with KEY and WOLD, to make sure what was intended if you want to change it.

concepticondata/concepticon.tsv
```diff
@@ -5,8 +5,8 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 4	COURTYARD	The house	An area wholly or partly surrounded by walls or buildings.	Person/Thing	
 5	GAZELLE	Animals	An antelope of the genus Gazella mostly native to Africa and capable of running at high speeds for long periods.	Person/Thing	
 6	EARTHQUAKE	The physical world	The violent shaking of the ground produced by deep seismic waves, originating from the epicentre.	Person/Thing	
-7	GATHER	Spatial relations	To heap up; to collect or gather (e.g. work, magazines, etc.).	Action/Process	
-8	CURSE	Religion and belief	An appeal or prayer for evil or misfortune to befall someone or something.	Action/Process	
+7	GATHER	Spatial relations	To collect or gather (e.g. work, magazines, etc.).	Action/Process	
+8	CURSE	Religion and belief	To wish for evil or misfortune to befall someone or something.	Action/Process
```
tresoldi: The previous definition for CURSE was indeed more related to the noun than to the verb, but "to curse" (which is the actual elicitation) is closer to "to cast" than "to wish" -- even in IE languages, you can trace back the second etymology of most comparanda to *preḱ- or *wenh1.

I would change it to something like "to imprecate evil or misfortune upon someone or something, usually by means of a prayer or implication".

concepticondata/concepticon.tsv
```diff
@@ -416,19 +416,19 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
 419	SEVERE	Miscellaneous function words	Intensely or extremely bad or unpleasant in degree or quality.	Property	
 420	FOREST	The physical world	A vegetation community dominated by trees and other woody shrubs, growing close enough together that the tree tops touch or overlap, creating various degrees of shade on the forest floor.	Person/Thing	
 421	RETREAT	Warfare and hunting	To retreat unwillingly and/or suddenly, often caused by sudden pain, shock, scare, fright, surprise, or similar.	Action/Process	
 422	NECESSITY	Cognition	Something that needs to be done.	Person/Thing	
 423	GET LOST	Basic actions and technology	To end up in an unknown place, to be not to be found again, to get lost, to irrecoverably slip away.	Action/Process	
-424	MARRIED MAN	Kinship	A married man	Person/Thing	
-425	BE LATE	Time	to be late	Action/Process	
-426	MARRIED WOMAN	Kinship	a married woman	Person/Thing	
+424	MARRIED MAN	Kinship	A married man.	Person/Thing	
+425	BE LATE	Time	To be later than the expected time.	Property
```
tresoldi: Are we sure about this being a property and not a verb? In Portuguese we have almost that, it is always in an inchoative aspect but it is a verb.

concepticondata/concepticon.tsv
```diff
@@ -489,7 +489,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 488	MANNER	Cognition	Way of performing or effecting anything.	Person/Thing	
 489	KIDNEY	The body	An organ in the body that produces urine.	Person/Thing	
 490	DESCENDANTS	Kinship	One who is the progeny of someone at any distance of time.	Person/Thing	
-491	CENTER	Spatial relations	The middle part, that can be clearly discerned from the enveloping part.	Person/Thing	
+491	CENTER	Spatial relations	What's in the mid of a plane or sphere.	Person/Thing
```
tresoldi: If it is person/thing referring to spatial relations, the original definition seems more appropriate. It should refer to the area/volume of something, not necessarily a plane or sphere (think about typical elicitations such as "what is in the center of the village?")

concepticondata/concepticon.tsv
```diff
@@ -524,7 +524,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 523	BACK OF HEAD	The body	The back part of the head.	Person/Thing	
 524	BARBEL	Animals	European freshwater fish belonging to the family of carps (Cyprinidae), Genus of barbus.	Person/Thing	
 525	BLOOD VESSEL	The body	The blood vessels are the part of the circulatory system that transports blood throughout the human body.	Person/Thing	
-526	HOWL	Speech and language	The protracted, mournful cry of a dog or a wolf.	Action/Process	
+526	HOWL	Speech and language	To cry like a wolf.	Action/Process
```
tresoldi: like a wolf, dog, or other canids.

concepticondata/concepticon.tsv
```diff
@@ -572,21 +572,21 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 571	STING	The body	To hurt, usually by introducing poison or a sharp point, or both.	Action/Process	
 572	STREW	Motion	To distribute objects or pieces of something over an area, especially in a random manner.	Action/Process	
 573	SWAY	Motion	To move or swing from side to side; or backward and forward; to rock.	Action/Process	
-574	TRANSIT	Motion	To pass over, across or through something.	Action/Process	
+574	TRANSIT	Motion	To cross something (a river, a street etc.).	Action/Process
```
tresoldi: This is based on the actual words mapped to this concept? TRANSIT as a gloss seems more general and less specific, almost MOVE (or perhaps the idea is to differentiate from MOVE?)

concepticondata/concepticon.tsv
```diff
@@ -670,22 +670,22 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+679	BRIGHT	Sense perception	Emitting or reflecting light readily or in large amounts. [WN:679]	Property	
+680	BREAK (DESTROY OR GET DESTROYED)	Basic actions and technology	To divide or be divided into pieces and thereby damaged.	Action/Process	
 681	DESTROY	Possession	To damage beyond use or repair.	Action/Process	
 682	SIDE	Spatial relations	One side of a leaf of a book or manuscript.	Person/Thing	
 683	PERSON	Kinship	A human being.	Person/Thing	
 684	OWN	Possession	To have rightful possession of property, goods or capital.	Action/Process	
 685	DECIDE	Cognition	To reach, make, or come to a decision about something.	Action/Process	
 686	EASY	Cognition	Requiring little skill or effort; posing no difficulty.	Property	
-687	NOTHING	Miscellaneous function words	Not any thing.	Other	
+687	NOTHING	Miscellaneous function words	Not anything.	Other
```
tresoldi: Minor quirk, but I'd keep the space.

concepticondata/concepticon.tsv
```diff
@@ -836,7 +836,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 835	CHAPTER	Speech and language	A part of a book, a literary work etc. with a distinct content, set apart using typography.	Person/Thing	
 836	SCREWDRIVER	Modern world	A hand tool used for driving screws.	Person/Thing	
 837	BLUE	Sense perception	Having the pure color of a clear sky; the primary color between green and violet in the visible spectrum, an effect of light with a wavelength between 450 and 500 nm.	Property	
-838	REMAINS	Spatial relations	The physical structure of a dead animal or person.	Person/Thing	
+838	REMAINS	Spatial relations	What remains after other things have been taken away (e.g. eaten or destroyed).	Person/Thing
```
tresoldi: Should it be 'spatial relations'?

concepticondata/concepticon.tsv
```diff
@@ -670,22 +670,22 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 669	RODENT	Animals	Any of the relatively small placental mammals that constitute the order Rodentia, having constantly growing incisor teeth specialized for gnawing.	Person/Thing	
 670	ROOT	Agriculture and vegetation	The absorbing and anchoring organ of a vascular plant; it bears neither leaves nor flowers and is usually subterranean.	Person/Thing	
 671	SAND	The physical world	A loose material consisting of small mineral particles, or rock and mineral particles, distinguishable by the naked eye; grains vary from almost spherical to angular, with a diameter range from 1/16 to 2 millimeters.	Person/Thing	
-672	EARLY	Time	Before the expected time.	Other	
+672	EARLY	Time	Towards the end of a time period or before the expected time.	Property
```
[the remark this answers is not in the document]
LinguList: Yes, this is in fact crucial: when distinguishing things, to make a decision between, say, "child (descendant)" and "child (young human)", we NEED a list that makes the distinction. Although, since it's but a definition, it is not that crucial to change it as long as no re-mapping is involved.

concepticondata/concepticon.tsv
```diff
@@ -848,13 +848,13 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
-850	SQUARE	Spatial relations	An open area in a town, sometimes including the surrounding buildings.	Person/Thing	
+850	SQUARE	Spatial relations	A geometrical figure with four sides that meet each other on right angles, or something resembling its shape.	Person/Thing	
 851	STABLE	Animals	A building or structure usually with stalls that is used to house and feed horses, cattle or other animals.	Person/Thing	
 852	PRIEST	Religion and belief	A minister of the Catholic church empowered to administer the sacraments, most particularly that of the Eucharist or Holy Communion, as well as those of confession and extreme unction.	Person/Thing	
 853	ALTAR	Religion and belief	Table for religious ceremonies.	Person/Thing	
 854	KETTLE	Food and drink	A metal container used to boil any kind of liquid.	Person/Thing	
 855	SEIZE	Possession	To take possession of by force or authority.	Action/Process	
-856	HIRE	Possession	To hold under a lease or rental agreement of goods and services.	Action/Process	
+856	HIRE	Possession	To employ a person.	Action/Process
```
tresoldi: To employ a person or people in exchange of some kind of remuneration, usually a payment.

concepticondata/concepticon.tsv
```diff
@@ -996,12 +996,12 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 995	QUIVER	Warfare and hunting	A container for arrows, crossbow bolts or darts, such as those fired from a bow, crossbow or blowgun.	Person/Thing	52
 996	DEAF	The body	Not having the faculty of hearing, or only having a restricted capability to hear.	Property	
 997	BAOBAB TREE	Agriculture and vegetation	Genus of big trees (Adansonia) from the mallow family, growing in tropical countries	Person/Thing	
-998	PUT	Spatial relations	To set or place an object in a different place than it original was.	Action/Process	
+998	PUT	Spatial relations	To set or place an object somehwere.	Action/Process
```
tresoldi: somehwere -> somewhere

concepticondata/concepticon.tsv
```diff
@@ -996,12 +996,12 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
 997	BAOBAB TREE	Agriculture and vegetation	Genus of big trees (Adansonia) from the mallow family, growing in tropical countries	Person/Thing	
-998	PUT	Spatial relations	To set or place an object in a different place than it original was.	Action/Process	
+998	PUT	Spatial relations	To set or place an object somehwere.	Action/Process	
 999	JOINT	The body	Any part of the body where two bones join.	Person/Thing	
 1000	ANXIETY	Emotions and values	An unpleasant complex combination of emotions that includes fear, apprehension and worry, and is often accompanied by physical sensations such as palpitations, nausea, chest pain and/or shortness of breath.	Person/Thing	
 1001	BLACKSMITH	Basic actions and technology	A craftsman expert in iron and other metals work.	Person/Thing	
 1002	STATUE	Basic actions and technology	A three-dimension work of art, usually of a person or animal, usually created by sculpting, carving, moulding, or casting.	Person/Thing	
-1003	PERMIT	Social and political relations	To approve a specific action.	Action/Process	
+1003	PERMIT	Social and political relations	To give permission to an action or to not try to stop someone from doing it.	Action/Process
```
tresoldi: I'd only keep the first part (give permission), the second (not try to stop) looks like a different concept to me (even though it is clear that it is not a meaning such as "facilitate")

concepticondata/concepticon.tsv
```diff
@@ -1045,7 +1045,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1044	PILLOW	The house	Cushion placed under the head for sleeping.	Person/Thing	
 1045	POET	Speech and language	A person who writes poems.	Person/Thing	
 1046	SHUT	Spatial relations	To move (a door, a window, etc.) so that it closes its opening.	Action/Process	
-1047	PURSUE	Motion	Trying to achieve or to reach something.	Action/Process	
+1047	PURSUE	Motion	Trying to achieve, to catch or to reach something or someone.	Action/Process
```
tresoldi: "to catch or to reach" -> "to catch, or to reach" (comma)

concepticondata/concepticon.tsv
```diff
@@ -1082,18 +1082,18 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1081	BILL	Possession	Sum owed in a restaurant, or the relative bill.	Person/Thing	
 1082	MERCHANT	Possession	A businessperson engaged in trade.	Person/Thing	
 1083	MESS	The physical world	A state of confusion and disorderliness.	Person/Thing	
-1084	CALL	Speech and language	To order or summon by using one's voice.	Action/Process	
+1084	CALL	Speech and language	To ask or order someone to come.	Action/Process
```
tresoldi: This is a difficult concept to define, but from what I remember from the datasets I've worked on the original meaning seems closer to what conceptlists tend to imply. In particular, I'd at least at the hierarchical relationship expressed by original, at least as an alternative, with something like "To ask or summon someone"; at the same time, a "CALL" is not necessarily used to motivate a movement, but can be more generally to ask for attention and thus establish a communication.

concepticondata/concepticon.tsv
```diff
@@ -1082,18 +1082,18 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
 1087	CONCEIVE	The body	To become pregnant.	Action/Process	
-1088	SINK (DESCEND)	Motion	To cause a boat to go down in the water.	Action/Process	
+1088	SINK (DESCEND)	Motion	To go down in water, mud, quicksand etc. or to cause something to go down in it.	Action/Process	
 1089	FRAXINUS	Agriculture and vegetation	Any of the trees belonging to the genus Fraxinus.	Person/Thing	
 1090	BE BORN	The body	To start living.	Action/Process	
 1091	SALTY	Sense perception	Containing salt.	Property	
-1092	HOPE	Emotions and values	To want something to happen.	Action/Process	
-1093	MIDDLE	Spatial relations	Being in the midway between the ends.	Person/Thing	
+1092	HOPE	Emotions and values	To want something to happen or the belief in the possibility that something one wants will happen.	Action/Process	
+1093	MIDDLE	Spatial relations	What is equally far away from two ends or being in that position.	Other
```
tresoldi: There is probably room for a MIDDLE OR CENTER general concept.

concepticondata/concepticon.tsv
```diff
@@ -1190,7 +1190,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1189	ULCER	The body	An open sore of the skin, eyes or mucous membrane, often caused by an initial abrasion and generally maintained by an inflammation and/or an infection.	Person/Thing	
 1190	HARE	Animals	Mammal of the family hares and rabbits (Leporidae) with long ears, short tail and hindlegs which are shorter than the forelegs and permit running quickly.	Person/Thing	
 1191	COOKED RICE	Food and drink	Boiled rice.	Person/Thing	
-1192	THREATEN	Speech and language	To put in danger.	Action/Process	
+1192	THREATEN	Speech and language	To utter intentions of injury or punishment against. [WN:1192]	Action/Process
```
tresoldi: Not necessarily `intentions`, but simply (as implied by the gloss), `threats`. I'd say "To make a threat of injury or punishment against someone", but please check if the meaning in the wordlist that use it is not something like "menace" (e.g., "the beast threatens us").

concepticondata/concepticon.tsv
```diff
@@ -1265,7 +1265,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1264	CHANGE	Spatial relations	To make different.	Action/Process	
 1265	HIGH	Spatial relations	Having relatively great height.	Property	
 1266	BRIDE	Kinship	A woman who is going to get married or has just got married.	Person/Thing	
-1267	NARROW	Spatial relations	Having a small width.	Property	
+1267	NARROW	Spatial relations	(Of a path, cave etc.) having a small width.	Property
```
tresoldi: I remember some examples using objects, maybe add something like "of a cane, straw, etc."

concepticondata/concepticon.tsv
```diff
@@ -1365,7 +1365,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1364	HEALTHY	The body	In good physical and mental condition; free from disease.	Property	
 1365	KITCHEN	The house	A room equipped for preparing and cooking food.	Person/Thing	
 1366	BOY	Kinship	A young male person, usually a child or teenager.	Person/Thing	
-1367	CROSS	Spatial relations	A geometrical figure consisting of two straight lines or bars intersecting each other such that at least one of them is bisected by the other.	Person/Thing	
+1367	CROSS	Spatial relations	A geometrical figure consisting of two straight lines or bars bisecting each other, or anything of similar shape.	Person/Thing
```
tresoldi: "...or any figure or object of similar shape"?

concepticondata/concepticon.tsv
```diff
@@ -1673,7 +1673,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1672	WRITE	Speech and language	To form (letters, words or symbols) on a surface in order to communicate.	Action/Process	
 1673	ARM	The body	The upper limb, extending from the shoulder to the wrist and sometimes including the hand.	Person/Thing	
 1674	POOR	Possession	With little or no possessions or money.	Property	
-1675	PROMISE	Speech and language	A transaction between two persons whereby the first person undertakes in the future to render some service or gift to the second person or devotes something valuable now and here to his use.	Action/Process	
+1675	PROMISE	Speech and language	To assure someone that one will do something or that something will happen.	Action/Process
```
tresoldi: Not only `to assure` someone; for most societies, the elicitation probably picks something like "to commit".

concepticondata/concepticon.tsv
```diff
@@ -1694,7 +1694,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1693	THURSDAY	Time	The fourth day of the week in Europe and in systems using the ISO 8601 norm; the fifth day of the week in the United States of America.	Person/Thing	
 1694	WEDNESDAY	Time	The third day of the week in Europe and in systems using the ISO 8601 norm; the fourth day of the week in the United States of America.	Person/Thing	
 1695	TUESDAY	Time	The second day of the week in Europe and in systems using the ISO 8601 norm; the third day of the week in the United States of America.	Person/Thing	
-1696	SURRENDER	Warfare and hunting	The act of surrendering to the enemy.	Action/Process	
+1696	SURRENDER	Warfare and hunting	To give up to the enemy.	Action/Process
```
tresoldi: At least in English we have the related meaning of "to yield", e.g., "to surrend the village/the arms". Should we include it? Is it found in the concept lists and in the data?

concepticondata/concepticon.tsv
```diff
@@ -1723,7 +1723,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1722	EIGHTY	Quantity	The cardinal number occurring after seventy-nine and before eighty-one, represented in Roman numerals as LXXX and in Arabic numerals as 80.	Number	
 1723	RESULT IN	Cognition	Having a specific result, a logical consequence.	Action/Process	
 1724	NINETY	Quantity	The cardinal number occurring after eighty-nine and before ninety-one, represented in Roman numerals as XC and in Arabic numerals as 90.	Number	
-1725	CORRECT (RIGHT)	Emotions and values	That are in accordance with fact.	Property	
+1725	CORRECT (RIGHT)	Emotions and values	To be in accordance with fact.	Property
```
tresoldi: Not only with fact, but also with what is expected.

concepticondata/concepticon.tsv
```diff
@@ -1733,14 +1733,14 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
-1735	TEAR (SHRED)	Basic actions and technology	A tug given with force in order to remove something.	Action/Process	
+1735	TEAR (SHRED)	Basic actions and technology	To pull at something in such a way that it ends up divided.	Action/Process	
 1736	TWENTY FOUR	Quantity	The cardinal number occurring after twenty-three and before twenty-five, represented in Roman numerals as XXIV and in Arabic numerals as 24.	Number	
 1737	OKRA	Agriculture and vegetation	-	Person/Thing	
 1738	TWENTY FIVE	Quantity	The cardinal number occurring after twenty-four and before twenty-six, represented in Roman numerals as XXV and in Arabic numerals as 25.	Number	
 1739	DECORATE	Clothing and grooming	To make more attractive by adding ornament, colour, etc.	Action/Process	
 1740	PINE	Agriculture and vegetation	Evergreen coniferous tree of the genus Pinus.	Person/Thing	
 1741	ABOVE	Spatial relations	In or to a higher place; higher than; on or over the upper surface.	Other	
-1742	TRAVERSE	Basic actions and technology	To go beyond, to pass here.	Action/Process	
+1742	TRAVERSE	Motion	To go or pass through something.	Action/Process
```
tresoldi: Related to the TRANSIT above?

concepticondata/concepticon.tsv
```diff
@@ -1782,12 +1782,12 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
 1783	PAIN	Emotions and values	An unpleasant, usually localised physical sensation that is often the result of an injury, disease or other ailment.	Person/Thing	
-1784	WANT	Emotions and values	To have a strong desire for something.	Action/Process	
+1784	WANT	Emotions and values	To want to do something or to wish for something to happen.	Action/Process	
 1785	HAVE SEX	The body	To take part in sexual activity (most commonly sexual intercourse) with another person for the purposes of sexual pleasure.	Action/Process	549
 1786	DRY UP	Basic actions and technology	To lose moisture, usually through evaporation or absorption.	Action/Process	
 1787	LOOM	Clothing and grooming	A frame or machine (made of wood or other material) in which a weaver forms cloth out of thread; a machine for interweaving yarn or threads into a fabric.	Person/Thing	
 1788	TELEVISION	Modern world	A device for receiving television signals and displaying them in visual form.	Person/Thing	
-1789	AGE	Time	The period of time that a person, animal or plant has lived or is expected to live.	Person/Thing	
+1789	AGE	Time	The amount of time that a person, animal or plant has lived.	Person/Thing
```
tresoldi: Or that an object came into being (but not sure if this should be added to the definition).

concepticondata/concepticon.tsv
```diff
@@ -1930,7 +1930,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1929	STUDY	Cognition	To follow a course of study; to be enrolled at an institute of learning.	Action/Process	
 1930	RAVINE	The physical world	A deep narrow valley or gorge in the earth's surface worn by running water.	Person/Thing	
 1931	BEESWAX	Animals	a yellowish or dark brown wax secreted by honeybees for constructing honeycombs.	Person/Thing	
-1932	GUESS	Cognition	To succeed to know, or discover, by way of surmises.	Action/Process	
+1932	GUESS	Cognition	To make a surmise.	Action/Process
```
tresoldi: Maybe just "to suppose"? Make a surmise seems a bit too technical.

concepticondata/concepticon.tsv
```diff
@@ -1942,7 +1942,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1941	SOLDIER	Warfare and hunting	A person who is actively engaged in battle, conflict or warfare.	Person/Thing	
 1942	NEAR	Spatial relations	Not far distant in time or space or degree or circumstances.	Property	
 1943	CROWN (OF HEAD)	The body	A royal, imperial or princely headdress; a diadem.	Person/Thing	
-1944	GOD	Religion and belief	A supernatural, typically immortal being with superior powers.	Person/Thing	
+1944	GOD	Religion and belief	The deity of a monotheistic religion.	Person/Thing
```
tresoldi: Considering the words we have (even from IE in NorthEuralex and everything), the original is the correct. While most data might be collected referring to Abrahamic religions, this is just an obvious colexification. I'd strongly suggest to keep the original here.

concepticondata/concepticon.tsv
```diff
@@ -2115,7 +2115,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2114	BELLY OR STOMACH	The body	The lower part of the front of the torso, including the interior part below.	Person/Thing	
 2115	CEASE OR FINISH	Basic actions and technology	To stop doing something, voluntarily or involuntarily.	Action/Process	
 2116	REST OR SLEEP	Basic actions and technology	Lie down or sleep.	Action/Process	
-2117	DESIRE	Emotions and values	To have affection for someone or to want to do something.	Action/Process	
+2117	DESIRE	Emotions and values	To have affection for someone or to want to do something or to wish for something to happen.	Action/Process
```
tresoldi: All correct, but I'd keep the "have affection" at the end, as the most specific one. Thus, to want, to wish, and to have affection.

concepticondata/concepticon.tsv
```diff
@@ -2247,7 +2247,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2246	BELL	Basic actions and technology	A simple sound-making device.	Person/Thing	
 2247	CATERPILLAR	Animals	The larvae of members of the order Lepidoptera (the insect order comprising butterflies and moths).	Person/Thing	
 2248	KNOW (SOMEBODY)	Cognition	To have met and recognize somebody.	Action/Process	
-2249	THIN (OF SHAPE OF OBJECT)	Spatial relations	Being narrow in size.	Property	
+2249	THIN (OF SHAPE OF OBJECT)	Spatial relations	Being very narrow (of flat or long objects).	Property
```
tresoldi: Why 'very' narrow? This is usually elicited in plain oposition with 'thick', right? It is not necessarily 'very' narrow

concepticondata/concepticon.tsv
```diff
@@ -2303,8 +2303,8 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2302	CIRCULAR (ROUND IN TWO DIMENSIONS)	Spatial relations	Having a circular shape.	Property	
 2303	SPHERICAL (ROUND IN THREE DIMENSIONS)	Spatial relations	Having a spherical shape.	Property	
 2304	THEE (OBLIQUE CASE OF YOU)	Kinship	Oblique cases of the second person singular pronoun.	Other	
-2305	THIN (OF HAIR AND THREAD)	Spatial relations	Being very narrow (of long objects, like &quot;hairs&quot; or &quot;wires&quot;).	Property	
-2306	THIN (OF LEAF AND CLOTH)	Spatial relations	Being very narrow (of flat objects, like &quot;leaves&quot; or &quot;cloth).	Property	
+2305	THIN (OF HAIR AND THREAD)	Spatial relations	Being very narrow (of long objects, like hairs or wires).	Property	
+2306	THIN (OF LEAF AND CLOTH)	Spatial relations	Being very narrow (of flat objects, like leaves or cloth).	Property
```
tresoldi: Here the 'very' is far more acceptable (see comment for THIN (OF SHAPE OF OBJECT) above)

concepticondata/concepticon.tsv
```diff
@@ -2676,7 +2676,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2675	FIRE OR FIREWOOD	The physical world	The state of combustion in which inflammable material burns, producing heat, flames and often smoke or the Wood intended to be burned.	Person/Thing	
 2676	TWO HUNDRED	Quantity	The cardinal number occurring after one hundred and ninety-nine and before two hundred one, represented in Roman numerals as CC and in Arabic numerals as 200.	Number	
 2677	SPEARTHROWER	Warfare and hunting	A spear-thrower or atlatl is a tool that uses leverage to achieve greater velocity in dart-throwing, and includes a bearing surface which allows the user to store energy during the throw.	Person/Thing	
-2678	HAIR OR FUR	The body	The cover of a mammal or a bird.	Person/Thing	
+2678	HAIR OR FUR	The body	The hair covering the body of a mammal, including the body hair of humans.	Person/Thing
```
tresoldi: This relates to the comments to FUR above.

concepticondata/concepticon.tsv
```diff
@@ -2902,7 +2902,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2905	KNIT	Basic actions and technology	Knitting is a method by which yarn is manipulated to create a textile or fabric for use in many types of garments.	Action/Process	
 2906	TRANSLATE	Basic actions and technology	To communicate of the meaning of a source-language text by means of an equivalent target-language text.	Action/Process	
 2907	ROT	Basic actions and technology	To be exposed to decomposition.	Action/Process	
-2908	SINK (DISAPPEAR IN WATER)	Basic actions and technology	To descend completely, leaving no trace in the surface.	Action/Process	
+2908	SINK (DISAPPEAR IN WATER)	Basic actions and technology	To go down in water, mud, quicksand etc.	Action/Process
```
tresoldi: So, the difference to the other SINK is that this one does not allow for agency?

concepticondata/concepticon.tsv
```diff
@@ -2994,7 +2994,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2998	DAMP	Sense perception	Refers to the presence of a liquid, especially water, often in trace amounts.	Property	
 2999	EVERYWHERE	Spatial relations	In all places.	Other	
 3000	LIVING	The body	Being alive.	Property	
-3001	POWERFUL	Emotions and values	Having great power.	Property	
+3001	POWERFUL	Emotions and values	Being done using a huge amount of energy.	Property
```
tresoldi: Thinking from elicitations and other concept list, the original seems closer to most intentions. This usually refers to a human (leader, guide) or to a supernatural being which has a lot of power, correct?

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
 3180	SHOOT (WITH SLINGSHOT)	Warfare and hunting	Use a rubber-based projectile weapon to propel projectiles against a target.	Action/Process	
 3181	MUSTARD	Food and drink	It can be a yellow sauce with a strong taste or a plant which is used to make mustard sauce.	Person/Thing	
+3182	MARRIAGE OR WEDDING	Kinship	The ceremony of marrying someone or the state of being married.	Person/Thing	
+3183	BREATH OR BREATHE	The body	The air that goes into one's lungs or the act of breathing	Other	
+3184	BE ALIVE OR LIFE	The body	The act of being alive or the life itself.	Other	
+3185	BE DEAD OR DIE OR KILL	The body	To be not alive or to cease to live or to cause the end of a life.	Other	
+3186	BE DEAD OR DEATH OR DIE	The body	To be not alive or the end of a life or to cease to live.	Other	
+3187	DROWNED	The body	To have perished by suffocation in water.	Property	
+3188	MIGHTY OR POWERFUL OR STRONG	Emotions and values	Holding much (political, social, ...) power or being done using a huge amount of energy or having great physical strength.	Property
```
tresoldi: This relates to the comments of POWERFUL above

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3182	MARRIAGE OR WEDDING	Kinship	The ceremony of marrying someone or the state of being married.	Person/Thing	
+3183	BREATH OR BREATHE	The body	The air that goes into one's lungs or the act of breathing	Other	
+3184	BE ALIVE OR LIFE	The body	The act of being alive or the life itself.	Other	
+3185	BE DEAD OR DIE OR KILL	The body	To be not alive or to cease to live or to cause the end of a life.	Other	
+3186	BE DEAD OR DEATH OR DIE	The body	To be not alive or the end of a life or to cease to live.	Other	
+3187	DROWNED	The body	To have perished by suffocation in water.	Property	
+3188	MIGHTY OR POWERFUL OR STRONG	Emotions and values	Holding much (political, social, ...) power or being done using a huge amount of energy or having great physical strength.	Property	
+3189	STAB TO DEATH	The body	To kill by stabbing with a knife.	Action/Process	
+3190	BROKEN (NOT WORKING)	Basic actions and technology	Not capable of working.	Property	
+3191	BROKEN (SHATTERED)	Basic actions and technology	Broken into pieces.	Property
```
tresoldi: But this refers only to objects, correct? If I remember correctly, in the Tanzanian languages the examples for BROKEN seem to come from "broken bone"; I like the definitions we have now (certainly better than the confusion of all the previous "brokens"), but I can't really see what would map to "fractured". This one only referes to "fragmented", in a sense.

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3185	BE DEAD OR DIE OR KILL	The body	To be not alive or to cease to live or to cause the end of a life.	Other	
+3186	BE DEAD OR DEATH OR DIE	The body	To be not alive or the end of a life or to cease to live.	Other	
+3187	DROWNED	The body	To have perished by suffocation in water.	Property	
+3188	MIGHTY OR POWERFUL OR STRONG	Emotions and values	Holding much (political, social, ...) power or being done using a huge amount of energy or having great physical strength.	Property	
+3189	STAB TO DEATH	The body	To kill by stabbing with a knife.	Action/Process	
+3190	BROKEN (NOT WORKING)	Basic actions and technology	Not capable of working.	Property	
+3191	BROKEN (SHATTERED)	Basic actions and technology	Broken into pieces.	Property	
+3192	BRONZE OR COPPER	The physical world	Chemical element with symbol Cu and atomic number 29 or an alloy made from it.	Person/Thing	
+3193	PAINT (WALL)	Basic actions and technology	To apply color to a wall.	Action/Process	
+3194	PAINT (WALL) OR PAINTING	Basic actions and technology	To apply color to a wall or to make a painting or drawing.	Action/Process
```
tresoldi: Do we have wordlists which actually make this distinction, between 3193 and 3194?

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3186	BE DEAD OR DEATH OR DIE	The body	To be not alive or the end of a life or to cease to live.	Other	
+3187	DROWNED	The body	To have perished by suffocation in water.	Property	
+3188	MIGHTY OR POWERFUL OR STRONG	Emotions and values	Holding much (political, social, ...) power or being done using a huge amount of energy or having great physical strength.	Property	
+3189	STAB TO DEATH	The body	To kill by stabbing with a knife.	Action/Process	
+3190	BROKEN (NOT WORKING)	Basic actions and technology	Not capable of working.	Property	
+3191	BROKEN (SHATTERED)	Basic actions and technology	Broken into pieces.	Property	
+3192	BRONZE OR COPPER	The physical world	Chemical element with symbol Cu and atomic number 29 or an alloy made from it.	Person/Thing	
+3193	PAINT (WALL)	Basic actions and technology	To apply color to a wall.	Action/Process	
+3194	PAINT (WALL) OR PAINTING	Basic actions and technology	To apply color to a wall or to make a painting or drawing.	Action/Process	
+3195	DAMAGE OR INJURE	Possession	To put a thing in bad condition by making it suffer some damage or to cause physical harm to a living creature.	Action/Process
```
tresoldi: 'making suffer' does not sound very good, maybe just "to cause damage" (which does not necessarily imply intention)?

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3187	DROWNED	The body	To have perished by suffocation in water.	Property	
+3188	MIGHTY OR POWERFUL OR STRONG	Emotions and values	Holding much (political, social, ...) power or being done using a huge amount of energy or having great physical strength.	Property	
+3189	STAB TO DEATH	The body	To kill by stabbing with a knife.	Action/Process	
+3190	BROKEN (NOT WORKING)	Basic actions and technology	Not capable of working.	Property	
+3191	BROKEN (SHATTERED)	Basic actions and technology	Broken into pieces.	Property	
+3192	BRONZE OR COPPER	The physical world	Chemical element with symbol Cu and atomic number 29 or an alloy made from it.	Person/Thing	
+3193	PAINT (WALL)	Basic actions and technology	To apply color to a wall.	Action/Process	
+3194	PAINT (WALL) OR PAINTING	Basic actions and technology	To apply color to a wall or to make a painting or drawing.	Action/Process	
+3195	DAMAGE OR INJURE	Possession	To put a thing in bad condition by making it suffer some damage or to cause physical harm to a living creature.	Action/Process	
+3196	BE LEFT	Spatial relations	To continue to be in the same place after others have left or been removed.	Property
```
tresoldi: This gloss is a bit difficult in English, maybe something with 'remain' or even 'perdure'? And, also, is it a spatial relation?

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3195	DAMAGE OR INJURE	Possession	To put a thing in bad condition by making it suffer some damage or to cause physical harm to a living creature.	Action/Process	
+3196	BE LEFT	Spatial relations	To continue to be in the same place after others have left or been removed.	Property	
+3197	SUSPICION	Cognition	An assumption without direct evidence.	Person/Thing	
+3198	NOTHING OR ZERO	Miscellaneous function words	Not anything or the cardinal number that denotes no quantity or amount at all.	Other	
+3199	GUESS (SUCCESSFULLY)	Cognition	To succeed to know, or discover, by way of surmises.	Action/Process	
+3200	COUNTRYSIDE	The physical world	The region outside of a city or town.	Person/Thing	
+3201	HINDER OR PREVENT	Social and political relations	To make it difficult for something to occur or to fully stop it from occurring.	Action/Process	
+3202	CANNON	Warfare and hunting	A bullet firing weapon mounted on a carriage.	Person/Thing	
+3203	CANNON OR GUN	Warfare and hunting	A bullet-firing weapon mounted on a carriage or to be held in the hands or arms.	Person/Thing	
+3204	INSULT	Social and political relations	To use swear-words towards someone.	Action/Process
```
tresoldi: Does this come from some wordlist where indeed the only meaning is using swear words? Isn't it a more general sense of being rude or to demean someone?

concepticondata/concepticon.tsv
LinguList: also, be careful with the gloss: paint a wall is usually not meant as "to paint a real wall", but to distinguish "paint (something like a picture)" from "paint (=color something)". Taking elicitation glosses to literal is tricky: we need to understand why people elicit them. And here, I don't see that a concept is THAT specific.

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3197	SUSPICION	Cognition	An assumption without direct evidence.	Person/Thing	
+3198	NOTHING OR ZERO	Miscellaneous function words	Not anything or the cardinal number that denotes no quantity or amount at all.	Other	
+3199	GUESS (SUCCESSFULLY)	Cognition	To succeed to know, or discover, by way of surmises.	Action/Process	
+3200	COUNTRYSIDE	The physical world	The region outside of a city or town.	Person/Thing	
+3201	HINDER OR PREVENT	Social and political relations	To make it difficult for something to occur or to fully stop it from occurring.	Action/Process	
+3202	CANNON	Warfare and hunting	A bullet firing weapon mounted on a carriage.	Person/Thing	
+3203	CANNON OR GUN	Warfare and hunting	A bullet-firing weapon mounted on a carriage or to be held in the hands or arms.	Person/Thing	
+3204	INSULT	Social and political relations	To use swear-words towards someone.	Action/Process	
+3205	GRAVEN IMAGE	Religion and belief	A representation of a deity who is considered to be false.	Person/Thing	
+3206	ELF OR FAIRY	Religion and belief	A human-like, often youthful-looking being with supernatural powers.	Person/Thing
```
tresoldi: I see we have the "often", but this sounds a bit Tolkienesque -- even medieval North European elves and fairies where not necessarily youthful, human-like, or had supernatural powers. I'd use something like mythical beings similar but distinct from humans, related to nature, etc, and usually with supernatural or magic powers.

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3199	GUESS (SUCCESSFULLY)	Cognition	To succeed to know, or discover, by way of surmises.	Action/Process	
+3200	COUNTRYSIDE	The physical world	The region outside of a city or town.	Person/Thing	
+3201	HINDER OR PREVENT	Social and political relations	To make it difficult for something to occur or to fully stop it from occurring.	Action/Process	
+3202	CANNON	Warfare and hunting	A bullet firing weapon mounted on a carriage.	Person/Thing	
+3203	CANNON OR GUN	Warfare and hunting	A bullet-firing weapon mounted on a carriage or to be held in the hands or arms.	Person/Thing	
+3204	INSULT	Social and political relations	To use swear-words towards someone.	Action/Process	
+3205	GRAVEN IMAGE	Religion and belief	A representation of a deity who is considered to be false.	Person/Thing	
+3206	ELF OR FAIRY	Religion and belief	A human-like, often youthful-looking being with supernatural powers.	Person/Thing	
+3207	GO UP OR RISE (ONESELF)	Motion	To move from a lower to a higher position or to get up from a lying, sitting, or kneeling position.	Action/Process	
+3208	CENTER OR MIDDLE	Spatial relations	What's in the mid of a plane or sphere or equally far away from two ends.	Person/Thing
```
tresoldi: Oh, I missed it the first time! Sorry for the other comment

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3204	INSULT	Social and political relations	To use swear-words towards someone.	Action/Process	
+3205	GRAVEN IMAGE	Religion and belief	A representation of a deity who is considered to be false.	Person/Thing	
+3206	ELF OR FAIRY	Religion and belief	A human-like, often youthful-looking being with supernatural powers.	Person/Thing	
+3207	GO UP OR RISE (ONESELF)	Motion	To move from a lower to a higher position or to get up from a lying, sitting, or kneeling position.	Action/Process	
+3208	CENTER OR MIDDLE	Spatial relations	What's in the mid of a plane or sphere or equally far away from two ends.	Person/Thing	
+3209	MAKE A SPEECH	Speech and language	To give a speech in front of an audience	Action/Process	
+3210	KNIFE (FOR EATING)	Food and drink	A knife used while eating food.	Person/Thing	
+3211	HOPE (FAITH)	Emotions and values	The belief in the possibility that something one wants will happen.	Person/Thing	
+3212	HOPE (SOMETHING)	Emotions and values	To want something to happen.	Action/Process	
+3213	HIRE OR RENT	Social and political relations	To employ a person or to pay for using an object.	Action/Process
```
tresoldi: See comments for HIRE above

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3221	PLANTAGO	Agriculture and vegetation	A plant of the genus Plantago, with a rosette of sessile leaves about 10 cm long with a narrow part instead of a petiole, and with a spike inflorescence with the flower spacing varying widely among the species.	Person/Thing	
+3222	TWIG	Agriculture and vegetation	A branch that doesn't arise from the trunk but from another branch.	Person/Thing	
+3223	BRANCH OR TWIG	Agriculture and vegetation	A woody part of a tree arising from the trunk or from another branch.	Person/Thing	
+3224	WRING	Basic actions and technology	To take a cloth and twist its ends into different directions in order to dry it.	Action/Process	
+3225	SQUEEZE OR WRING	Basic actions and technology	To apply pressure from more than one side or to twist a cloth's ends into different directions.	Action/Process	
+3226	FASTEN OR SQUEEZE	Basic actions and technology	To close tightly or to apply pressure from more than one side.	Action/Process	
+3227	SPHERE	Spatial relations	Any spherical object.	Person/Thing	
+3228	BALL OR SPHERE	Spatial relations	An object, generally spherical, used for playing games, or any spherical object.	Person/Thing
```
tresoldi: How exactly this relates to concept BALL above?

concepticondata/concepticon.tsv
```diff
@@ -3174,3 +3174,55 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3222	TWIG	Agriculture and vegetation	A branch that doesn't arise from the trunk but from another branch.	Person/Thing	
+3223	BRANCH OR TWIG	Agriculture and vegetation	A woody part of a tree arising from the trunk or from another branch.	Person/Thing	
+3224	WRING	Basic actions and technology	To take a cloth and twist its ends into different directions in order to dry it.	Action/Process	
+3225	SQUEEZE OR WRING	Basic actions and technology	To apply pressure from more than one side or to twist a cloth's ends into different directions.	Action/Process	
+3226	FASTEN OR SQUEEZE	Basic actions and technology	To close tightly or to apply pressure from more than one side.	Action/Process	
+3227	SPHERE	Spatial relations	Any spherical object.	Person/Thing	
+3228	BALL OR SPHERE	Spatial relations	An object, generally spherical, used for playing games, or any spherical object.	Person/Thing	
+3229	STINK	Sense perception	To emit an unpleasant odour.	Action/Process	
+3230	DECEIVE	Cognition	To cause someone to believe an untruth; to practice trickery or fraud.	Action/Process	
+3231	DEITY	Religion and belief	A supernatural, typically immortal being with superior powers.	Person/Thing
```
tresoldi: I see, this is related even to Leibniz in terms of the concept GOD above. To make sure people to confuse, I'd have something as GOD (ABRAHAMIC) in the gloss above.

tresoldi: But we should check how things are mapped right now -- are we sure words now mapped to GOD don't actually relate to the DEITY concept?

concepticondata/concepticon.tsv
```diff
@@ -572,21 +572,21 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
 573	SWAY	Motion	To move or swing from side to side; or backward and forward; to rock.	Action/Process	
-574	TRANSIT	Motion	To pass over, across or through something.	Action/Process	
+574	TRANSIT	Motion	To cross something (a river, a street etc.).	Action/Process	
 575	TREAD	Motion	To step or walk (on or over something); to trample.	Action/Process	
 576	DANDRUFF	The body	Small portion of dead skin cells shedding from the scalp.	Person/Thing	
 577	FOREFINGER	The body	The finger between thumb and long finger.	Person/Thing	
 578	KISS	Emotions and values	To touch with the lips to express love or affection.	Action/Process	
 579	BATTLE	Warfare and hunting	An instance of combat in warfare between two or more parties wherein each group will seek to defeat the others.	Person/Thing	
-580	FUR	Clothing and grooming	The hair-covered, dressed pelt of a mammal, used in the making of garments and as trimming or decoration.	Person/Thing	
+580	FUR	The body	The body-hair of an animal or a piece of cothing made thereof.	Person/Thing
```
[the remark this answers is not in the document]
tresoldi: It probably colexifies in most languages anyway.

concepticondata/conceptrelations.tsv
Schweikhard: Actually not. FUR refers to body hair of mammals or to a pelt made of it, whereas HAIR OR FUR only refers to body hair of mammals. Not optimal, but that's what is/was linked to them.

concepticondata/conceptlists/List-2014-1280.tsv
Schweikhard: WOLD has "the speech" (and is linked to the noun) but Key has "speech (make a) [english]; discours (faire une) [french]; discursar  [portuguese]; речь (сделать) [russian]; pronunciar un discurso [spanish]", i.e. it rather seems to refer to the verb, so I left it there. On IDS however, both meanings can be found depending on the language: https://ids.clld.org/parameters/18-222#2/20.3/149.6

concepticondata/concepticon.tsv
Schweikhard: That's better.

concepticondata/concepticon.tsv
Schweikhard: Seems a bit like the issue with "to be alive" (which currently is as "action/process" but which I feel might also rather be "property"). It's about the meaning, not about how it's expressed in a given language after all.

concepticondata/concepticon.tsv
Schweikhard: I changed the original definition because explicitly refers to a center separate from the surroundings, not to the position in the center in general. By "plane or sphere" I meant that this "center" can be both in 2- and 3-dimensional space (whereas our "middle" is only in 1-dimensional space, e.g. in the middle of a row of things).

concepticondata/concepticon.tsv
Schweikhard: Transit includes elicitations like "cross (river)", whereas traverse is "go through (a hole)".

Schweikhard: We might want to change the name of them of course, but I tried to avoid that.

concepticondata/conceptlists/List-2014-1280.tsv
LinguList: Nope, it is entirely not clear where it refers to to me. It is just a very messy concept list we have there. Only portugues has a verb in the gloss, all others have a noun and a verb in brackets. Please link to speech for now. The link is as arbitrary, and we want to keep the links between IDS and WOLD, and key is just ids.

concepticondata/conceptrelations.tsv
Schweikhard: The definitions are rather overlapping instead of proper hyperonyms.
And with the one "HAIR OR FUR"-list we don't even know what it refers to exactly.
We do have 3 for "PELT", though.

concepticondata/concepticon.tsv
LinguList: Making glosses more transparent is the best choice.

Schweikhard: Should this be done by adding a new concept and moving all links to the new concept, or by renaming the concept directly, changing only the name and not the number in the concept lists? (so far I only used the first method except for in one case, as far as I see).

concepticondata/conceptlists/List-2014-1280.tsv
Schweikhard: So, Key, List-2014 and Haspelmath should always be linked to the same concept? In case of clear differences, on which one should I base my choice? Haspelmath is the only one with further information in the form of example sentences of these 3, as far as I see, and the oldest one.

concepticondata/concepticon.tsv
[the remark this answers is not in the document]
Schweikhard: I don't think we have a list that includes both meanings on separate concepts. Buck includes both meanings in the same concept (but with a focus on the latter), the rest are without available definition or only refer to the latter.
And yeah, typo, will change "end" to "beginning". Or we leave it as the original definition, I don't mind either way.

concepticondata/concepticon.tsv
Schweikhard: REMAIN is in spatial relations, too, and this is basically its nominalization.

concepticondata/conceptlists/List-2014-1280.tsv
LinguList: Look, List-2014 is DIRECTLY taken from IDS, and IDS is Key-2016. This is also mentioned in the note, at least the overlap of 1280 concepts is both reflecting IDS and WOLD. IDS is only in translated form, so you COULD always check them. WOLD and IDS also both have their internal IDs, that are interlinked, and those IDs are reflecting that people thought they belong to together, or simply copied them. List-2014 also lists all concept IDs from Key and Comrie, and therefore also WOLD: http://clics.lingpy.org/concepts.php

We may want to add the IDS_KEY to the List-list, but for the time being, just don't make to many changes separating those, it is also not that important, if they are fuzzy. They ARE fuzzy, those lists, especially Key, but we can't change the whole world at once.

concepticondata/concepticon.tsv
Schweikhard: A lot of the linked lists have "let", which would be the second meaning, or both "let" and "permit". We could add a hyperconcept and a new concept for "let" but might be a tad overspecified.

concepticondata/conceptlists/List-2014-1280.tsv
Schweikhard: Yes, that's why I assumed to best base the choice on Haspelmath/WOLD.

concepticondata/concepticon.tsv
Schweikhard: You are right. I think I'll change it to "To address or summon someone". But I'll open an issue for that concept anyway since some of the linked glosses might have even a more different meaning (closer to "scream").

concepticondata/concepticon.tsv
Schweikhard: Borin's definition includes both meanings, Buck distinguishes it from "yield", the others are without explanation. I think my definition could be understood to include both meanings.

concepticondata/concepticon.tsv
Schweikhard: Ah, that might be a good way to distinguish it from TRUE.

concepticondata/concepticon.tsv
Schweikhard: See my comment above.

concepticondata/concepticon.tsv
Schweikhard: One definition we have is close to mine, the other is basically "to estimate".
I guess I'll just include it all, i.e. "To suppose, to estimate or to make a surmise."

concepticondata/concepticon.tsv
Schweikhard: I moved many of the entries to a new concept DEITY. I only left those at GOD that were published before 1900 or have "God" instead of "god".

concepticondata/concepticon.tsv
Schweikhard: See comment below.

concepticondata/concepticon.tsv
Schweikhard: All the other "thin" concepts already had the "very" in their definitions. But I guess it's best to remove it from all of them.

Schweikhard: On second thought, I will remove the "very" only from THIN and from THIN (OF SHAPE OF OBJECT).

concepticondata/concepticon.tsv
Schweikhard: SINK (DESCEND) is the hyperonym. It is perhaps a misnomer as it includes both "cause something/someone else to sink and SINK (DISAPPEAR IN WATER)". We have list that clearly only refers to "cause to sink".

concepticondata/concepticon.tsv
Schweikhard: Only a single list is linked to powerful: https://concepticon.clld.org/values/Dellert-2017-1016-537

concepticondata/concepticon.tsv
Schweikhard: In this I included Key, List and Buck since they have all three words. I could re-link Key and List to STRONG since in Haspelmath the concept is named that.

concepticondata/concepticon.tsv
Schweikhard: I would not make a difference between BROKEN (SHATTERED) and "fragmented".

concepticondata/concepticon.tsv
Schweikhard: > Do we have wordlists which actually make this distinction, between 3193 and 3194?

Buck makes the distinction between PAINT (WALL) ("paint a house") and PAINTING ("paint as artist").
However, I now noticed we already have PAINT (SOMETHING) which is what I meant by PAINT (WALL). I'll remove that new concept and relink all I linked to it to PAINT (SOMETHING) and also adjust the hyperconcept.

concepticondata/concepticon.tsv
Schweikhard: I just took it from the existing definition of DAMAGE. I'll adjust it on both to "To put a thing in bad condition by causing damage to it". We could also just write "To cause damage to something" but that would be not really a definition.

concepticondata/concepticon.tsv
Schweikhard: I'll adjust it to "To remain in the same place after others have left or been removed."

It's about staying at the same position in space, so in that way it fits the category. But I don't know how "spatial relation" is defined, it includes a whole variety of concepts, GATHER, CORNER, ... even HUNCHBACK, which certainly is wrong (changed it to "The body").

concepticondata/concepticon.tsv
Schweikhard: BALL is for the ball used for playing, SPHERE is for anything that is round in 3 dimensions, and BALL OR SPHERE is for the lists that don't specify it. I guess we could drop the last one and make SPHERE a hyperconcept of BALL since most playing balls are also spherical.

concepticondata/conceptlists/List-2014-1280.tsv
Schweikhard: I'll link those 3 to the same concepts, except for when the glosses are called differently in English. For example, I linked "banana tree" on Key and List to BANANA TREE but left it as BANANA on Haspelmath since there it's named "banana" and seems to refer to the fruit.

concepticondata/conceptrelations.tsv
tresoldi: I agree with this, but I can see other cases of overlapping for South American languages (but I cannot find any actual wordlist). In Southern Brazilian Portuguese we have "pelo" which refers to body hair of animals and humans (but not head hair), "pelagem" which is only FUR, "pele/pelego" which is both FUR and PELT (while "pele" is also SKIN).

Once more, the issue is to know that Concepticon does not map reality into words, but establishes links among concept lists. We should only list disjunctive CONCEPT if such disjunction differentiates between at least two concept lists that we included. If new concept lists will demand more work to fix it, than we will work on it -- otherwise, we will fail as every single attempt at an universal language.

concepticondata/concepticon.tsv
tresoldi: Yes, this was clearer later with the MIDDLE concept. Maybe you should add a bit of these descriptions (1-dimensional, 2- or more) in the definitions? It might seem pedantic, but this is the kind of thing that helps people who are aligning conceptlist to concepticon.

concepticondata/concepticon.tsv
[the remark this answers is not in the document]
tresoldi: Ok, good alternative than. But if the lists tend to focus on the latter meaning, I'd swap them and have the "towards the end..." as the second one.

concepticondata/concepticon.tsv
tresoldi: Ok. :+1:

concepticondata/concepticon.tsv
tresoldi: This might be the case of checking if the same language has different entries in different wordlists, reflecting the different meanings. Maybe also change the gloss, to PERMIT OR ALLOW , so that we can have the more specific PERMIT and ALLOW if a new list distinguishes them, in the future?

concepticondata/concepticon.tsv
tresoldi: Great. We knew this would necessarily imply some remapping in the conceptlists, anyway. :+1:

concepticondata/concepticon.tsv
tresoldi: But narrow focus more on the idea of something going through something else (water through straw, person through cave or path, missile through path [narrow shot]), and thing is more of width in relation to the common/expected (thin as slim), correct?

concepticondata/concepticon.tsv
tresoldi: Yes, if it is not in any list, let's keep it out. The elicitations, in any case, probably pick the definition you have. :+1:

concepticondata/concepticon.tsv
tresoldi: :smiley:

concepticondata/concepticon.tsv
tresoldi: Yes, neither do I! My suggestion would be to have it as BROKEN (SHATTERED OR FRACTURED), as fractured implies that something can be fixes/still used.

concepticondata/concepticon.tsv
tresoldi: Both would be good, I guess. As I had in a comment above, my idea is that the definitions are here to help people when linking conceptlist, they are not necessarily definitions (or "good definitions") in a lexicographic sense.

I'm ok with both.

concepticondata/concepticon.tsv
tresoldi: Hm, this is a question for @LinguList

concepticondata/concepticon.tsv
tresoldi: If a more broad definition of "being rude and demean" is not at odds with what we have in Bulakh, I suggest we include it in the Concepticon definition. It would be useful later if other concept lists pick this up.

concepticondata/concepticon.tsv
tresoldi: Nothing linked? Sorry, but where does it come from?

concepticondata/concepticon.tsv
tresoldi: Ah, ok. As for hyperconcept, I'd only have it if needed given some wordlist -- and, as you said, most but not all BALLS are spherical.

concepticondata/concepticon.tsv
tresoldi: Great :+1:

concepticondata/concepticon.tsv
Schweikhard: Don't know. Some lists are linked to ELF OR FAIRY and write "elf, fairy" or similar, and I guess they were originally linked to either ELF or FAIRY, or those concepts were added prematurely in case those single meanings would come up one day.

concepticondata/concepticon.tsv
Schweikhard: Yes, that's why I would put cane (and straw, if it's not a drinking straw) to THIN.

concepticondata/concepticon.tsv
Schweikhard: Seems a bit long as a concept name, though.

concepticondata/concepticon.tsv
Schweikhard: I've now dropped that hyperconcept BALL OR SPHERE and made SPHERE the hyperconcept of ball - at least prototypical ones are spherical. As you said elsewhere, no striving for a universal language ^^

concepticondata/concepticon.tsv
Schweikhard: Bulakh also has a different concept of "to offend" (linked to OFFEND, "To hurt the feelings of; to displease; to make angry; to insult.") and we need to keep those 2 separate somehow.

concepticondata/concepticon.tsv
Schweikhard: IDS at least has both meanings in the languages I understand. I'll add the hyperconcept and ALLOW.

concepticondata/concepticon.tsv
Schweikhard: Ah, I now see what you meant. I thought of a cane: https://en.wikipedia.org/wiki/Walking_stick
and not of a cane: https://en.wikipedia.org/wiki/Cane_(grass)
(albeit the latter is both narrow and thin)

concepticondata/concepticon.tsv
tresoldi: I'd still guess that the more likely elicited answer is a verb than a noun (and this considering that languages really distinguish them and, especially, that the researcher doesn't strip all the morphology), but without any additional material your guess is as good as mine.

So, for consistency, let's keep what you are proposing. We can always change later if needed.

concepticondata/concepticon.tsv
tresoldi: I'd change the gloss: we should be really careful about adding new concepts, especially when there is even a partial overlapping, as Concepticon is not intended to expand to include all possible meanings and concepts (imagine if we were to add even conlangs, like Klingon).

PR #511, opened 2018-09-10
Fix more deprecation warnings

chrzyki:


chrzyki:
'Invalid' escape sequences in string literals get a deprecated warning in 3.6.x and, I **think**, a `SyntaxError` exception in 3.7. Passing the strings raw to `re` avoids that.

xrotwang:
I see. Ok, so they got promoted from "unneeded" to "invalid" :)

xrotwang:
but I guess in that case, I'd rather remove the unneeded backslashes, no?

chrzyki:
Sure, then it'd look like [this](https://github.com/clld/concepticon-data/pull/511/commits/798ae68e135358ccad1fa80e9ea57e16823a715e).

This

```python
re.compile(r'[^\]]\(:(ref|bib):[A-Za-z0-9\-]+\)')
```

would still need `r` because of the` \(` escape sequence. Or would you prefer all raw strings for regular expressions **and** the unnecessary escape sequences removed?

pyconcepticon/check_data.py
```diff
@@ -9,8 +9,8 @@
 
 SUCCESS = True
 NUMBER_PATTERN = re.compile('(?P<number>[0-9]+)(?P<suffix>.*)')
-REF_WITHOUT_LABEL_PATTERN = re.compile('[^\]]\(:(ref|bib):[A-Za-z0-9\-]+\)')
-REF_WITHOUT_LINK_PATTERN = re.compile('[^(]:(ref|bib):[A-Za-z0-9\-]+')
+REF_WITHOUT_LABEL_PATTERN = re.compile(r'[^\]]\(:(ref|bib):[A-Za-z0-9\-]+\)')
```
xrotwang: shouldn't the backslash in the group be removed, too?

pyconcepticon/util.py
```diff
@@ -16,7 +16,7 @@
 
 REPOS_PATH = Path(pyconcepticon.__file__).parent.parent
 PKG_PATH = Path(pyconcepticon.__file__).parent
-ID_SEP_PATTERN = re.compile('\.|,|;')
+ID_SEP_PATTERN = re.compile(r'\.|,|;')
```
xrotwang: Same here: shouldn't the backslash be superfluous here, too?

xrotwang:
Alright. I got confused. Thanks for checking.

Christoph Rzymski <notifications@github.com> schrieb am Mo., 10. Sep. 2018,
20:24:

PR #538, opened 2019-02-06
Fixing one bug of ardila, and adding castro-sui list

LinguList:
This is a rather large list. Chinese helped, the new api is fine, but I was struggling with a stupid bug, as whitespace is not tolerated. Note that, as a convention, we now mark superfluous concepts not linked to anything by an asterisk. I should've used them, but they can be used in a new run, when new lists necessitate new concepts.


xrotwang:
What does this asterisk-marking add over `CONCEPTICON_ID` being empty?

xrotwang:
Ah, I see the asterisk is added to concept set glosses. But concept sets which are linked from no concept are a bug, right? Of course they could happen when a concept set becomes orphaned by re-mapping concepts. But then, I think removing the concept set would be more appropriate - we have the `retired.json` after all.

xrotwang:
I'm not sure about "consecutive and not empty". Why do think this is the case? I think we should rather document that this is a false assumption. Serial primary keys in relational databases have mostly the same properties as concept set ids - including that there is **no guarantee** that for any given primary key all the preceding keys still identify objects.

LinguList:
I find it important for practical purposes. It definitely facilitates generating new ideas when working with larger lists, and also checking that IDs have not been used before, or similar. We don't need to impose this now, but we can maintain this silently, even without saying that it would be needed, just for the about 9 remaining lists to be added, as those asterisked concepts are clearly orphaned and can be replaced without invoking retired.json (which is why I marked them). I can also keep my own personal lists, but since I hope that I won't be the only person submitting new lists for the SI, I'd say: let's keep them there now, all who add new lists and add new concepts use those first for new concepts, and afterwards, we ignore this (as all asterixes have disappeared).

xrotwang:
As far as I can see, the only complication with our current setup - where concept sets can be retired - is that determining the next "free" ID requires looking up `retired.json` as well. But this can easily be encapsulated into a small API method. We do the same for Glottolog languoid IDs.

LinguList:
Yes, but if you want to add some 20 new concepts, this is an extreme pain in the neck. It just took me about an hour to check all the links, making sure the definitions I took from Wikipedia would also be documented in the metadata, and making sure I got the right categories. I would go crazy if I had now to call the api for each of the 20 concepts.

xrotwang:
Ok, I think I could agree to having these asterisked, potential concept sets for work-in-progress in between releases. But then we should have a release task to properly retire all of these before release. This would still mean, that the assumption of "a consecutive range of concepticon IDs" is wrong!

LinguList:
A procedure that WOULD work is another practice that is IN FACT needed:

* adding a new concept list, one observes missing concepts
* I currently mark them with asterisk, to indicate that I later have to add them
* imagine a command that reads in the not-yet-added concept lists, checks basic mappings, and creates new templates for the new concepticon IDs (thus the next *n* lines for `concepticon.tsv`). 

In such a setting, plus a command for single cases, like `concepticon new_id`, this would be convenient, I guess.

LinguList:
In the past, I made some custom script for that. But this may better rest in the pyconcepticon api?

xrotwang:
@LinguList you only have to call the API **once**. This will give you the smallest free ID. From there you can just count.

LinguList:
> @LinguList you only have to call the API **once**. This will give you the smallest free ID. From there you can just count.

Also true ;)

xrotwang:
Actually, we have one more tool available now: Working on concept lists will now happen in `concepticon-dev`. We could accept orphaned concept sets in this dev repos! But moving the complete lists to `concepticon` would also add new concept sets - but ignore the orphaned ones.

LinguList:
Yes, this sounds in fact very cool!

LinguList:
You know what: we should test this with our lists for the Sci-Data project. Start from concepticon-dev, etc., and prepare the new list. I actually wanted to do that today, but was too lazy, also  a bit anxious the new api would not work as expected, etc.

LinguList:
Do you agree with me to merge this one now?

xrotwang:
Yes we can merge. The "*CROSSBOW" would then disappear into `retired.json` if nothing is linked to it when more stuff is merged from `concepticon-dev`.

PR #541, opened 2019-02-21
Tiago

tresoldi:
This PR adds a number of concepts for the `yanglalo` and the `bodtkhobwa` datasets; I am making the pull request on the concepts alone, first, before doing a PR for new conceptlists, so we can discuss innovations in a single place.

It also corrects an error in the previous description of FLYING FOX (a bat, which was confused with FLYING SQUIRREL, the actual Sciuridae).


chrzyki:
Looks good!

tresoldi:
Fixed some wrong concept relations pointed by @Schweikhard

PR #547, opened 2019-03-01
Added IMPEL

Schweikhard:
Split up DRIVE into DRIVE and IMPEL and added a hyperonym DRIVE OR IMPEL OR RIDE as the data in Ringe2002 seems inconsistent in whether it refers to DRIVE or IMPEL or RIDE. Also linked "drive away" in Kraft1981 to DISPEL and adjusted some definitions.

@LinguList DRIVE (CATTLE) contains only one link currently, and the elicitation gloss is Russian: гнать, with the annotation скот - could you check it?


LinguList:
> 
@LinguList DRIVE (CATTLE) contains only one link currently, and the elicitation gloss is Russian: гнать, with the annotation скот - could you check it?

This means "jagen, treiben", but since it is a bit ambiguous in Russian, they added "skot", as I think (skot means cattle).

Schweikhard:
I guess it can stay there then as it is likely slightly more specific than IMPEL.

PR #557, opened 2019-03-18
Update

Schweikhard:
Made some corrections to previous mappings and definitions that came up during mapping Huang.


LinguList:
Nice, I am happy to merge this. Thanks a lot!

PR #564, opened 2019-03-29
Added Carling lists for DIACL mapping

chrzyki:
WIP PR for adding the Carling lists. There might be some editorial stuff that I've missed ...

Curation of mappings has already happened here:
- [Carling-2017-104a](https://github.com/concepticon/concepticon-dev/pull/1)
- [Carling-2017-209b](https://github.com/concepticon/concepticon-dev/pull/2)
- [Carling-2017-112e](https://github.com/concepticon/concepticon-dev/pull/3)
- [Carling-2017-240c](https://github.com/concepticon/concepticon-dev/pull/4)
- [Carling-2017-97d](https://github.com/concepticon/concepticon-dev/pull/5)
- [Carling-2017-10h](https://github.com/concepticon/concepticon-dev/pull/6)
- [Carling-2017-127g](https://github.com/concepticon/concepticon-dev/pull/7)
- [Carling-2017-87f](https://github.com/concepticon/concepticon-dev/pull/8)
- [Carling-2017-87k](https://github.com/concepticon/concepticon-dev/pull/9)
- [Carling-2017-97i](https://github.com/concepticon/concepticon-dev/pull/10)
- [Carling-2017-97j](https://github.com/concepticon/concepticon-dev/pull/11)


concepticondata/conceptlists.tsv
```diff
@@ -224,3 +224,14 @@ Marrison-1967-917	Marrison, Geoffrey Edward	1967		917	basic,areal	English	Naga
 Kolipakam-2018-100	Kolipakam, Vishnupriya and Jordan, Fiona M. and Dunn, Michael and Greenhill, Simon J. and Bouckaert, Remco and Gray, Russell D. and Verkerk, Annemarie	2018		100	basic,areal	English	Dravidian		Kolipakam2018		A 100 item wordlist for Dravidian languages based on [Kassian et al's list](:ref:Kassian-2010-116)		
 Reinhard-1970-337	Reinhard, Johan and Toba, Tim	1970		337	basic,areal	English	Kusunda		Reinhard1970		There are 337 items in the book.		
 Abraham-2018-307	Abraham, Binny and Sako, Kara and Kinny, Elina and Zeliang, Isapdaile	2018		307	basic,areal	English	Tshangla and Khobwa	https://www.sil.org/system/files/reapdata/12/94/06/129406896871137053545983118013436547675/silesr2018_009.pdf	Abraham2018	https://www.sil.org/system/files/reapdata/12/94/06/129406896871137053545983118013436547675/silesr2018_009.pdf	307 item wordlists for Tshangla and Khobwa dialects in Western Arunachal Pradesh.		
+Carling-2017-104a	Carling, Gerd (ed.)	2017	a	104		English		https://diacl.ht.lu.se/WordList/Index	Carling2017		Swadesh 100
```
LinguList: Hi, several things here: author-format follows bibtex-standard, no editor sign. Use rather all official authors of the diacl paper in PLOS one. Then, pleas provide a description, saying, e.g., that this is what they thing is a Swadesh 100 list, and the description should end with a dot, as a normal sentence. Not just two words, ideally link it, like: 

```
According to the authors, this list represens the [100 concepts of Swadesh (1955)](:ref:Swadesh-1955-100). The four additional concepts in our Concepticon are due to ...
```
Maybe check the concepts to see why we have 104 for 100 ?

```
$ concepticon union Carling-2017-104a Swadesh-1955-100
```
The output will give you those concepts occurring only in one list. In fact, this is also some kind of a sanity check.

concepticondata/conceptlists/Carling-2017-104a.tsv
```diff
@@ -0,0 +1,105 @@
...
+Carling-2017-104a-93	93	2110	WARM (OF WEATHER)	warm (of weather)	1093		
+Carling-2017-104a-94	94	948	WATER	water (n)	1094		
+Carling-2017-104a-95	95	1212	WE	we (1pl)	1095		
+Carling-2017-104a-96	96	1236	WHAT	what?	1096		
+Carling-2017-104a-97	97	1335	WHITE	white (colour)	1097		
+Carling-2017-104a-98	98	1235	WHO	who?	1098		
+Carling-2017-104a-99	99	962	WOMAN	woman	1099		
+Carling-2017-104a-100	100	1424	YELLOW	yellow (colour)	1100		
+Carling-2017-104a-101	101	2099	CHILD	child	1101		
+Carling-2017-104a-102	102	1131	WE (INCLUSIVE)	we (1pl incl.)	1103
```
LinguList: Ah, they deliberately extend swadesh by exclusive, inclusive, etc? So here you have what to write in the little description.

concepticondata/references/references.bib
```diff
@@ -2070,3 +2070,10 @@ @book{Reinhard1970
 year={1970},
 address ={Kathmandu}
 }
+
+@misc{Carling2017,
```
LinguList: There is an official publication in PLOS ONE on the project. In case they do not provide an official how-to-quote statement, I urge to use this one, as it represents DIACL. See [here](http://bibliography.lingpy.org/evobib.php?key=Carling2018) for the bibtex, etc.

concepticondata/conceptlists.tsv
LinguList: Ah, more missing: please supply categories, as we do aleady in all other lists, meaning: areal list, basic list, etc. This is also straightforward, but they should not be missing.

concepticondata/conceptlists/Carling-2017-127g.tsv
```diff
@@ -0,0 +1,128 @@
...
+Carling-2017-127g-52	52	932	BARLEY	barley	12052	Crops & fruit	
+Carling-2017-127g-53	53	23	RYE	rye	12053	Crops & fruit	
+Carling-2017-127g-54	54	60	OAT	oats	12054	Crops & fruit	
+Carling-2017-127g-55	55		*LENTILS	lentils	12055	Crops & fruit	
+Carling-2017-127g-56	56	605	GRAIN	grain (generic)	12056	Crops & fruit	
+Carling-2017-127g-57	57	3062	TURNIP	turnip	12057	Crops & fruit	
+Carling-2017-127g-58	58	1320	APPLE	apple	12058	Crops & fruit	
+Carling-2017-127g-59	59	904	BERRY	berry	12059	Crops & fruit	
+Carling-2017-127g-60	60	920	GRAPE	grape	12060	Crops & fruit	
+Carling-2017-127g-61	61		*FLAX	flax	12061	Crops & fruit
```
LinguList: Okay, sorry, more work now, but **all** concepts with an asterisk need to be added NOW to concepticon, otherwise, it doesn't make sense to provide them. If you add concepts with OR, you also need to make the linking. So this will take a bit more time, but you have the advantage of being able to make this quickly with grepping the new definitions from the concept lists. Also a consistency check, and I don't mind if you leave certain new conceps out, but then remove the starred gloss.

concepticondata/references/references.bib
xrotwang: The PLOS ONE article focusses on the Typology side of DIACL, though. So in terms of a database description, [this PDF](https://diacl.ht.lu.se/Content/documents/DiACL-lexicology.pdf) (linked from the database landing page) is probably the best we can get.

LinguList:
I guess, to see why those checks fail, you ideally run `concepticon test` locally...

concepticondata/references/references.bib
LinguList: I'd still say that the misc is not enough to quote it, right? I'd say: quote it as a `@book`, and give the university as publisher, etc., but also include all people involved. Ah, just saw the PDF, this also gives the authors.

concepticondata/conceptlists/Carling-2017-127g.tsv
chrzyki: Awesome, thanks for the feedback! Just wanted to 'do' the PR now so that I could work on this progressively. Thanks for reporting all the issues so quickly. :)

Schweikhard:
I gave the mapping a quick check, and most looks good to me! Just some suggestions:

Should we really link "wine (rice)" to WINE in general?
I would rather link "rice (grain)" to RICE (UNCOOKED) than to RICE in general as there is also cooked rice in the list.
I don't understand why black is sometimes mapped to BLACK, and sometimes to BLACK OR DARK.
WASH is transitive (by the definition and by having "wash clothes" as a hyponym), so I don't think "to wash/to bath" should be linked to it - there are some concepts linked to it that would need to be moved to a reflexive concept, though, or the definition changed.
We already have POINT (POINTED END), no need to add that as a new concept.

And, nothing to do with the mapping itself, but we might want to change the definition of OAR now we are adding a hyperonym as it's currently not really differentiating it from PADDLE, and move some things from PADDLE to either OAR or the new hyperonym.

chrzyki:
Thanks @Schweikhard, will check! POINTED is fixed in the commit that will follow in a couple of minutes.

concepticondata/conceptlists/Carling-2017-127g.tsv
chrzyki: @LinguList @Schweikhard see https://github.com/clld/concepticon-data/pull/564/commits/ecdbb943f68290c1948bd4453974da8a4303bf0d.

Schweikhard:
That's some nice new concepts - could you also update conceptrelations.tsv with the new hyp(er)onym-relations?

Schweikhard:
Oh, OAR OR PADDLE is for PADDLE or RUDDER? If that's the case I would suggest you rename it, if not, the definition.

chrzyki:
> That's some nice new concepts - could you also update conceptrelations.tsv with the new hyp(er)onym-relations?

Thanks! Will do.

chrzyki:
> Oh, OAR OR PADDLE is for PADDLE or RUDDER? If that's the case I would suggest you rename it, if not, the definition.

Well, its either or, not differentiating between:

> The difference between oars and paddles are that oars only have one blade, and are used exclusively for rowing, whereas paddles can have either one or two blade and are not rowed.

Schweikhard:
Paddles by the definition of a list linked are small oars without an oarlock - is that what you mean by "are not rowed"?
I assumed you were thinking of rudders (Steuerruder) as your definition for OAR OR PADDLE is because you wrote "either for paddling _or navigation._"

chrzyki:
I hope I've addressed most of immediate issues and problems. The list descriptions aren't particularly 'cool' :) because they're only providing only few information themselves.

Thanks for the help @Schweikhard and @LinguList.

chrzyki:
> Yes, so I agree with @Schweikhard that we need the concept relations for all concepts with an OR, but that should then be okay.

Thanks! Added a number of relations. All the other things have been settle I think/hope. :)

Schweikhard:
STREAM is already a hyponym of RIVER (i.e. a small river) - by that, having a hyperonym of those two is redundant and you can link that "river, stream, brook" to RIVER.

The better option however might be to instead remove that hyponym relation between STREAM and RIVER.

Is LEATHER OR HIDE a hyperonym of SKIN (ANIMAL)?

chrzyki:
> STREAM is already a hyponym of RIVER (i.e. a small river) - by that, having a hyperonym of those two is redundant and you can link that "river, stream, brook" to RIVER.
> 
> The better option however might be to instead remove that hyponym relation between STREAM and RIVER.
> 
> Is LEATHER OR HIDE a hyperonym of SKIN (ANIMAL)?

For RIVER: Since the elicitation gloss in the original is one of those 'catch all' kind of things and could potential refer to different sizes of flowing I'd be against linking all of river, stream, and brook to just RIVER.

LEATHER OR HIDE: I find this one weird because it either refers to the thing coming from the animal after being manually processed (e.g. tanned) or while still 'on the animal'. As such, I don't see a real relationship to SKIN exclusively.

Schweikhard:
I agree with both and have removed the RIVER narrower STREAM relation.
Some concepts linked to RIVER better fit the new hyperonym.

chrzyki:
> I agree with both and have removed the RIVER narrower STREAM relation.
> Some concepts linked to RIVER better fit the new hyperonym.

Thanks!

chrzyki:
Will wait for a final 'ok' and then I'll merge. Thanks for all the help!

LinguList:
Do we have the conceptrelations now? I didn't find them. If they are there, I'm fine with it and you can merge!

chrzyki:
Yep, here: https://github.com/clld/concepticon-data/pull/564/files#diff-63bb61c43e529bc566c18911fe6be401

Issue #572, opened 2019-04-03
Typos: Yakhontov 1991 and Starostin 1991, 2007, and 2010


natalia-morozova:
I have noticed some typos in Russian, so I checked the [source paper](http://starling.rinet.ru/Texts/Starostin_AP.pdf) and did not find those typos there. 

Since the lists by Yakhontov and Starostin are connected, the typos are consistent throughout all the lists. Please have a look:
[Yakhontov 1991 100](https://concepticon.clld.org/contributions/Yakhontov-1991-100)

| CONCEPTLIST | ID | ENGLISH | RUSSIAN | CORRECTED |
| ------------- | ------------- | ------------- | ------------- |------------- |
| Yakhontov 1991 100 | [Yakhontov-1991-100-17](https://concepticon.clld.org/values/Yakhontov-1991-100-17) | earth | земль | земл**я** |
| Yakhontov 1991 100 | [Yakhontov-1991-100-27](https://concepticon.clld.org/values/Yakhontov-1991-100-27) | full | полны | полны**й** |
| Yakhontov 1991 100 | [Yakhontov-1991-100-44](https://concepticon.clld.org/values/Yakhontov-1991-100-44) | long | длинний | длинн**ы**й |
| Yakhontov 1991 100 | [Yakhontov-1991-100-53](https://concepticon.clld.org/values/Yakhontov-1991-100-53) | near | блиский | бли**з**кий |
| Yakhontov 1991 100 | [Yakhontov-1991-100-78](https://concepticon.clld.org/values/Yakhontov-1991-100-78) | star | звесда | зве**з**да |
| Yakhontov 1991 100 | [Yakhontov-1991-100-100](https://concepticon.clld.org/values/Yakhontov-1991-100-100) | yellow | желты | желты**й** |

[Yakhontov 1991 65](https://concepticon.clld.org/contributions/Yakhontov-1991-65)

| CONCEPTLIST | ID | ENGLISH | RUSSIAN | CORRECTED |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| Yakhontov 1991 65 | [Jachontov-1991-65-12](https://concepticon.clld.org/values/Jachontov-1991-65-12) | earth | земль | земл**я** |
| Yakhontov 1991 65 | [Jachontov-1991-65-29](https://concepticon.clld.org/values/Jachontov-1991-65-29) | long | длинний | длинн**ы**й |
[...]

PR #575, opened 2019-04-03
Update Yakhontov-1991-35.tsv

natalia-morozova:
#572


chrzyki:
Will check how to proceed with the ID changes.

PR #582, opened 2019-04-04
consistent concept IDs

xrotwang:


SimonGreenhill:
Looks like this should be an automated check that raises an error, right? are there any good reasons for where the concept list ID does not match the filename?

xrotwang:
@SimonGreenhill don't think so. Plain ASCII for the concept list IDs seems to be at least common sense - if not stated policy.

PR #584, opened 2019-04-04
correcting problems in mitter-hofer-bena

LinguList:


LinguList:
forgot to merge, I now also added hale.

PR #589, opened 2019-04-04
Metadata fixes

chrzyki:
Adds a number of automatically generated (default) metadata files and fixes minor metadata issues.

Most of the new files were automatically generated (but I still had a quick look at the lists to check for consistency and rule out weirdness). Should I generate the other (possibly default) metadata files as well or just go for the lists with custom columns?


xrotwang:
What's the idea behind this? I'd agree that the "we lookup the metadata somewhere else if it's missing" approach is somewhat intransparent. But if we go fully explicit, we should also remove the lookup functionality in pyconcepticon.

LinguList:
If we have to provide metadata with each new concept list, this will be a bit painful, so we'd need a simple method to create it.

xrotwang:
So far, the only use case of the metadata is for extended, custom validation. E.g. for the WOLD list, the metadata specifies that the non-standard column `URL` is of datatype `anyURI`; or for ABVD, the metadata could specify a `valueUrl` property, specifying how to turn the local ID into a URL to the website. Unless we plan on doing this more often, I wouldn't bother with this right now.

xrotwang:
Oh, and if we do add metadata, it should be for **all** lists and **all** columns in each list.

xrotwang:
Right now, I count:
```
(concepticon) forkel@shh.mpg.de@dlt4803010l:~/venvs/concepticon/concepticon-data$ la concepticondata/conceptlists/*.tsv | wc -l
239
(concepticon) forkel@shh.mpg.de@dlt4803010l:~/venvs/concepticon/concepticon-data$ la concepticondata/conceptlists/*.json | wc -l
191
```

LinguList:
I agree, we are in a state where we can use it, also to interlink between concept lists, but doing it for all lists would be a bit too much (also for contributors).

chrzyki:
Well, right now it is a bit inconsistent: we have default metadata (i.e. explicit) for many lists and no explicit data for others.

Either way, I could filter the automatically created default lists in this PR and just add the non-default ones or be 100% explicit.

On 5 Apr 2019 08:16, Johann-Mattis List <notifications@github.com> wrote:

If we have to provide metadata with each new concept list, this will be a bit painful, so we'd need a simple method to create it.

chrzyki:
That's what is partially done in this PR. Should be all columns for all lists with 50 or so that could be added as well.

On 5 Apr 2019 08:25, Robert Forkel <notifications@github.com> wrote:

Oh, and if we do add metadata, it should be for all lists and all columns in each list.

LinguList:
If we have 239, we need to add at least one small list before release of 2.0!

xrotwang:
239 is prime - that's as cool as it gets. Prime numbers are created by God, all the others are made by humans.

LinguList:
Given that I already prevented the 13 new datasets in lexibank yesterday, I'll see to achieve the beautiful human-made number of 240 lists, which mirrors the number of hours 10 days have and has so may nice ways to be divided by prime numbers.

xrotwang:
AFAIKT it has **exactly one** way to be divided by prime numbers.

chrzyki:
Cool that the PR about metadata turns into a meta discussion about numbers. :)

I vote in favour of adding metadata for all lists and being 100% explicit since this gives us more leeway for adding norm data later on to lists. I could easily add the other metadata files today. I also like the idea of having these information available further down the road.

On 5 Apr 2019 08:32, Johann-Mattis List <notifications@github.com> wrote:

Given that I already prevented the 13 new datasets in lexibank yesterday, I'll see to achieve the beautiful human-made number of 240 lists, which mirrors the number of hours 10 days have and has so may nice ways to be divided by prime numbers.

xrotwang:
Hm. I'm still not sure about the added value. I guess it would make sense, if we spent a bit of effort turning the metadata into useful schemas, e.g. specifying things like a regex for the local IDs of IDS concepts - which will make our integrity tests better.

LinguList:
Yes, I agree with @xrotwang that it may be too early (?). It would also require a review of column contents, which we have done some time ago, but maybe we don't have the time, to make this a requirement for the future.

xrotwang:
It still won't be required when contributing new lists. The required columns in these can still be read using the default metadata. So in terms of the workflow, nothing would change.

xrotwang:
I see now that pyconcepticon also infers additional data in concept lists solely based on the metadata. So right now, additional columns in lists without explicit metadata would go undetected.

LinguList:
If nothing changes, I'm fine with whatever solution.

xrotwang:
Ok. Unless @chrzyki wants to do that, I can have a look at creating the explicit metadata.

xrotwang:
Seems to be useful, I guess. I already detected one bug:
```
--- a/concepticondata/conceptlists/Mitterhofer-2013-300.tsv-metadata.json
+++ b/concepticondata/conceptlists/Mitterhofer-2013-300.tsv-metadata.json
@@ -43,9 +43,7 @@
                         "name": "ENGLISH"
                     },
                     {
-                        "datatype": "string", 
-                        "valueUrl": "Mitterhofer-2013-95-{MITTERHOFER_2013_95}", 
-                        "name": "MITTERHOFER_2013_95"
+                        "name": "MITTERHOFER_2013_92"
                     }
                 ],
                 "primaryKey": [
diff --git a/concepticondata/conceptlists/Mitterhofer-2013-355.tsv-metadata.json b/concepticondata/conceptlists/Mitterhofer-2013-355.tsv-metadata.json
index 23ee274..3e6199a 100644
--- a/concepticondata/conceptlists/Mitterhofer-2013-355.tsv-metadata.json
+++ b/concepticondata/conceptlists/Mitterhofer-2013-355.tsv-metadata.json
@@ -48,9 +48,7 @@
                         "name": "MITTERHOFER_2013_300"
                     },
                     {
-                        "datatype": "string", 
-                        "valueUrl": "Mitterhofer-2013-95-{MITTERHOFER_2013_95}", 
-                        "name": "MITTERHOFER_2013_95"
+                        "name": "MITTERHOFER_2013_92"
                     }
                 ],
                 "primaryKey": [
```

Apparently the column name was changed in the metadata file, but not in the tsv file.

So yes, I'd say we add metadata for all lists, but also include a check, that makes sure column names in metadata and in the list itself are consistent.

chrzyki:
Yep, I changed that here https://github.com/clld/concepticon-data/pull/589/files#diff-cd029401b5b13027e3253a3cea0ee13d and wanted to ask how to proceed with that but forgot about it this morning, sorry ...

chrzyki:
Closing this, @xrotwang is looking into it. Thanks!

PR #592, opened 2019-04-05
update

### LINKED ISSUE #440: LIVING vs. BE ALIVE: systematic checking of glosses
this is in order, as we seem to have instances where BE ALIVE is used but LIVING should be used instead...

### LINKED ISSUE #552: EARTH WET SUBSTANCE is identical with EARTH SOIL
Or what is the difference?

Schweikhard:
fixes #440
fixes #552


LinguList:
Your error:
```
ERROR:Dellert-2017-1016: wrong conceptset GLOSS for ID 1244: BE ALIVE -> THICK
```

LinguList:
Nice, thanks!

Schweikhard:
Fixed it, just a Zahlendreher.

Waiting for your reply on https://github.com/clld/concepticon-data/issues/549
but besides that, I'm done with my part for this release I think.

Schweikhard:
I've removed the redundant sameas conceptrelations after checking that they are actually also specified in concepticon.tsv (and adding them there if not).
There is now only one sameas relation left (legendary being and mythical creature) because here an older concept was replaced by a newer.

Issue #603, opened 2019-04-18
Issues with typos, IDs, and mapping in Pallas 1786 and 1789


natalia-morozova:
There are some issues with the lists by Pallas. I have a few things to report.

1. Pallas 1786 441

- IDs in Pallas 1786 441 start with [Pallas-441-2](https://concepticon.clld.org/values/Pallas-1786-441-2) instead of **Pallas-441-1**, thus, the last ID is [Pallas-441-442](https://concepticon.clld.org/values/Pallas-1786-441-442) instead of 441.

- Some entries were mixed up in the process of digitalization in [Pallas 1786 441](https://concepticon.clld.org/contributions/Pallas-1786-441), thus one Russian gloss **Ѣшь** is missing. Here is how it looks like on Concepticon:

|  ID | FRENCH | LATIN |RUSSIAN | GERMAN|
| ------- | ------- | ------- |------- |------- |
| [Pallas-1786-441-111](https://concepticon.clld.org/values/Pallas-1786-441-111) | Tu mange | Edis | ѣмъ | Ißßeßt |
| [Pallas-1786-441-112](https://concepticon.clld.org/values/Pallas-1786-441-112) | Il mange | Edit | Ѣсть | Ißßet |

And this is how it is in the [source provided for the list](https://cdstar.shh.mpg.de/bitstreams/EAEA0-788B-3A7A-F5BD-0/Pallas1786.pdf) (page 4, towards the end of the left column):

| ID | FRENCH | LATIN |RUSSIAN | GERMAN|
| ------- | ------- | ------- | ------- |------- |
| Pallas-1786-441-111 | Je mange | Edo | ѣмъ | Eßßen |
| Pallas-1786-441-112 | Tu mange | Edis | Ѣшь | Ißßeßt |
| Pallas-1786-441-113(?) | Il mange | Edit | Ѣсть | Ißßet |

If the corrections are accepted, all following IDs should be changed correspondingly. 

- I have also noticed some mapping errors:

| ID | ENGLISH GLOSS | CONCEPT SET | CORRECTED |
| ------ | ------ | ------ | ------ |
| [Pallas-1786-441-38](https://concepticon.clld.org/values/Pallas-1786-441-38) | nail (noun) | [502 NAIL (TOOL)](https://concepticon.clld.org/parameters/502) | [1896 FINGERNAIL OR TOENAIL](https://concepticon.clld.org/parameters/1896) |
[...]

PR #604, opened 2019-04-23
Pallas 1786 and 1789 updates.

natalia-morozova:
closes issue #603


LinguList:
Ah, we have an ugly change in the list ID now. That means, that you'll  have to do two things tto fix this:

1. change the conceptlists.tsv to modify the ID of the list Pallas-DATE-442 (instead of 441).
2. add a replacment statement to the "retired.json". 

The latter is in fact a bit annoying, but if you are good in regular expressions, you could do it yourself. If not, let us know and @chrzyki or me will look into this afterwards. In this case, open an issue that mentions that we need to work on retired.json and assign it to @chrzyki or myself or us both.

natalia-morozova:
Got it, I would like to do it on my own (with @chrzyki 's assistance, of course).

LinguList:
Good spirit!

SimonGreenhill:
great work @natalia-morozova 🎉

Issue #529, opened 2018-12-10
Veijdemo-2016-167


LinguList:
This is an intereting list, containing 167 "content words" of Swadesh's list of 200 items, and adding meta-data, age of acquisition, etc.

https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0147924

LinguList:
Ah: contains the supplement as well, in csv format.

chrzyki:
Giving this to @kristina-pianykh as an easierish introductory list (I need to get a better grasp of the Muna-Buton stuff myself). This should be a quick and easy one (given the CSV supplementary) and also pushes us further towards finally getting started with the norm data idea. ;)

PR #613, opened 2019-06-21
Vejdemo list

Kristina-Pianykh:
Vejdemo's list (2016, 167 items) as referred to in issue #529.


SimonGreenhill:
Well done @Kristina-Pianykh  🎉

concepticondata/conceptlists/Vejdemo-2016-167.tsv
```diff
@@ -0,0 +1,168 @@
...
+Vejdemo-2016-167-5	BARK	1204	5	bark of a tree	3.79	0.275929132	0.161204268	0.063386155	0.434200743	0.201591512	0.519262982	8	5	4.52	2366	3.4	1		Noun
+Vejdemo-2016-167-6	BELLY	1251	6	belly	4.39	0.618613538	0.403010671	0.253544621	0.542750929	0.907161804	0.986599665	32		4.8	1.9295	3.75	5	4.05	Noun
+Vejdemo-2016-167-7	BIG	1202	7	big	3.41	2.079123691	2.06878811	4.500417014	1.356877323	0.75596817	1.713567839	1119	4.5	3.66	5388	4.33	13	2.89	Adjective
+Vejdemo-2016-167-8	BIRD	937	8	bird	2.15	0.237426512	0.161204268	0.443703086	0.325650558	0.100795756	0.155778894	63	6.4	5	7.5535	3.83	5	3.52	Noun
+Vejdemo-2016-167-9	BLACK	163	9	black	1.91							203	6.4	3.76	7.2595	3.58	5	3.56	Adjective
+Vejdemo-2016-167-10	BLOOD	946	10	blood	2.19	0.499507569	0.806021341	0.063386155		0.453580902	0.675041876	132	6.4	4.86	8647	5.76	5	4.89	Noun
+Vejdemo-2016-167-11	BONE	1394	11	bone	1.34	0.582561615		1.077564637	0.325650558	0.25198939	0.675041876	46	6.4	4.9	7.55	4.75	3	5.53	Noun
+Vejdemo-2016-167-12	CHILD (YOUNG HUMAN)	1304	12	child young	6.31	1.438809674	0.725419207	2.408673895	1.953903346	0.75596817	1.350083752	626	6.4	4.78	5838	5.33	4	5.15	Noun
```
chrzyki: I remember that the original list had `_` and you removed most of those which is good. However, just having no delimiter at all looks weird, i.e. I'd prefer `bark (of a tree)` or `stone, rock` etc.

concepticondata/conceptlists.tsv
```diff
@@ -239,3 +239,4 @@ Carling-2017-97j	Carling, Gerd and Larsson, Filip and Cronhamn, Sandra and Farre
...
 Yang-2011-1014	Yang, Cathryn	2011		1014	questionnaire	Chinese	Lalo Languages	https://www.sil.org/system/files/reapdata/45/63/47/45634767234331504727329755903316120678/Yang_Lalo_Sept_2011.pdf	Yang2011		A survey of Lalo regional varieties for research in phylogenetics, dialectometry and sociolinguistics.		
+Vejdemo-2016-167	Vejdemo, Susanne and Hörberg, Thomas	2016		167	basic	English	Global	https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0147924#pone.0147924.ref032 	Vejdemo2016		This list represents 167 content words, derived from [the 200-item Swadesh list (1952)](:ref:Swadesh-1952-200), with a few concepts coming from  [Swadesh’s list of 215 items (1955)](:ref:Swadesh-1955-215). It was originally compiled with the purpose of assessing the rate of lexical replacement for the specified concepts in English, Danish, Dutch, German and Swedish, according to word class, frequency, age of acquisition, synonyms, arousal, imageability and average mutual information either from published databases or from corpora and lexica.
```
LinguList: `Swadesh’s` should be spelled with `'` instead of `’` as we have been using normal apostroph throughout.

concepticondata/conceptlists/Vejdemo-2016-167.tsv
LinguList: You mean the delimiter of the English gloss? I guess that should reflect the original as faithfully as possible, right? So I'd say you better render this literally. Or did I not understand something here? So if the original elicitation gloss has a `_`, please keep it.

LinguList: Yes, specifically because in some cases, the underscore is still there. So please let's go for underscores here, as in the original.

concepticondata/conceptlists/Vejdemo-2016-167.tsv
```diff
@@ -0,0 +1,168 @@
...
+Vejdemo-2016-167-69	RIVER	666	69	river stream brook	3.69	0.614277532		0.380316931	0.488475836	0.705570292	0.882747069	115		4.89	8.6355	4.22	1		Noun
+Vejdemo-2016-167-70	ROAD	667	70	road	4.86	1.511196739	0.349275915	3.169307756	1.573977695	0.957559682	1.505862647	318	6.6	4.75	8.3685	3.81	2	4.55	Noun
+Vejdemo-2016-167-71	ROOT	670	71	root	1.75	0.629759917	0.564214939	1.077564637	0.379925651	0.50397878	0.623115578	44	6	4.34	5681	3.62	8	5.94	Noun
+Vejdemo-2016-167-72	ROPE	1218	72	rope	6.09	0.487999433	0.188071646	0.316930776	0.705576208	0.554376658	0.675041876	29	6.6	4.93	4.8825	4.58	2	5.44	Noun
+Vejdemo-2016-167-73	ROTTEN	1728	73	rotten	4.76	1.421712038	2.176257622	0.760633862	1.465427509	1.51193634	1.194304858	8		3.87	4004	4.9	3	6.95	Adjective
+Vejdemo-2016-167-74	SALT	1274	74	salt	1.01	0.260085729	0.376143293	0.190158465	0.271375465	0.151193634	0.311557789	25	6.2	4.89	7.9325	4.53	3	5.05	Noun
+Vejdemo-2016-167-75	SAND	671	75	sand	4.26		0		0	0	0	41	6.7	5	8.0265	3.43	2	4.63	Noun
+Vejdemo-2016-167-76	SEA	1474	76	sea ocean	2.53	0.805628083	0.537347561	0.824020017	0.92267658	0.705570292	1.038525963	127	6.4	4.79	8272	2.8	3		Noun
```
LinguList: we have sea or ocean: http://concepticon.clld.org/parameters/2282 2282, SEA OR OCEAN.

concepticondata/conceptlists/Vejdemo-2016-167.tsv
chrzyki: Sorry, for some reason I didn't think this through and actually recommended removing those back then. Apologies!

concepticondata/conceptlists/Vejdemo-2016-167.tsv
Kristina-Pianykh: Shall I then map it to the SEA OR OCEAN (2282) concept set? I took [the Swadesh list (1952)](https://concepticon.clld.org/contributions/Swadesh-1952-200) as a reference, where this same concept is mapped to [the SEA concept (1474)](https://concepticon.clld.org/parameters/1474). In contrast, the SEA OR OCEAN concept set is mapped to just 3 concepts and, for this reason, I thought it might be somewhat inconsistent since the Vejdemo list is almost entirely based on the 200-item Swadesh's list.

Kristina-Pianykh:
I've looked through the text and there aren't any references to the IDS or WOLD projects. Instead, it says the current study uses 167 out of 173 lexical items, listed in the study by Pagel et al. (2007), which in turn builds upon the data from Dyen, James & Cole 1992:

> The data concerns 167 of the 173 items listed as open class words in [2].
By [2] meaning Pagel et al. (2007).

>  Pagel et al. [2] calculated a relative rate of lexical replacement for the primary words (cf. primary designating expressions in [7]) of the 200 concepts of the Swadesh list, based upon data from Dyen, James & Cole [8] on the frequency of change of these concepts in Indo-European language varieties.

LinguList:
Okay, @Kristina-Pianykh, the list is actually provided in concepticon already, i.e., the list by Pagel et al. 2007, so this is where you should compare. And it seems this is clearly teh list they take it from, so no need to mention Swadesh, but rather this list, as it is the direct "ancestor". The "rock" is strange, but the rest seems like the Pagel-2007 list. So I'd say: change the text accordingly, and this is it (only mention Pagel-list).

concepticondata/conceptlists/Vejdemo-2016-167.tsv
Kristina-Pianykh: @LinguList since we've figured out that the Vejdemo list is built upon the list by [Pagel et al. (2007)](https://concepticon.clld.org/contributions/Pagel-2007-200), would it still make sense to map the concept sea_ocean to the concept set [SEA OR OCEAN (2282)](https://concepticon.clld.org/parameters/2282) given that the exact same concept in Pagel's list is mapped to [SEA (1474)](https://concepticon.clld.org/parameters/1474)?

LinguList: You're right, you could in fact just follow all mappings we used for pagel 2007!

LinguList:
Hi there, it's been a week since we last discussed this. @Kristina-Pianykh, if you want me to check again, if my requests have been addressed, let me know, otherwise, I won't react and just wait.

concepticondata/references/references.bib
```diff
@@ -2099,3 +2099,16 @@ @Phdthesis{Yang2011
   Note                                = {PhD Thesis},
   Address                             = {Bundoora}
 }
+
+@Article{Vejdemo2016,
+  Title                               = {{S}emantic {F}actors {P}redict the {R}ate of {L}exical {R}eplacement of {C}ontent {W}ords},
```
LinguList: we usually never capitalize the text, only if its place names, etc., so the {} brackets should be removed here.

concepticondata/references/references.bib
```diff
@@ -2099,3 +2099,16 @@ @Phdthesis{Yang2011
...
+  Author                              = {Vejdemo, Susanne and Hörberg. Thomas},
+  Journal                             = {PLoS ONE},
+  Year                                = {2016},
+  Volume                              = {11},
+  Number                              = {1},
+  Doi                                 = {10.1371/journal.pone.0147924},
+  Url                                 = {https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0147924#pone.0147924.ref032}
+}
```
LinguList: we add page numbers in journals like plos as well, just use the page numbers of the manuscript.

concepticondata/conceptlists.tsv
```diff
@@ -239,3 +239,4 @@ Carling-2017-97j	Carling, Gerd and Larsson, Filip and Cronhamn, Sandra and Farre
...
 Yang-2011-1014	Yang, Cathryn	2011		1014	questionnaire	Chinese	Lalo Languages	https://www.sil.org/system/files/reapdata/45/63/47/45634767234331504727329755903316120678/Yang_Lalo_Sept_2011.pdf	Yang2011		A survey of Lalo regional varieties for research in phylogenetics, dialectometry and sociolinguistics.		
+Vejdemo-2016-167	Vejdemo, Susanne and Hörberg, Thomas	2016		167	basic	English	Global	https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0147924#pone.0147924.ref032 	Vejdemo2016		This list represents 167 content words, built upon the list by [Pagel et al. (2007)](:ref:Pagel-2007-200). It was originally compiled with the purpose of assessing the rate of lexical replacement for the specified concepts in English, Danish, Dutch, German and Swedish, according to word class, frequency, age of acquisition, synonyms, arousal, imageability and average mutual information either from published databases or from corpora and lexica.
```
LinguList: As this is not only a basic list, but also a list with numbers, please add "acquisition" as tag, and compare with similar lists of this types, maybe even adding a new tag "speech norms". @chrzyki, please could you see that you help, as you added a similar list with speech norms before?

Kristina-Pianykh: @LinguList I have just looked at the lists with the acquisition tag on Concepticon and they are all based on the studies specifically focused on the acquisition of the listed concepts by children. In contrast, this particular study by Vejdemo investigates the likelihood of lexical replacement ("borrowability") of the concepts, based on a range of criteria like frequency, imageability, arousal and the age of acquisition, among them. So technically, it's an analysis of the stability of these concepts in a number of languages. The ranked tag also seems appropriate since Pagel's list which is the basis for Vejdemo's list is also tagged as ranked

concepticondata/references/references.bib
LinguList: sorry, but this is not what I meant: I meant that you also need to write them as small, obviously: bibtex uses the `{T}` to indicate capitalization, but our systems ignore this, so you need to write in small capitals to get this right, as it will still be a capitalized title, which is not the standard style. It is possible that our bibliography is inconsistent. I go through it once a while, but this can obviously be avoided, if we do it more carefully generally upon adding a new list. But if this is done, you can merge it

LinguList: Obviously, the pages are also still missing, right? Just use the page nmbers in hte paper: 1-15. Again, one could say this is nitpicking, but we are linguists, and we need to insist on good bibliographies, physics or biologists can do without it, if they prefer, but clean bibliographies are part of our standard. So just add the page-range (1-15, obviously), and this is fine.

PR #622, opened 2019-07-22
concept relations: hyponym

Schweikhard:
As discussed before, I renamed most of the "narrower" and all of the "instanceof" concept relations to "hyponym" and only left those like ARM OR HAND as "narrower".
Some were a bit difficult to decide (HAIR (BODY) for example), but I only left those as "narrower" where I was fairly sure they should be.
Also fixed some mistakes I noticed while going through the list (mainly redundant or swapped entries).


LinguList:
sorry, but when did we discuss this? This needs to be hard-coded, so we can't do that now, and we have some reason for use of narrower, broader. So I am afraid that we'll have to close this for now, as it is against our current policy and would require a discussion specifically also with @xrotwang and @SimonGreenhill

Schweikhard:
Okay, then I'll leave it in my fork for now. We discussed it about 2 months ago. I must have misunderstood that it also needs a change to the code.

LinguList:
I don't remember that I sanctionized the renaming of all those existin grelations, so I think this is a misunderstanding. What you can do without problems is ADDING those relaions, with hyponym or whatever, as long as you leave the narrower/broader in there. But you can't delete the ones we have now, also because we have tests that will check for those relations.

Schweikhard:
Ah, I see - I can correct that quickly.

PR #623, opened 2019-07-22
concept-relations

Schweikhard:
I hope it's correct now - in the pre-existing rows I made some corrections (redundant and swapped stuff) and I added new rows for those concept relations that are hyponymic and not metonymic.


LinguList:
Nope, you break the rules. See [here](https://travis-ci.org/concepticon/concepticon-data/builds/562123922?utm_source=github_status&utm_medium=notification). The crucial test is that Swadesh-1952-200 and Swadesh-1955-100 together have 207 concepts. We capture this with the narrower-relations, as the concepticon-ids would be 213. Your modifications breaks this, so it breaks our test, so it breaks the original criterion why we instantiated the concept relations.

Schweikhard:
Okay, then I'll just add the corrections and remove the hyperonyms for now.

LinguList:
Can you run the check on your locak computer before? It's fast and just takes `concepticon test` in commandline, will immediately have you see if it's okay, before you submit?

Schweikhard:
Thanks for reminding me of that, I'll do so from now on.
There seem to be no conflicts left now.

xrotwang:
Looks good. All the corrections make sense, and the removals probably, too. I guess for starting out towards more systematic relations, removals are good anyway :)

LinguList:
After this misunderstanding has been solved now, as I hope, @Schweikhard, you could start with your new relations in a separate file, to see if it makes sense to do this more systematically. If you then use this to create some network, you can investigate it. It would be excellent to have a clear hierarchy or some ontology on top of concepticon, but we haven't had the time to follow up, also because it is growing so rapidly, so if you want to look into that, you do us a great favor, but lets not do the big changes in a single PR, let's first maybe discuss a relation hierarchy you develop (the one you just had there) on some spreadsheet, and then discuss how to integreate it for the next release?

SimonGreenhill:
All good here!

Schweikhard:
Sure - I'd like to avoid too many types of relations, but there certainly are different forms of hyperonymic relations, so if this is a major change in any case, I can also take a look which of those would be straightforward enough to include them in the list.

LinguList:
The problem is that people mess up with concept comparison for multiple reasons, and the idea of a concept hieararchy is not primarily dealin with the question why people mess up. Take CHILD for example: Swadesh has it in two versions, as sibling and as young infant. What is the relation between the two? So when we add an overarching concept here, this is more technical, accepting that many languages do that, without a semantic criterion in mind. Any hierarchy that goes beyond this is welcome, and could be implemented in parallel to our system, I am sure, but it would then need to be thoroughly done for a larger range, probably with some idea of research in mind as well (e.g., compare the relations with wordnet, with clics, etc.). I assume you kept your other relations, right? so we can discuss this a bit later.

Schweikhard:
The main idea of different kinds of metonymy would be to distinguish between purely language-specific relations (BARK) and those that actually have some semantic underpinning (CHILD), so that then the latter can be used for actual studies - for which further sub-categories would then be needed, which need further discussing.

And yes, I have a backup. The hyponym changes were mainly a matter of search and replace, though, so would be easy to redo.

xrotwang:
I like the idea of making the actual Concepticon data more meaningful and
transparent using concept relations - i.e. avoid any ideas of making it
more complete. So any rationale for relations which avoids the "so
concepticon is an ontology, after all" trap is good.

PR #625, opened 2019-07-22
adding more concepts and refining old links

LinguList:
This PR is based on work by many people. It proposes new definitions that we'll add lists for later, as the 5 new lists need to be prepared. To keep the PR small, I introduce it already now. What also can be added later (and is in part in preparation) are the relations.


LinguList:
There's a certain danger of few errors in the data, but I will merge this now, so I can add five more lists. Afterwards, thorough checking can be done. For the Chinese lists, I am to a large degree convinced that they have improved greatly now.

LinguList:
Oh dammit, I wanted to make the adding of more lists a second PR, not this. I forgot to merge it. Sorry to everybody.

PR #627, opened 2019-07-24
adding three more lists

LinguList:


xrotwang:
Do we have sources for these?

xrotwang:
Btw. I'd find it more convenient if you would create PRs from branches **within** this repository, rather than from your fork. That way we could collaborate on the PRs, e.g. I just ran `concepticon stats` on this PR, but can't push the changes, because I don't have push access to your fork.

LinguList:
All sources are in references.bib.

LinguList:
As to the branching thing: I can do that, but I keep messing it up. If you have a good short-tutorial that illustrates this, I can use it in the future. But I suppose I'll add the stats now?

LinguList:
Or do you mean the PDF or similar by sources?

xrotwang:
Yes, PDFs was what I was thinking of.

LinguList:
They are like 100 pages for two of them, I think we rather leave them out, as they also deliver the words in the languages. In the past, I would at times cut the real data out of the PDF, but it would take quite long to do this...

LinguList:
Can we merge?

PR #630, opened 2019-07-29
updating data, adding Wang Feng's thesis

LinguList:


LinguList:
I made a mistake: these are Brian Allen's data, the data by Wang Feng is different. Will be added later.

Schweikhard:
Already had started looking into the first couple concepts, and noticed some potential issues with the mapping. I mainly went by the English glosses, so this would need to be confirmed by a Chinese speaker.
Is field meadow 旱地 FIELD or GRASSLAND?
The months are specified as lunar, not solar, should they really be linked to JANUARY etc?
hare,rabbit 兔 - shouldn't that be linked to LEPORID (RABBIT OR HARE)?
mouse 老鼠 is linked to MUROID (MOUSE OR RAT), is that correct?
feather,hair 毛, shouldn't that be FEATHER OR FUR OR HAIR and not only FEATHER?
So one might want to check whether the mapping of Allen is reliable.

LinguList:
This is a wrong list, the correct list is Allen-2007-500, I accidentally did the mapping twice, as I am working on data I converted 10 years ago. So this is all irrelevant, but thanks for looking at it anyway.

Schweikhard:
4 of the 5 potential mistakes are on the website as well, though - but if you did the same "mistakes" twice, they probably aren't mistakes.

LinguList:
> Is field meadow 旱地 FIELD or GRASSLAND?
correct mapping after my PR
> The months are specified as lunar, not solar, should they really be linked to JANUARY etc?
These mean "january", I assume people in Yunnan use the normal Chinese calendar.
> hare,rabbit 兔 - shouldn't that be linked to LEPORID (RABBIT OR HARE)?
isn't this the case in Allen?
> mouse 老鼠 is linked to MUROID (MOUSE OR RAT), is that correct?
yes, it's correct

Schweikhard:
Then hare,rabbit 兔 and mouse 老鼠 need to be corrected in Allen.

LinguList:
I'll add a fix and submit this with the corrected version of Wang 2004.

PR #632, opened 2019-07-31
Conceptrelations in concepticon.json

Schweikhard:
I added some definitions for conceptrelations (most of the existing ones and some new ones) to the concepticon.json.


Schweikhard:
Sorry bout that - I had set my editor to use spaces, the online json-checker seems to have changed that. Corrected.

LinguList:
Nice. It's no big deal with the spaces, but it makes reviewing easier. Thanks for reacting so quickly.

Issue #459, opened 2018-03-24
Serva-2010-100


LinguList:
The list is based on estimates for AN and IE languages (Dyen and ABVD, 200 concepts originally, although not clear how to link with Swadesh ?), and they report the most stable 100 items in their data.

* http://iopscience.iop.org/article/10.1088/1742-5468/2010/03/P03015/meta

Eprint on Arxiv is here:

* http://arxiv.org/abs/0912.0821

```

  	
@article{1742-5468-2010-03-P03015,
  author={Filippo Petroni and Maurizio Serva},
  title={Lexical evolution rates derived from automated stability measures},
  journal={Journal of Statistical Mechanics: Theory and Experiment},
  volume={2010},
  number={03},
  pages={P03015},
  url={http://stacks.iop.org/1742-5468/2010/i=03/a=P03015},
  year={2010},
  abstract={Phylogenetic trees can be reconstructed from the matrix which contains the distances between all pairs of languages in a family. Recently, we proposed a new method which uses normalized Levenshtein distances among words with the same meaning and averages over all the items of a given list. Decisions about the number of items in the input lists for language comparison have been debated since the beginning of glottochronology. The point is that words associated with some of the meanings have a rapid lexical evolution. Therefore, a large vocabulary comparison is only apparently more accurate than a smaller one, since many of the words do not carry any useful information. In principle, one should find the optimal length of the input lists, studying the stability of the different items. In this paper we tackle the problem with an automated methodology based only on our normalized Levenshtein distance. With this approach, the program of an automated reconstruction of language relationships is completed.}
}
```

LinguList:
![petroni-2010-100](https://user-images.githubusercontent.com/2916210/37948838-45af4036-3192-11e8-9f4b-04bcccf199dc.png)

Original list in the text on page 9.

PR #638, opened 2019-08-09
added Petroni list

Kristina-Pianykh:
Issue #459. I would just like to point out a few things:

1. Even though the given study uses data for languages of the Indo-European and Austronesian families, this list is said to represent the lexical items for the former [(p. 7)](https://iopscience.iop.org/article/10.1088/1742-5468/2010/03/P03015/pdf):

> The complete list of the 100 most stable terms for the Indo-European group can
be found in table 1.

2. The plurality of _you_ in this list is not specified. In contrast, Dyen's dataset, on which this list is based, contains _you_ in both singular and plural. I assumed that from the typological perspective on markedness this reportedly stable pronoun is more likely to be singular (thus linked to THOU).

3. Same goes for _right_ (either RIGHT (CORRECT) or RIGHT).


concepticondata/references/references.bib
```diff
@@ -2206,3 +2206,16 @@ @article{Hays1972
...
+@article{Petroni2010,
+	Author = {Petroni, Filippo and Serva, Maurizio},
+	Journal = {Journal of Statistical Mechanics: Theory and Experiment},
+	Title = {Lexical evolution rates derived from automated stability measures},
+	Year = {2010},
+	Publisher = {IOP Publishing},
+	Volume = {2010},
+	Number = {3},
+	Doi = {10.1088/1742-5468/2010/03/p03015},
+	Url = {https://doi.org/10.1088%2F1742-5468%2F2010%2F03%2Fp03015},
```
LinguList: delet the url, I suggest: it\s identical with the doi

LinguList:
Your json file is corrupted. Otherwise, the mappings should be fine. I follow your reasoning. You can check the test yourself on your computer:

```
concepticon test
```

or 

```
concepticon tests
```
Don't know which, but check `concepticon help` otherwise. The json syntax is quite strict, if you use a text editor with syntax highlight, it will also show you where the problem lies. @chrzyki can also indicate this quickly, I guess.

SimonGreenhill:
It's the trailing comma on line 36:

```json
                        "name": "GLOSS",
```

Remove that and it'll be fine :)

SimonGreenhill:
Since I was here I fixed the json error, but now it doesn't like "S(i)":

"ValueError: S(i) is not a valid name"

rename it to "Si"?

LinguList:
> rename it to "Si"?

Yes, @Kristina-Pianykh, the column names must be capitals in our conventions, and no brackets or similar are allowed, so you could write `S_I` instead. But then you also need to change that column name in the metadata json file. Please check the modification of the json file by @SimonGreenhill, so you remember the specifics of json syntax.

concepticondata/conceptlists/Petroni-2010-100.tsv
```diff
@@ -0,0 +1,101 @@
...
+Petroni-2010-100-37	37	Wind	960	WIND	0.19535
+Petroni-2010-100-38	38	Snake	730	SNAKE	0.19485
+Petroni-2010-100-39	39	Sing	1261	SING	0.19434
+Petroni-2010-100-40	40	Stone	857	STONE	0.19369
+Petroni-2010-100-41	41	Suck	1421	SUCK	0.19196
+Petroni-2010-100-42	42	Mouth	674	MOUTH	0.19067
+Petroni-2010-100-43	43	Dig	1418	DIG	0.19052
+Petroni-2010-100-44	44	Live	1422	BE ALIVE	0.18716
+Petroni-2010-100-45	45	Root	670	ROOT	0.18715
+Petroni-2010-100-46	46	Hair	2648	HAIR (HEAD)	0.18522
```
chrzyki: Is this specific, i.e. HEAD HAIR, or just HAIR?

LinguList: yes, if they claim, it's the swadesh-200, it shoudl be the one they have there.

Kristina-Pianykh: @SimonGreenhill @LinguList @chrzyki thank you for your help!

Issue #340, opened 2017-06-14
Lithgow-1976-50


lannin:
sealang.net/archives/pl/pdf/PL-C39.441.pdf list for Milne Bay languages (New Guinea).

LinguList:
thanks for pointing to this. Would you be able to help us with some further information?

* pages of the word list
* ideally the reference in bibtex format (we can do it ourself, but if you have it in your data, this would be great, you could just paste it here in)
* some note on the origin of this list and why you think it's useful (languages covered, is it an areal list, etc.)
* your name and affiliation, so we can thank you in our acknoledgments section

lannin:
The wordlist can be found at page 513. 
Glottolog gives this reference:
author     = {David R. Lithgow},
journal    = {Pacific Linguistics: Series C},
pages      = {441-523},
title      = {Austronesian languages: Milne Bay and adjacent islands (Milne Bay Province). In Austronesian languages, ed. by S. A.Wurm},
volume     = {39},
year       = {1976},
hhtype     = {comparative (computerized assignment from "languages")},
inlg       = {English [eng]},
macro_area = {Papua},
src        = {silpng}
I think it can be useful as an areal list of Milne Bay region of Papua New Guinea.
Best wishes, Evgeniya Korovina
PS If you interested in it is possible for me to provide more links to areal wordlists.

LinguList:
A it's you, Evgeniya, welcome here on github! Thanks a lot!

LinguList:
> PS If you interested in it is possible for me to provide more links to areal wordlists.

would be **definitely** interesting!

Kristina-Pianykh:
@LinguList I'd like to work on this list but I was wondering if it's still relevant.

LinguList:
It is.

Kristina-Pianykh:
@LinguList one more thing:

There are a few lists in the source, the first one being the vocabulary, the second - the free pronouns and the rest of them - some morphological elements (affixes). Shall I take the list with the pronouns together with the vocabulary (p. 516 in the [source](http://sealang.net/archives/pl/pdf/PL-C39.441.pdf)?

LinguList:
I'd make two lists then, I guess, if the book makes a strong distinction here.

lannin:
I think it's may be better as different lists...

вт, 1 окт. 2019 г., 14:42 Kristina-Pianykh <notifications@github.com>:

PR #650, opened 2019-10-02
Lithgow-1976-50 and Lithgow-1976-7

Kristina-Pianykh:
#340. Unfortunately, although the source states that the current list is based on Swadesh (see the description of the list) there is no explicit reference, indicating which of his lists was used in this case.

Another issue is the concept _ground_, absent in all of Swadesh's lists. I guess it might've been meant as EARTH (SOIL) but I'm not entirely sure.


LinguList:
That is not unusual, that people don't really quote what they use. Concepticon checks automatically for similarity on the website. Ground is probably the equivalent of earth, but I assume you linked it still to the literal concept, right?

LinguList:
Looks nice, many thanks!

PR #654, opened 2019-10-09
update

### LINKED ISSUE #649: Identical concept sets
There are a few instances of similar, if not identical, concept sets:

|ID|Name|Definition|Semantic field|Ontological category|Representation
|---|---|---|---|---|---|
|140|PADDY FIELD| A flooded field where rice is grown.| Agriculture and vegetation| Person/Thing|17|
|283|PADDY| Wet land in which rice is grown.|Agriculture and vegetation| Person/Thing|0|

_________________________________________________________________________________________

|ID|Name|Definition|Semantic field|Ontological category|Representation
|---|---|---|---|---|---|
|217|BETELNUT|The seed of the areca palm (Areca catechu).|Agriculture and vegetation| Person/Thing|15|
|2369|BETEL NUT|The seed of the areca palm (Mama'an in Meranau common version).| Food and drink| Person/Thing|0|

_________

|ID|Name|Definition|Semantic field|Ontological category|Representation
|---|---|---|---|---|---|
|732|YOUR (SINGULAR)| belonging to the second person that is addressed by the first person.|Possession | Other|10|
|2091|YOUR|belonging to the second person that is addressed by the first person|Possession| Other|0|

____________
Here, if one of the concepts is to be cut out, all the corresponding links should be moved accordingly.

|ID|Name|Definition|Semantic field|Ontological category|Representation
|---|---|---|---|---|---|
|241|TREE STUMP|A small remaining portion of the trunk of a tree with the roots still in the ground.| Agriculture and vegetation|Person/Thing |8|
|1023|STUMP OF TREE|The short piece left over after cutting off the most part.| Agriculture and vegetation|Person/Thing |3|

___________________
I'm not sure if these concept sets differ to the extent they should both be kept but then, in this case, this difference should be reflected in their definitions. On the other hand, there's just one entry _to live_, mapped to BE ALIVE, which makes LIVE irrelevant.

|ID|Name|Definition|Semantic field|Ontological category|Representation
|---|---|---|---|---|---|
|1422|BE ALIVE|To be alive; to have a life.|The body|Action/Process|94|
|2309|LIVE|To be alive; to have a life.|The body|Action/Process|0|

_____
The specification FRESHWATER in the concept's name is redundant.

|ID|Name|Definition|Semantic field|Ontological category|Representation
|---|---|---|---|---|---|
|467|CRAYFISH|A freshwater crustacean resembling small lobster.|Animals|Person/Thing|1|
|2813|CRAYFISH (FRESHWATER)| freshwater crustaceans resembling small lobsters| Animals|Person/Thing|1|

___________
Another issue relates to the distinction between BRAISE and STEAM which seems rather vague. Moreover, their definitions don't reflect the concepts' POSs.

|ID|Name|Definition|Semantic field|Ontological category|Representation
|---|---|---|---|---|---|
|2068|BRAISE (VERB)|A method of cooking using steam that avoids overcooking or burning food.| Food and drink|Action/Process|3|
|2067|STEAM (VERB)|A method of cooking using steam.|Food and drink|Action/Process|4|

I'd suggest the following definition of BRAISE (VERB): to cook meat or vegetables at a low temperature in a covered pan with a small amount of liquid.
STEAM (VERB) - to cook food, using steam.

Schweikhard:
fixed #649
And fixed part of stuff I noticed while annotating.


concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
```diff
@@ -581,8 +581,8 @@ BeijingDaxue-1964-905-579	579	挑	290	verb	body-part related actions	三、肢
 BeijingDaxue-1964-905-580	580	措	290	verb	body-part related actions	三、肢體動作	動詞	collect	cuò	7	GATHER	措
 BeijingDaxue-1964-905-581	581	跟	291	verb	body-part related actions	三、肢體動作	動詞	follow	gēn	1085	FOLLOW	跟
 BeijingDaxue-1964-905-582	582	逃跑	291	verb	body-part related actions	三、肢體動作	動詞	flee, run away	táo pǎo	2142	RUN AWAY	逃跑
-BeijingDaxue-1964-905-583	583	脫	292	verb	daily actions	四、日常操作	動詞	take off (clothes)	tuō	1755	UNDRESS	脱
-BeijingDaxue-1964-905-584	584	戴	292	verb	daily actions	四、日常操作	動詞	wear (on head)	dài	976	WEAR	戴
+BeijingDaxue-1964-905-583	583	脫	292	verb	daily actions	四、日常操作	動詞	take off (clothes)	tuō	3538	TAKE OFF	脱
```
LinguList: What is the difference between UNDRESS and TAKE OFF here? Do you think we need it to make more sense of these things? I would say "tuō" is pretty much the same as undress?

Schweikhard: The reasoning was that, from what I've been told, tuō needs a direct object, so it always refers to taking off particular pieces of clothing, whereas undress is a kind of intransitive taking off.

LinguList: Yes, the question, that I am asking myself here, however, is: does that really reflect the data? The point is: if we find the regular tuō yīfu in other datasets of SEA languages, which has the object and therefore intransifies the verb, as it is usual in Chinese glossing, what is it that they answer in the questionnaire? I would suspect, but this needs to be checked, that the questionnaires even if they explicitly ask for tuō yīfu, they will often only answer tuō or the equivalent in the respective languages under investigation. This should definitely be checked, by looking at cases of tuō in our data so far. Let me do that later.

LinguList: Screenshot_2019-10-10_12-38-12

LinguList: Please check the situation here. As far as I can tell, the majority of cases are those where the verb-object compound is used and at least alluded to. 

Here's another account of all verbs that are listed under "undress":

Screenshot_2019-10-10_12-38-12

You can see that the current linking is in fact very consistent, as the majority of sources are from SEA languages, where Chinese was a glossing language, and that they express "undress", specifically also because Chinese has a specific verb here, while English and German are a bit shakier. I suggest that we leave undress as is here, since the cases are all strikingly similar, and the usage of verb-object or verb-only in Chinese glossing is not consistently interpretable as you would do if you deal with a langauge like German, where verb-object compounds that express intransitive verbs do not exist. 

If we realize in later stages that we should distinguish those, we can do it, but for now it seems to be an overkill, also since the number of cases is not that much, and since I know that the Beijing source is just using a different glossing.

LinguList: However, looking at "take off" in our website, we can find some inconsistencies:

Screenshot_2019-10-10_12-38-12

Dellert should be linked to "undress", and the "take off one's clothes" gloss is obsolete, and should be merged with undress, so the entry in Lama should be linked to undress instead of the longer one we replaced long time ago.

The fact that "take off one's clothes" exists as a gloss is further important, as it shows that "take off" is not needed in any case...

Schweikhard: I see your point. Some seem to not mean fully undress, though, namely "take off (e.g. clothes)" (Bodt), and also "take off (EX:one's coat )" on Dellert. So I would link those two at "take off" and leave the others with "undress". Would you agree?

I did merge the obsolete gloss already.

Schweikhard: If that's the only issue, I would merge the pull request and fix this in the next commit afterwards.

LinguList: sorry for late reply: Bodt surely means take of clothes, so I'd leave it with undress, as I want to be able to compare across SEA languages here, and Bodt comes from the area. If Dellert has no real concept that means to undress completely, I'd also leave it there, as it is an overkill to make an extra concept for one concept list, if it is more or less compatible (and you do not know what they actually list there as translations in Northeuralex, or have you checked that?).

Schweikhard: Okay, I guess you convinced me. I now looked into Northeuralex, and while the words I understand all either mean "take off" or both, it's likely not worth it for a single list. My main motivation was the fact that we do distinguish between the opposites (putting on and getting dressed).

I'll remove "take off", adjust the definition of "undress" to include both meanings, and merge this pull request then.

LinguList: Ah, interesting, I did not know we have the two opposites. This actually points to something important with respect to the relations: To make sure we are getting better in specifying meanings, at least in the Swadesh range and a bit over it, we really need to get back to the relation issue. And this would be an example for tagging something as "opposite". I guess, this may in fact help to make some semantic fields clearer, although it is of course quite difficult.

Schweikhard: We can only mark them as proper opposites if we distinguish the meanings in both cases, though ;) I don't think we have a hyperonym for PUT ON and DRESS UP.

I intend to systematically add these kinds of concept relations at some point. And yes, they make things clearer, and also help in mapping by showing one what one may be overlooking.

LinguList: Okay, I just submitted a PR that essentially allows to have multiple relations, changing pyconcepticon code minimally. This means, that it should no longer fail, if one changes the order. It was much easier than I thought, to implement this.

PR #674, opened 2019-11-05
Huber

LinguList:
please see this check of huber, which now follows the source. There may still be typing errors, but the major mappings should work now, with the 368 concepts we find in the source.

PR #675, opened 2019-11-05
Small fixes to Huber-1992-375

tresoldi:
Continuing https://github.com/concepticon/concepticon-data/pull/674 , not much to fix.

The only question is: should we really add the duplicate `sangre/blood` concept? I can see that it is indeed listed in the Index and Appendix, both in Spanish and English, but it is clealy an error and the author/editor judged it so. In fact, the first `BLOOD` entry (concept number 33, page 71 in the PDF) is there, but the second is not (we jump from `carne/flesh`, concept number 265 and page 258, to `grasa/grease`, concept number 267 and page 259).

A minor thing, and it should be kept if we want the published *list* and not the published *data*. Other than that, looks really good!


tresoldi:
Wait! I forgot to revert the "wet", which is indeed an adjective (the Spanish gloss was wrong). Just one minute.

tresoldi:
Done for the mojado/wet

tresoldi:
ok

LinguList:
However, if you do so, @tresoldi, you need to renumber the Huber-1992-375-X, to be consistent, although I basically do not care, it is a unique identifier, but I kept renumbering. And you may have to go to conceptlists.tsv and change the number of actual concepts from 168 to 167.

LinguList:
Sorry, I didn't spot that...

LinguList:
ah, I even forgot to order the concepts by their ID, which is annoying, sorry also for that.

tresoldi:
@LinguList I just renumbered keeping the original order, do you prefer me to follow the order in the file?

tresoldi:
Ops, forgot that, thank you @xrotwang !

xrotwang:
@tresoldi what does the order of concepts in the list reflect now? Occurrence in the book?

xrotwang:
@tresoldi Ah, I see - alphabetic order of original gloss. I guess that makes sense, @LinguList ?

tresoldi:
@xrotwang yes, it is the order @LinguList had (just skipping the repeated and unused concept for "blood"). Maybe listing following the line order would be more homogeneous with other concept lists, but I am happy with that (and the new release should not depend on something so trivial).

LinguList:
sorry, had to go home, now back online.

LinguList:
I suppose: use the order of the ID field, that is: reorder, as this is numerical, following the page occurrences, as we traditionally do in concepticon. So the integers in the last position of id, after `-` should just increase.

tresoldi:
I'm on it.

LinguList:
The alphabetic order was just what I used to manually (yes) link the concepts in the huber-cysouw data with the concept list. This was just to get this done, as I did not want to write a script, so I sorted these things in libre-office.

LinguList:
Note one thing: the retired.json may not work in all cases, as I could not (and won't) manually check it, so given that I used the order of the book, which we used also before, I suppose this works in 99% of all cases, but there may be retired ids that link to the wrong concept.

LinguList:
One thing, @tresoldi, with 367 concepts in conceptlists.tsv, it is 8 entries we can't find in the book ;), not 7, as it states now. But we can put that into an issue as well. I am tired today...

tresoldi:
Fixing that as well

LinguList:
Okay, now we have an unordered column in numbers, as the numbers don't follow the ids, but I am fine with that, thanks @tresoldi.

tresoldi:
Ah, I see, you want to follow the order of the IDs, of course it is better. Just a sec (hope I am not holding you guys)

LinguList:
I usually do the ids after fixing the rest with vim, as it offers nice functions for that. This time, I messed it up, as I was eager to get it over with...

xrotwang:
ok, seems good to merge, right?

LinguList:
We have to make an issue with respect to retired.json. With the ids changed in numeration, going async with the number column, this means, we are definitely wrong. So what has to be done, is to take the old version and manually check that they are in sync. But I'd say: I make an issue now, flag it as 2.3, and we don't care for this right now, okay?

LinguList:
voila, see #676, can't believe we have already that many issues.

xrotwang:
hm, ok, I see. I think I'll fix it now.

LinguList:
Okay, then let's merge first, right?

tresoldi:
Up to you, I should finish soon the (hope right) new IDs soon (taking care of dinner at the same time :wink:)

chrzyki:
And, for testing purposes, here the dataset with the fixed list. Will redo once this has been merged:

https://github.com/lexibank/hubercolumbian/commit/28938e2948ee0d91ce9c10291c34327558245580

tresoldi:
Just pushed a new numbering, is that the right one?

tresoldi:
@chrzyki don't worry, I didn't touch the gloss in digital source :wink:

chrzyki:
Still different (and better now!) than before. We previously split of the Spanish part and compared only against the English part. Cleaner now this way I think, with the full SPANISH_ENGLISH glosses.

xrotwang:
@tresoldi Is the idea that the new renumbering keeps old IDs as much as possible? If so, that doesn't seem to be the case, e.g. [abajo](https://concepticon.clld.org/values/Huber-1992-374-369) used to be `-369` and is 361 now. I guess because we want to have sequential IDs?
Anyway, if that's the intended renumbering, we still have to fix retired.json.

LinguList:
Yes, that was my idea from the beginning: add a reference to the digital gloss, or how we call it, so the things remain completely stable. It will also get rid of missing concepts in the digitization, which is not our fault. So we say: digital source says spanish-english, but concepticon officially gives spanish and englihs. And we can easily use those for the future, and again, lexibank-cldf adds value to the data, as it offers a readable gloss for spanish and english. 

If there are mismatches, @chrzyki, as I may have introduced errors in the spanish-english construct, without wanting so, you may have to refine with a dictionary manually, but I assume there are maximally 1-5 entries that may not work now. So this can be done manually and left as is or potentially refined.

chrzyki:
@LinguList no mismatches as far as I can tell! Ran fine without any manual fixes.

tresoldi:
Anything I can do? Also considering https://github.com/concepticon/concepticon-data/issues/676

xrotwang:
ok, I'll let this sit over night, then come back tomorrow morning to merge and release :)

PR #682, opened 2019-12-05
Added Tolmie-1884-211

chrzyki:
This adds 211 items from the "Comparative Vocabularies of the Indian Tribes of British Columbia". It only consists of what the authors consider their main vocabulary (the Dawson-Tolmie Vocabulary).


chrzyki:
Sorry for including this, meant to upload this rather.

On 5 Dec 2019 13:02, Robert Forkel <notifications@github.com> wrote:

@xrotwang commented on this pull request.

Please don't merge yet. We don't want >80 MB PDFs in the repos permanently (I'm not sure whether they will be deleted with the branch, though).

xrotwang:
@chrzyki can you upload it? Would take quite a while over my bad line.

chrzyki:
Yep, will do!

On 5 Dec 2019 13:09, Robert Forkel <notifications@github.com> wrote:

@chrzyki<https://github.com/chrzyki> can you upload it? Would take quite a while over my bad line.

concepticondata/conceptlists.tsv
```diff
@@ -275,3 +275,4 @@ Othaniel-2017-300	Othaniel, Nlabephee Kefas	2017		300	basic	English	Jen language
...
 Joo-2019-100	Joo, Ian	2019		100	ranked, basic	English	Global		Joo2019		This list of 100 items is based on the [Leipzig-Jakarta concept list](:ref:Tadmor-2009-100) and was employed in the study of phonosemantic patterns in the basic vocabulary of 66 genealogically unrelated languages.	1-12	
+Tolmie-1884-211	Tolmie, William F. and Dawson, George M.	1884		211	basic	English	Languages of the Native Tribes of British Columbia	https://open.library.ubc.ca/collections/bcbooks/items/1.0222394	Tolmie1884	Tolmie1884	This collection of 211 items, based on [Instructions for Research Relative to the Ethnology and Philology of America](:bib:Gibbs1863), was collected by the authors during expeditions in the British Columbia territory during the winters of 1875 and 1876. It contains basic vocabulary terms in addition to local flora and fauna terms. The core list of 211 items, also known as Dawson-Tolmie Vocabularies, is extended by appendices with extended vocabularies for specific languages in the area, but is not considered part of the main vocabulary list.
```
LinguList: If they are also know as Dawson-Tolmie Vocabs, you could add this as an alias for the list.

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-3	3	Boy	1366	BOY
+Tolmie-1884-211-4	4	Girl	1646	GIRL
+Tolmie-1884-211-5	5	Infant	1289	BABY
+Tolmie-1884-211-6	6	My father (said by son)		
+Tolmie-1884-211-7	7	My father (said by daughter)		
+Tolmie-1884-211-8	8	My mother (said by son)		
+Tolmie-1884-211-9	9	My mother (said by daughter)		
+Tolmie-1884-211-10	10	My husband		
+Tolmie-1884-211-11	11	My wife		
+Tolmie-1884-211-12	12	My son (said by father)
```
LinguList: kinship terms that depend on the EGO are in fact in part available in concepticon. I'd also say that the "my" construct here can be savely ignored. It is a phrase they use for elicitation, as this triggers the word, similar to using the imperative in verbal elicitation, and we tend to link those, unless there's an elicitation of the word for "father" or similar in isolation. So I'd link husband, wife, elder brother, elder sister, etc., but maybe leave the father-mother block unlinked, or add this as a distinction (there's "papa" vs, "father", and then there's the perspective of male egos, etc., maybe something we'll also find useful for kinbank and other projects?)

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-49	49	Town, village		
+Tolmie-1884-211-50	50	Chief	2394	CHIEF
+Tolmie-1884-211-51	51	Warrior		
+Tolmie-1884-211-52	52	Friend	1325	FRIEND
+Tolmie-1884-211-53	53	House	1252	HOUSE
+Tolmie-1884-211-54	54	Skin lodge		
+Tolmie-1884-211-55	55	Kettle	854	KETTLE
+Tolmie-1884-211-56	56	Bow	994	BOW
+Tolmie-1884-211-57	57	Arrow	977	ARROW
+Tolmie-1884-211-58	58	Axe, hatchet	2083	AXE (CLASSIFIER)
```
LinguList: careful, classifier is obviously wrongly linked here.

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-70	70	Morning	1339	MORNING
+Tolmie-1884-211-71	71	Evening	1629	EVENING
+Tolmie-1884-211-72	72	Spring	1537	SPRINGTIME
+Tolmie-1884-211-73	73	Summer	1342	SUMMER
+Tolmie-1884-211-74	74	Autumn	1276	AUTUMN
+Tolmie-1884-211-75	75	Winter	1471	WINTER
+Tolmie-1884-211-76	76	Wind	960	WIND
+Tolmie-1884-211-77	77	Thunder	1150	THUNDER
+Tolmie-1884-211-78	78	Lightning	1654	LIGHTNING
+Tolmie-1884-211-79	79	Rain	658	RAIN (PRECIPITATION)
```
LinguList: Using the noun here is justified by the context, an interesting example of how linking depends on structure of a concept list.

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-80	80	Snow	784	SNOW
+Tolmie-1884-211-81	81	Fire	221	FIRE
+Tolmie-1884-211-82	82	Water	948	WATER
+Tolmie-1884-211-83	83	Ice	617	ICE
+Tolmie-1884-211-84	84	Earth, land
```
LinguList: We should have a concept that covers the term earth/land. If not, we should add it, as this is a very typical colexification of "earth" in english: "earth = land" "earth = soil" "earth = the earth"

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-98	98	Leaf	628	LEAF
+Tolmie-1884-211-99	99	Bark	1204	BARK
+Tolmie-1884-211-100	100	Grass	606	GRASS
+Tolmie-1884-211-101	101	Pine	1740	PINE
+Tolmie-1884-211-102	102	Maize	506	MAIZE
+Tolmie-1884-211-103	103	Squash	864	PUMPKIN
+Tolmie-1884-211-104	104	Flesh, meat	2615	FLESH OR MEAT
+Tolmie-1884-211-105	105	Dog	2009	DOG
+Tolmie-1884-211-106	106	Buffalo	934	BUFFALO
+Tolmie-1884-211-107	107	Bear (black)
```
LinguList: you did not link the bear because it is a black bear? I'd say: link to BEAR for now.

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-116	116	Fly	1504	FLY (INSECT)
+Tolmie-1884-211-117	117	Mosquito	1509	MOSQUITO
+Tolmie-1884-211-118	118	Snake	730	SNAKE
+Tolmie-1884-211-119	119	Rattlesnake	871	RATTLESNAKE
+Tolmie-1884-211-120	120	Bird	937	BIRD
+Tolmie-1884-211-121	121	Egg	744	EGG
+Tolmie-1884-211-122	122	Feathers	1201	FEATHER
+Tolmie-1884-211-123	123	Wings	1257	WING
+Tolmie-1884-211-124	124	Goose	1187	GOOSE
+Tolmie-1884-211-125	125	Duck (mallard)	1360	DUCK
```
LinguList: Otherwise, if you don't link the bear, you should not link the mallard duck to duck (whatever that is...)

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-126	126	Turkey	2386	TURKEY
+Tolmie-1884-211-127	127	Pigeon	1853	DOVE
+Tolmie-1884-211-128	128	Fish	227	FISH
+Tolmie-1884-211-129	129	Salmon	944	SALMON
+Tolmie-1884-211-130	130	Sturgeon
```
LinguList: This will require some digging, but I'd be inclined to add the Sturgeon, as it is an example of flora and fauna. It should be the typical representative in this area, of course, as there are too many species of the Stör.

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-126	126	Turkey	2386	TURKEY
+Tolmie-1884-211-127	127	Pigeon	1853	DOVE
+Tolmie-1884-211-128	128	Fish	227	FISH
+Tolmie-1884-211-129	129	Salmon	944	SALMON
+Tolmie-1884-211-130	130	Sturgeon		
+Tolmie-1884-211-131	131	Name	1405	NAME
+Tolmie-1884-211-132	132	White	1335	WHITE
+Tolmie-1884-211-133	133	Black	163	BLACK
+Tolmie-1884-211-134	134	Red	156	RED
+Tolmie-1884-211-135	135	Light blue
```
LinguList: We are very bad in colors, but we should enhance. This is worth adding light blue (Russian goluboj, approximately) to the list.

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-128	128	Fish	227	FISH
+Tolmie-1884-211-129	129	Salmon	944	SALMON
+Tolmie-1884-211-130	130	Sturgeon		
+Tolmie-1884-211-131	131	Name	1405	NAME
+Tolmie-1884-211-132	132	White	1335	WHITE
+Tolmie-1884-211-133	133	Black	163	BLACK
+Tolmie-1884-211-134	134	Red	156	RED
+Tolmie-1884-211-135	135	Light blue		
+Tolmie-1884-211-136	136	Yellow	1424	YELLOW
+Tolmie-1884-211-137	137	Light green
```
LinguList: Same case. I'd go as far as checking to which degree svg color names reflect these variants.

concepticondata/conceptlists/Tolmie-1884-211.tsv
Kristina-Pianykh: There are instances of earth/land entries in other lists, mapped to LAND.

xrotwang: Doesn't IDS have this? https://wold.clld.org/meaning/1-21 ?

concepticondata/conceptlists/Tolmie-1884-211.tsv
SimonGreenhill: https://en.wikipedia.org/wiki/Mallard :)

concepticondata/conceptlists/Tolmie-1884-211.tsv
```diff
@@ -0,0 +1,212 @@
...
+Tolmie-1884-211-80	80	Snow	784	SNOW
+Tolmie-1884-211-81	81	Fire	221	FIRE
+Tolmie-1884-211-82	82	Water	948	WATER
+Tolmie-1884-211-83	83	Ice	617	ICE
+Tolmie-1884-211-84	84	Earth, land		
+Tolmie-1884-211-85	85	Sea	1474	SEA
+Tolmie-1884-211-86	86	River	666	RIVER
+Tolmie-1884-211-87	87	Lake	624	LAKE
+Tolmie-1884-211-88	88	Valley	921	VALLEY
+Tolmie-1884-211-89	89	Prairie
```
Kristina-Pianykh: map to GRASSLAND? If it's not too much of generalization as prairie is related to grasslands of North America, as far as I understood.

LinguList: Yes, I think we did that in other cases as well.

chrzyki:
Thanks for the pointers, will change accordingly.

chrzyki:
Updated, please check.

concepticondata/conceptrelations.tsv
```diff
@@ -1217,3 +1217,19 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
 3074	FIX	instanceof	2243	REPAIR
 1755	UNDRESS	instanceof	3538	TAKE OFF
 158	PUT ON	antonymof	3538	TAKE OFF
+3603	EARTH OR LAND	instanceof	1228	LAND
+3603	EARTH OR LAND	instanceof	626	LAND
```
LinguList: this is not possible: I assume 626 and 1228 have different concepticon glosses, right?

LinguList: Or you should write "narrower"

concepticondata/conceptrelations.tsv
```diff
@@ -1217,3 +1217,19 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
...
+3604	LIGHT BLUE	instanceof	837	BLUE
+3604	LIGHT BLUE	narrower	837	BLUE
+3605	LIGHT GREEN	instanceof	1425	GREEN
+3605	LIGHT GREEN	narrower	1425	GREEN
+3606	STURGEON	instanceof	227	FISH
+3606	STURGEON	narrower	227	FISH
```
LinguList: for a kind of fish in comparison to a fish, I suggest not to use the narrower relation. So instanceof is enough.

concepticondata/conceptrelations.tsv
LinguList: I suspect even if "EARTH OR LAND" needs to be linked here: it seems that they rather mean the land as in concepticon LAND. 
And of course, the concept needs to link to SOIl then as well (narrower relation).

concepticondata/conceptrelations.tsv
```diff
@@ -1217,3 +1217,19 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
 3074	FIX	instanceof	2243	REPAIR
 1755	UNDRESS	instanceof	3538	TAKE OFF
 158	PUT ON	antonymof	3538	TAKE OFF
+3603	EARTH OR LAND	instanceof	1228	LAND
+3603	EARTH OR LAND	instanceof	626	LAND
+3604	LIGHT BLUE	instanceof	837	BLUE
+3604	LIGHT BLUE	narrower	837	BLUE
+3605	LIGHT GREEN	instanceof	1425	GREEN
+3605	LIGHT GREEN	narrower	1425	GREEN
```
LinguList: I would rather say: these are distinct colors, so we should not subsume them, since there are external classifications that do a better job here (otherwise, we'd also have to subsume yellow and orange under BROWN, etc.).

LinguList: So just delete the relations.

concepticondata/conceptrelations.tsv
```diff
@@ -1217,3 +1217,19 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
...
+3606	STURGEON	instanceof	227	FISH
+3606	STURGEON	narrower	227	FISH
+3607	MOCCASIN	instanceof	1381	SHOE
+3607	MOCCASIN	narrower	1381	SHOE
+3609	TOWN OR VILLAGE	instanceof	1391	TOWN
+3609	TOWN OR VILLAGE	instanceof	930	VILLAGE
+3610	MALLARD	instanceof	1360	DUCK
+3610	MALLARD	narrower	1360	DUCK
```
LinguList: Same as above: if there's an instanceof, there's ussually no narrower relation.

concepticondata/conceptrelations.tsv
```diff
@@ -1217,3 +1217,19 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
...
+3603	EARTH OR LAND	instanceof	626	LAND
+3604	LIGHT BLUE	instanceof	837	BLUE
+3604	LIGHT BLUE	narrower	837	BLUE
+3605	LIGHT GREEN	instanceof	1425	GREEN
+3605	LIGHT GREEN	narrower	1425	GREEN
+3606	STURGEON	instanceof	227	FISH
+3606	STURGEON	narrower	227	FISH
+3607	MOCCASIN	instanceof	1381	SHOE
+3607	MOCCASIN	narrower	1381	SHOE
+3609	TOWN OR VILLAGE	instanceof	1391	TOWN
```
LinguList: this is obviously a narrower relation, no instanceof relation. Note that this passes the tests, because the tests do not test instanceof.

concepticondata/conceptrelations.tsv
```diff
@@ -1217,3 +1217,19 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
...
+3609	TOWN OR VILLAGE	instanceof	930	VILLAGE
+3610	MALLARD	instanceof	1360	DUCK
+3610	MALLARD	narrower	1360	DUCK
+3611	AMERICAN BLACK BEAR	instanceof	1820	BEAR
+3611	AMERICAN BLACK BEAR	narrower	1820	BEAR
```
LinguList: same as before, just delete this relation.

concepticondata/conceptrelations.tsv
chrzyki: > this is not possible: I assume 626 and 1228 have different concepticon glosses, right?

My mistake, will fix.

chrzyki:
> Only relations to be modified this time, in my opinion, and maybe think about the earth land thing, but well, I would also not oppose to leave it as is (having this meta-concept is justified by the number of colexifications we find).

Thanks, updated.

Schweikhard:
Those are some nice new concepts!

But what's the reason for not adding an instanceof relation for the hyponyms of TOWN OR VILLAGE and EARTH OR LAND?

Schweikhard:
Oh, and for the other new relations, the columns need to be swapped (e.g. FISH instanceof STURGEON and FISH narrower STURGEON).

concepticondata/conceptrelations.tsv
Schweikhard: Doesn't narrower include both instanceof and partof? And it is an instanceof (i.e. hyponymic) relation by the definition we have (i.e. "Subcategory of x", the opposite of "hyperonymof").

LinguList:
> But what's the reason for not adding an instanceof relation for the hyponyms of TOWN OR VILLAGE and EARTH OR LAND?

@Schweikhard, instanceof doesn't make sense here. An instance of "Schokokreme" is "Nutella", but you would not say that "hand" is an instance of the Russian concept "arm or hand". Instance of is a rather clear-cut relation that should relate to a "kind of fish" when you have fish as the species.

LinguList:
And note the importance of never using narrower if there are instanceof relations. The idea is much different here.

Schweikhard:
But a VILLAGE is an instance of the set of settlements included in TOWN OR VILLAGE, isn't it?

And thus far the majority of what is marked as instanceof is also marked as narrower.

Schweikhard:
The hand to рука́ relation would be partof.

LinguList:
> But a VILLAGE is an instance of the set of settlements included in TOWN OR VILLAGE, isn't it?

> And thus far the majority of what is marked as instanceof is also marked as narrower.

Yes, but this is wrong.

LinguList:
> But a VILLAGE is an instance of the set of settlements included in TOWN OR VILLAGE, isn't it?

> And thus far the majority of what is marked as instanceof is also marked as narrower.

Yes, but this is wrong.

LinguList:
The instanceof relation is meant for cases where you have Nutella and chocolate cream. It was never meant for cases where you have a broad underspecified concept and more specified subversions of it.

Schweikhard:
>The instanceof relation is meant for cases where you have Nutella and chocolate cream. It was never meant for cases where you have a broad underspecified concept and more specified subversions of it.

But that is the same thing logically.
Whether something is underspecified or not is culture-dependent, so we can't really make a difference between these two cases. Chocolate cream is also broad and underspecified.

LinguList:
Narrower/broader are meant to capture problems resulting from underspecification in concept lists. Thus, if somebody says "child", and we don't know what they mean, we leave it unspecified, and we can still COMPARE "child (young human)" with "child (descandant)". But "child (young human)" is not an instance of a child. An instance of a child is Robert Downey Junior as a child of Robert Downey. Or "Osnabrück" is an instance of "town or village", and "Berlin" also, and Berlin is an instance of a town and Osnabrück an instance of a village (in my opinion).

LinguList:
> But that is the same thing logically.
> Whether something is underspecified or not is culture-dependent, so we can't really make a difference between these two cases. Chocolate cream is also broad and underspecified.

No. End of discussion here. A fish is a species. A trout is a kind of a fish. "fish or whale" is not a species. I do not expect that there will be any surprises here. If there are borderline cases, they can be discussed in issues. But here, the cases are clear.

Schweikhard:
I see the point in having an additional type of relation for these arbitrary cases where we are not sure which of several meanings is meant.
Okay, let's proceed as you suggest. It just seems likely to me that this will lead to many borderline cases.

Perhaps a good working definition might be then: use instanceof only for natural cases of hyponymy, and narrower for any kind of hyponymy that seems cultural or chance of language history.
By this, instanceof, narrower, and partof shouldn't have any overlap.

I can go through the conceptrelations.tsv again to update it accordingly.

LinguList:
Look, @Schweikhard, when I first added instanceof relations, I had the instance of a class in mind, such as in object-oriented programming. Think of Plato, where you have an idea, and then you have a concrete instance of that idea somewhere. You can argue that this never stops, but you can easily see that most languages have different words for kinds than they have for species. They have a word for fish as a kind, and then they have very particular words for the species, like a trout, that often follow their own morphosyntax. This is what instance should be reserved to, at least when I added the definition. 

There are cases when languages use an original name for a species to refer to the kind itself. "pajaro" meant "sparrow", but now means "bird". But even if this happens, there's a concept for "bird" as a thing that flies, and there are instances of birds, such as sparrow, crow, etc. So this is a very concrete idea about "instanceof" here. I am curious to see which borderline cases can be found, as I haven't really seen any so far in the concepticon.

Schweikhard:
With those definitions it should work reasonably well. It may only refer to organisms then, leaving narrower as the catch-all category for the remainder, but for that semantic field there should really be no issues. I just wonder how it should be applied to other semantic fields, because there the issues would start, but I guess it simply shan't.

In some languages the word for fish however also refers to e.g. whales, or the word for bird also refers to bats or insects, so languages don't necessarily follow these natural categories. Having them nevertheless as relations in the system could then be used as a standard against which to test how close languages follow these natural categories, so it seems like a good idea to use that distinction.

LinguList:
Yes, @Schweikhard, but even if the language has "fish or whale", they distinguish it from "dolphin" and "trout". And I bet there is a statistical effect that shows that on average there is a different syntax in naming. Anyway, I think we are on the same page. Pragmatism is important here, and other relations can be used to reflect what instanceof is not meant to reflect.

chrzyki:
> Oh, and for the other new relations, the columns need to be swapped (e.g. FISH instanceof STURGEON and FISH narrower STURGEON).

Ah, mixed that up. Sorry, been a while since I've done the conceptrelations. Sorry for the mess. :)

chrzyki:
I've updated the directionality of the relations, @Schweikhard, could you please check? Thank you!

Schweikhard:
The following ones need to be swapped back now:
1228	EARTH (SOIL)	narrower	3603	EARTH OR LAND
626	LAND	narrower	3603	EARTH OR LAND
1391	TOWN	narrower	3609	TOWN OR VILLAGE
930	VILLAGE	narrower	3609	TOWN OR VILLAGE
The rest is correct.

Except, the row
1381	SHOE	instanceof	3607	MOCCASIN
should probably be deleted as a moccasin is not a species of shoe.

chrzyki:
Gotcha, thanks, done.

chrzyki:
I hope that I've fixed all issues. Whenever someone has time to give a final okay that'd be much appreciated. :)

PR #692, opened 2019-12-12
Added Snider-2004-1700

Kristina-Pianykh:
Long pending Snider's list. I introduced new concept sets only for those concepts which have unmapped correspondences in other lists.

I thought it would be smarter to open a separate PR for mappings in other lists in line with the newly added concept sets.


LinguList:
Super. I won't do this today, but maybe Monday, and I added more reviewers. For longer lists, it is better to have more than one person watching.

I ask reviewers to avoid inline comments, unless they are quick to fix, as this is often confusing. Instead, maybe, a check list of problems can be mentioned in the summary? Or you do inline comments AND make a checklist?

Schweikhard:
Some great new concepts!

SLEEPY: We already have TIRED (which however includes also weary/exhausted), which I guess thereby is a hyperonym to this?

ANTELOPE: even-toad -> even-toed, ad -> and

I didn't check the actual mapping; probably that's rather for someone speaking French.

concepticondata/concepticon.tsv
```diff
@@ -3596,3 +3596,97 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3619	STAMP	Motion	To put your foot down on the ground heavily and noisily; to walk in this way.	Action/Process	
+3620	LEAN	Basic actions and technology	To bend or move from a vertical position or to make something do that.	Action/Process	
+3621	SLEEPY	The body	Feeling tired and willing to go to sleep.	Property	
+3622	DISABLED PERSON	The body	A person who has an illness or an injury that restricts the way they live their life, especially by making it difficult for them to move.	Person/Thing	
+3623	BURN (INJURY)	The body	An injury, caused by exposure to heat, fire or chemical agents.	Person/Thing	
+3624	INTESTINAL WORM	Animals	The organism, also known as the parasitic worm, that infects and breeds in the gastrointestinal tract of humans and animals.	Person/Thing	
+3625	LEPROSY	The body	The infectious disease, caused by the bacteria Mycobacterium leprae, that affects the skin, as well as the peripheral nerves.	Person/Thing	
+3626	HEADACHE	The body	A pain in the head.	Person/Thing	
+3627	KNOW	Cognition	To have met and recognize somebody or to be certain or sure about something.	Action/Process
```
chrzyki: Hm, we have `1410 KNOW (SOMETHING) To be certain or sure about something.` Would this be more general? What is the difference?

LinguList: kennen + wissen, know facts and know a person -> know facts or a person

chrzyki:
Ah, missed @LinguList's request, will collect comments here:

New concepts:

- [ ] 3627 KNOW see comment above.
- [ ] 3622 DISABLED PERSON This could/should be expanded to also include mental disability?
- [ ] 3625 LEPROSY "The infectious disease, caused by the bacteria Mycobacterium leprae, that affects the skin, as well as the peripheral nerves." -> "An infectious disease caused by the bacteria ycobacterium leprae that affects the skin as well as the peripheral nerves." (change article, take care of commas)
- [ ] 3624 INTESTINAL WORM "The organism, also known as the parasitic worm, that infects and breeds in the gastrointestinal tract of humans and animals." -> "An organism, also known as parasitic worm, that infects ..."
- [ ] 3650 PUNISH This could potentially be broader, i.e. punishing children is not punishing them for breaking the law.

LinguList:
So my major points are shared above. I think, with respect to the rest, this is okay, but I cannot guarantee that I have not maybe overseen something.

LinguList:
Ah, concept relations are, of course, needed for KNOW, if we accept the new suggestion.

concepticondata/concepticon.tsv
Kristina-Pianykh: Yes, the idea behind the new concept is to create a hypernym of KNOW (SOMETHING) and KNOW (SOMEBODY) because some languages don't have separate lexemes for the two (e.g., in Russian) and you can find something like _know something or somebody_ as an entry.

Kristina-Pianykh:
Concerning MARROW, I wasn't sure if it relates to the tissue and not to the vegetable. The concepts linked to MARROW don't seem to disambiguate that.

LinguList:
The MARROW concepts all point to the body part, I guess, specifically, since one of the lists is also a body part list, and the Chinese translations point to this. So I would say: assume that none of the lists means the plant.

Kristina-Pianykh:
In regard to TICKLE, here it's meant to be transitive (also considering the corresponding French gloss). E.g.: He tickled her feet.

But NOT in the sense _My throat tickles._

LinguList:
> I thought it was clear from the definition and the concept TICKLE (SOMEBODY OR SOMETHING) would be too clumsy.
Explicit is better than implicit. Who knows: we may have another concept tickle (intransitive) later, etc. So TICKLE (SOMEBODY) should be okay, as you usually do not tickle things.

Kristina-Pianykh:
> Some great new concepts!
> 
> SLEEPY: We already have TIRED (which however includes also weary/exhausted), which I guess thereby is a hyperonym to this?
> 
> ANTELOPE: even-toad -> even-toed, ad -> and
> 
> I didn't check the actual mapping; probably that's rather for someone speaking French.

Judging by the definition of TIRED, it's rather a hyponym of SLEEPY since it implies some physical activity as a possible reason of becoming tired. Besides, TIRED is a necessary prerequisite of SLEEPY.

Schweikhard:
> Judging by the definition of TIRED, it's rather a hyponym of SLEEPY since it implies some physical activity as a possible reason of becoming tired. Besides, TIRED is a necessary prerequisite of SLEEPY.

No, the definition of TIRED also includes the possibility of being tired due to lack of sleep, so it's an either/or. SLEEPY could be argued to refer to being tired due to lack of sleep but not due to physical exhaustion, i.e. it's a specific kind of tired, that's why I consider it a hyponym.

LinguList:
If the checks pass, feel free to merge. Many thanks!

PR #698, opened 2019-12-19
concept relations

Schweikhard:
I removed the duplicate concept relations (i.e. cases where the same concept pair has both a narrower and an instanceof relation) and fixed some obvious mistakes.

The next step would be to go through the list and decide which relations are rather narrower and which ones are rather instanceof. I did some preliminary work on that while removing the duplicates, but there are many cases where I am not certain which would be the better option.


Schweikhard:
Had to change the concept relations between SKIN and SKIN (HUMAN) and SKIN (ANIMAL), respectively, back to narrower for the Swadesh test.

PR #707, opened 2020-01-09
Added Xiao-2012-213 Norm data on Chinese characters

AnnikaTjuka:
About the data:
The data includes norms on stroke counts, picturability, iconicity and frequency of Chinese characters.

About the mapping:
I already double checked some of the mappings with Mei-Shin.
- _zhi_ was mapped to HE OR SHE OR IT
- _fang_ was left unmapped because it can mean ANGULAR (SQUAREFORMED) and SQUARE
- _mo_ can be understood as END OF TIME / SPACE therefore, I left it unmapped
- _chuan_ the English translation "string" seems to be misleading because it is more of a stick or "Shashlik Spies"


concepticondata/conceptlists/Xiao-2012-213.tsv
```diff
@@ -0,0 +1,214 @@
...
+Xiao-2012-213-3	3	刀	dao1	knife	KNIFE	1352	2	0.2	6.74	32	531
+Xiao-2012-213-4	4	子	zi3	son	SON	1620	2	0.2	4.05	9	614.3
+Xiao-2012-213-5	5	卜	bu3	fortune telling			2	0.35	4.16	106.5	25.7
+Xiao-2012-213-6	6	了	liao3	to finish	FINISH	1766	2	0.4	2.39	15.5	8187.2
+Xiao-2012-213-7	7	力	li4	power	POWER	2526	2	0.4	2.39	89.5	2281.3
+Xiao-2012-213-8	8	厂	chang3	factory			2	0.45	5.39	106.5	631
+Xiao-2012-213-9	9	乃	nai3	therefore			2	0.45	1.39	212.5	25.2
+Xiao-2012-213-10	10	入	ru4	to enter	ENTER	749	2	0.5	2.74	61	1164.6
+Xiao-2012-213-11	11	又	you4	in addition			2	0.5	2.06	37.5	1192.9
+Xiao-2012-213-12	12	几	ji3	several			2	0.55	2.42	93	840.4
```
LinguList: "several" but also "some", so "some" can also be mapped here. But leave unmapped in case of doubt.

concepticondata/conceptlists/Xiao-2012-213.tsv
```diff
@@ -0,0 +1,214 @@
...
+Xiao-2012-213-29	29	大	da4	big	BIG	1202	3	0.4	4.2	15.5	6857.3
+Xiao-2012-213-30	30	己	ji3	oneself	SELF	1993	3	0.45	1.97	133	997.2
+Xiao-2012-213-31	31	寸	cun4	inch			3	0.5	4.52	75.5	268.2
+Xiao-2012-213-32	32	卫	wei4	to protect	PROTECT	2900	3	0.5	2.61	72	306.2
+Xiao-2012-213-33	33	习	xi2	be used to			3	0.5	1.35	83.5	1127.4
+Xiao-2012-213-34	34	才	cai2	ability	ABILITY	2527	3	0.5	1.58	140.5	613
+Xiao-2012-213-35	35	儿	er2	son	SON	1620	3	0.55	4.05	75.5	2499.1
+Xiao-2012-213-36	36	亡	wang2	dead	DEAD	1625	3	0.55	4.71	204	114.6
+Xiao-2012-213-37	37	么	mo	a word indicating a question			3	0.55	2.3	93	893.7
+Xiao-2012-213-38	38	万	wan4	million			3	0.55	3.23	122	648.6
```
LinguList: don't we have one million in concepticon? If not, I suggest to add it, also for the sake of the numerals projects. @chrzyki, can you check, is "one million" important? and can you advise how to add this to the meta-data?

concepticondata/conceptlists/Xiao-2012-213.tsv
AnnikaTjuka: I see. We already mapped a similar case in He-2010-207-132 几个 [chinese] to SOME, so I will add this mapping.

concepticondata/conceptlists/Xiao-2012-213.tsv
AnnikaTjuka: Yes, we have ONE MILLION in concepticon. I'll add the mapping.

AnnikaTjuka:
> Ah is that the one you mentioned on Tuesday? Nice data set!
> 
Yes, that's the one :)

> I think "also" could be added as a new concept, if it's not too abstract, there are already several other lists which have it.
>
I'll add "also" to the [left-over concepts issue](https://github.com/concepticon/pyconcepticon/issues/20).

AnnikaTjuka:
This branch is now ready to be merged.

PR #713, opened 2020-01-20
Dawson-1887-731

Kristina-Pianykh:
Quite a large number of concepts in this list are supposed to introduce slight differences between species but in many cases this distinction is rather ambiguous or we don't have a separate concept set for such a specific instance. For example:

Cormorant (large)
Cormorant (small)

or

Butterfly (a large species)
Butterfly (small yellow)

Even though we have CORMORANT and BUTTERFLY, I left these unmapped.

Another thing concerns `concepticon test`. I first ran it with the missing list (but with the reference to it in conceptlists.tsv) but the test ran just fine and no errors were returned.


Schweikhard:
That's quite an exhaustive list in parts... I wouldn't actively know the term for the instep even in my native language.
The mapping looks good to me, except:

God should better be mapped to DEITY, that's less specific.

And HANDLE is by the current definition rather the handle of a pot etc., so Knife-handle might not fit that well. But there are other knife handles mapped to HANDLE already. Similar with POINT (POINTED END) and Knife-point.

Their "a moon" seems to be a unit of time (i.e. a lunar calendar month), not the actual MOON. That one can be found later. Also "a half moon" might here rather be half of a lunar calendar month and not that phase of the lunar cycle.

AnnikaTjuka:
I would suggest the following changes:

Unmap:
- [ ] Dawson-1887-731-84 "rump" and BUTTOCKS, because some languages have a specific term for rump, e.g. _Steiß_/_Rumpf_
- [ ] Dawson-1887-731-139 "doorway" and DOOR because some languages have a specific name for the doorway, e.g. _mouth of the door_.
- [ ] Dawson-1887-731-445 "maggot" and LARVA, because _maggot_ is "Made" and _larva_ is "Larve"
- [ ] Dawson-1887-731-449 "fly-blow" and FLY, because I don't know what a fly-blow is.

Change mapping:
- [ ] Dawson-1887-731-144 "blaze" better mapped to FIRE 221 or leave it unmapped

Specificity:
In some cases, the list specifies certain objects, as @Schweikhard mentioned in his comment on _knife handle_. I would suggest to leave those cases unmapped.
- [ ] Dawson-1887-731-17 "war-spear"
- [ ]  Dawson-1887-731-207 "basket (for food)"

Constructions with "and" and "or":
There are some cases where two words are connected. I would suggest to leave them unmapped, but not sure if there is already a standard for those cases.
- [ ] Dawson-1887-731-208 "wooden water box or bucket"
- [ ] Dawson-1887-731-337 "antlers and horns"
- [ ] Dawson-1887-731-566 "soul or spirit"
- [ ] Dawson-1887-731-640 "cap or hat"

Additional mappings:
- [ ] Dawson-1887-731-204 "borer" could be mapped to DRILL 3423
- [ ] Dawson-1887-731-668 "barn" could be mapped to BARN OR SHED 3025

AnnikaTjuka:
Additional comment/question:

- Are there any languages which have different terms for left and right arm?
If not, I wouldn't introduce the two seperate concepts LEFT vs. RIGHT ARM.

LinguList:
And thanks for adding this nice list.

Kristina-Pianykh:
Thank you everyone for your comments!

I don't really understand the algorithm of handling the cases with "and" and "or". I usually try to follow the already existing mappings on concepticon and in the case of Dawson-1887-731-337 'antlers and horns' I followed the mapping of Swadesh-1950-165-143 'horn or antler' (and the definition of HORN (ANATOMY) seems to fit here).

In the case of Dawson-1887-731-566 "soul or spirit" - Borin-2015-1532-231 'the soul or spirit', List-2014-1280-408 'soul, spirit', etc. So if such cases should rather be left unlinked, shouldn't we then modify the analogous mappings in other lists too?


As for Dawson-1887-731-668 "barn", I think it's too specific to be mapped to BARN OR SHED 3025 since the concept set refers to a hypernym: a building which serves two functions, namely storing agricultural products or sheltering livestock.

Good point @AnnikaTjuka about the distinction between LEFT vs. RIGHT ARM! I just checked the corresponding lexemes from this questionnaire and the one in Hale-1973-1798 and it seems like there are no separate forms for the whole concept of LEFT ARM or RIGHT ARM.

AnnikaTjuka:
In case of HORN, I agree with you. The concept description can refer to both 'antlers and horns'.

The "soul or spirit" mapping is a bit confusing because we have the two concepts SOUL and SPIRIT. For example, Bodt-2019-664-466 spirit, soul, shadow [english] is mapped to SPIRIT. I would suggest to open an issue about those mappings and leave it unmapped in this list for the time being.

I understand the BARN OR SHED concept as a building with one function either storing agricultural products or sheltering livestock. @Schweikhard could you help?

concepticondata/conceptlists/Dawson-1887-731.tsv
```diff
@@ -0,0 +1,732 @@
...
+Dawson-1887-731-534	534	My son's daughter			Kinship	19	34
+Dawson-1887-731-535	535	My mother	1216	MOTHER	Kinship	19	34
+Dawson-1887-731-536	536	My father's father	2560	PATERNAL GRANDFATHER	Kinship	19	34
+Dawson-1887-731-537	537	My wife	1199	WIFE	Kinship	19	34
+Dawson-1887-731-538	538	My husband	1200	HUSBAND	Kinship	19	34
+Dawson-1887-731-539	539	Male orphan			Kinship	19	34
+Dawson-1887-731-540	540	Female orphan			Kinship	19	34
+Dawson-1887-731-541	541	Family	194	FAMILY	Kinship	19	34
+Dawson-1887-731-542	542	Name of tribe			Social Organisation, etc.	20	34
+Dawson-1887-731-543	543	Indian			Social Organisation, etc.	20	34
```
chrzyki: We have 3612 potentially?

concepticondata/conceptlists/Dawson-1887-731.tsv
```diff
@@ -0,0 +1,732 @@
...
+Dawson-1887-731-428	428	To swim	1439	SWIM	Fish, etc.	13	33
+Dawson-1887-731-429	429	Spine on fishes back			Fish, etc.	13	33
+Dawson-1887-731-430	430	Roe			Fish, etc.	13	33
+Dawson-1887-731-431	431	Frog	503	FROG	Reptiles	14	33
+Dawson-1887-731-432	432	Lizard	632	LIZARD	Reptiles	14	33
+Dawson-1887-731-433	433	Snake	730	SNAKE	Reptiles	14	33
+Dawson-1887-731-434	434	Snake (water)			Reptiles	14	33
+Dawson-1887-731-435	435	Ant	587	ANT	Insects, etc.	15	33
+Dawson-1887-731-436	436	Bee	665	BEE	Insects, etc.	15	33
+Dawson-1887-731-437	437	Butterfly (a large species)			Insects, etc.	15	33
```
chrzyki: That could be 1791?

concepticondata/conceptlists/Dawson-1887-731.tsv
```diff
@@ -0,0 +1,732 @@
...
+Dawson-1887-731-438	438	Butterfly (small yellow)			Insects, etc.	15	33
```
chrzyki: And that is probably *Eurema lisa*, we don't have a concept for that.

concepticondata/conceptlists/Dawson-1887-731.tsv
```diff
@@ -0,0 +1,732 @@
...
+Dawson-1887-731-167	167	Spring	849	SPRING (OF WATER)	Dwellings	4	29
+Dawson-1887-731-168	168	Water	948	WATER	Dwellings	4	29
+Dawson-1887-731-169	169	Bow, of wood	994	BOW	Implements and Utensils	5	30
+Dawson-1887-731-170	170	Bow string	2119	BOW STRING	Implements and Utensils	5	30
+Dawson-1887-731-171	171	Arrow	977	ARROW	Implements and Utensils	5	30
+Dawson-1887-731-172	172	Notch in arrow for string			Implements and Utensils	5	30
+Dawson-1887-731-173	173	Arrow-head of stone			Implements and Utensils	5	30
+Dawson-1887-731-174	174	Arrow feathers			Implements and Utensils	5	30
+Dawson-1887-731-175	175	Quiver			Implements and Utensils	5	30
+Dawson-1887-731-176	176	War-club (stone)			Implements and Utensils	5	30
```
chrzyki: Potentially 1763 but I'd have to look at the concrete lexemes again.

concepticondata/conceptlists/Dawson-1887-731.tsv
```diff
@@ -0,0 +1,732 @@
...
+Dawson-1887-731-96	96	Toes	1389	TOE	Part of the Body	2	29
+Dawson-1887-731-97	97	Large toe			Part of the Body	2	29
+Dawson-1887-731-98	98	Second toe			Part of the Body	2	29
+Dawson-1887-731-99	99	Third toe			Part of the Body	2	29
+Dawson-1887-731-100	100	Toe nail	2662	TOENAIL	Part of the Body	2	29
+Dawson-1887-731-101	101	Blood	946	BLOOD	Part of the Body	2	29
+Dawson-1887-731-102	102	Vein or artery	525	BLOOD VESSEL	Part of the Body	2	29
+Dawson-1887-731-103	103	Brain	1583	BRAIN	Part of the Body	2	29
+Dawson-1887-731-104	104	Bladder	1746	BLADDER	Part of the Body	2	29
+Dawson-1887-731-105	105	Gall	1780	BILE	Part of the Body	2	29
```
chrzyki: I think this rather refers to the gall bladder, not the substance stored within.

chrzyki:
Thanks, @Kristina-Pianykh! Left a couple of minor comments.

concepticondata/conceptlists/Dawson-1887-731.tsv
Kristina-Pianykh: As we discussed above, 'war-club' is too specific to be mapped CLUB (analogous to the potential mapping of 'war-spear' to SPEAR)

concepticondata/conceptlists/Dawson-1887-731.tsv
Kristina-Pianykh: Will fix that.

concepticondata/conceptlists/Dawson-1887-731.tsv
chrzyki: Gotcha, sorry, missed that.

concepticondata/conceptlists/Dawson-1887-731.tsv
Kristina-Pianykh: The Collins dictionary says it is an obsolete term for 'bile', which may be the case considering the date of the list publication. Also there are other instances of 'gall' (see Swadesh-1950-165-136, Smith-2015-97-48) linked to BILE.

PR #717, opened 2020-01-23
Added automatically mapped data by Brysbaert-2014 in concept_set_metadata

AnnikaTjuka:
I added my first automatically mapped dataset to concept_set_metadata. The original file consists of 40,000 words and the algorithm found 2,221 matches. 

I used the Kuperman file as a template for the metadata.json.


AnnikaTjuka:
Yes, I ran conception test before I pushed the commit.

I will use a recent concept list metadata.json file as a template. Maybe this will solve the problem.

LinguList:
In any way nice that you already applied the workflow!

LinguList:
@AnnikaTjuka, the error is easy (see [travis](https://travis-ci.org/concepticon/concepticon-data/builds/640845683?utm_source=github_status&utm_medium=notification))

```
col = {'datatype': {'base': 'integer', 'minimum': 1}, 'name': 'CONCEPTICON_ID'}

conv = <class 'int'>

    def read_dicts(fname, schema=None, **kw):

        kw['dicts'] = True

        res = read_all(fname, **kw)

        if schema:

            def identity(x):

                return x

            colspec = {}

            for col in schema['columns']:

                conv = {

                    'integer': int,

                    'float': float,

>               }.get(col['datatype'])

E               TypeError: unhashable type: 'dict'
```

You should not code the values for datatype as a dictionary. So instead write:

```
        "name": "CONCEPTICON_ID",
        "datatype": "integer"
        }, 
```

@xrotwang, given that we CAN use the extended annotation for datatype, is this a bug in pyconcepticon?

AnnikaTjuka:
The checks are still failing although I did the following:
First, I changed the datatype of the column CONCEPTICON_ID.
Second, I used a recent metadata.json from a concept list as a template.

concepticondata/concept_set_meta/concreteness-brysbaert-2014.tsv-metadata.json
```diff
@@ -0,0 +1,54 @@
+{
```
LinguList: The metadata.json is for concept lists, but now you need the one we use for metadata. See here:

https://github.com/concepticon/concepticon-data/blob/master/concepticondata/concept_set_meta/SimLex999.tsv-metadata.json

As an example.

AnnikaTjuka:
Ok, I found the bug.
The datatype of the column CONCEPTICON_ID has to be "string" instead of "integer"!

Issue #710, opened 2020-01-12
Replace Heggarty-2017-200 with the now published Heggarty-2020-200 as referenced in Scarborough2020


LinguList:
finally published at least somewhere.

```bibtex
@InCollection{Scarborough2020,
  author    = {Scarborough, Matthew},
  title     = {Cognacy and computational cladistics: Issues in determining lexical cognacy for Indo-European cladistic research},
  booktitle = {Dispersals and diversification: Linguistic and archaeological perspectives on the early stages of Indo-European},
  year      = {2020},
  editor    = {Matilde Serangeli and Thomas Olander},
  publisher = {Brill},
  pages     = {179-208},
  address   = {Leiden},
}
```

LinguList:
![Scarborough2020](https://user-images.githubusercontent.com/2916210/72222598-de419e00-3566-11ea-94cc-619b56f0b6f8.png)
![Scarborough2020-1](https://user-images.githubusercontent.com/2916210/72222601-e00b6180-3566-11ea-8a39-b8c988600e51.png)

LinguList:
I suppose, whoever adds this list, should convert the two PNG files to pdf. Contact me if you need the article, I have it in PDF from the library.

PaulHeggarty:
This list is already in Concepticon:  https://concepticon.clld.org/contributions/Heggarty-2017-200

LinguList:
Then we should rename Heggarty-2017-200 to Scarborough-2017-200, since there is no real citation available for Heggarty-2017-200. Since Scarborough is citable, this one should be preferred (and Scarborough, unfortunately, made no reference to the Heggarty-list). The alternative is to check if they are really the same (no spelling differences, capitalization, etc.) and change the source in Heggarty-2017-200. 

The letter solution is the preferred one, as Heggarty is listed as main author (which is probably correct, and Anderson and Scarborough can be listed as the three main authors), and would otherwise not be mentioned. Then, following the procedure in the list of [Yakhontov-1991-100](https://concepticon.clld.org/contributions/Yakhontov-1991-100), we attribute the list to another author (team) than the source that mentions the list. 

We can even leave the date.

xrotwang:
There are differences:
- different order. "ant" is number 146 in Heggarty-2017-200, and first here
- different capitalization: Heggarty-2017-200 is all lowercase, this is all uppercase
- different wording of gloss: Heggarty-2017-200 has "bark", this has "(TREE) BARK"

[...]

PR #741, opened 2020-02-19
Replace Heggarty-2017-200 with Heggarty-2020-200

Kristina-Pianykh:
As mentioned in #710, I replaced Heggarty-2017-200 with Heggarty-2020-200 as published in Scarborough2020. The replacing list contains the items in alphabetical order (unlike the replaced list) so they had to be properly linked in `retired.json`.

I also left the **designers** of the list in the column AUTHOR in `conceptlists.tsv` and NOT the author of the publication where the list is published.

All the mappings are exactly the same as in the replaced list.


Kristina-Pianykh:
Yes, I mapped all the IDs manually.

Issue #733, opened 2020-02-11
RAISE (A CHILD) change to RAISE (BRING UP)


AnnikaTjuka:
The German mapping _aufziehen_ in Dellert-2017-1016-912 can also refer to animals. So it might be better to change the name of the concept to RAISE (BRING UP). I'll put this suggestion in an issue and map _rear_ to RAISE (A CHILD) for the time being.

_Originally posted by @AnnikaTjuka in https://github.com/concepticon/concepticon-data/pull/731_

Schweikhard:
Alternatively we could just add a new concept RAISE (BRING UP) that includes both raising a child and raising an animal, and leave RAISE (A CHILD) for those cases where it's clear.

AnnikaTjuka:
I'll change it to RAISE (BRING UP) because there is only one list which specifies child but uses "bring up": TeilDautrey-2008-430-61 bring up child

PR #749, opened 2020-02-21
Changed concept RAISE (A CHILD) to RAISE (BRING UP)

AnnikaTjuka:
This PR is based on the discussion in #733 . It changes the concept name and the mappings.

LibreOffice also deleted some leftover zeros in the Hale list. I hope this is ok.


chrzyki:
Looks fine to me, thanks!

Issue #730, opened 2020-02-07
HE-GOAT, MALE GOAT and FEMALE GOAT


Kristina-Pianykh:
In the presence of two duplicate concepts HE-GOAT and MALE GOAT, all the relevant glosses are mapped to the former while it is not listed as a replacement of the latter.

If we add the replacement ID of MALE GOAT to HE-GOAT, I suggest renaming FEMALE GOAT to SHE-GOAT for the sake of consistency in naming. There's just one gloss mapped to FEMALE GOAT so I assume renaming wouldn't be of a drastic impact.

AnnikaTjuka:
This schema could also be applied to boar or rather pig in general, see also #466.

LinguList:
Yes, we need consistency, but MALE vs. FEMALE are more flexible terms. So I suggest to test first to indicate HE-GOAT as an alias for MALE GOAT, so we can leave the other cases.

I also suggest to apply some query into meta-data and add information on male-female counterparts for animals and kinship terms. @Schweikhard may have a good idea how to name these, given that you have been working on the concept relations. I'd like to have this stored as information that SOW (FEMALE PIG) is the counterpart of BOAR (MALE PIG) [as a new concept]. 

Any ideas on a good name for a relation here? Maybe just "opposite_sex"?

And the rule for naming is: if there is a term, like "stallion" for male horse, we use that term and put the specification with "male/female" in brackets.

Schweikhard:
The relation name "opposite_sex" seems good - we would need consistency which sex we put into which column, though, as this is a non-directional relation and all other relations we have are directional, with the exception of antonymof.

Also, thus far all relation names just consist of a single alphabetic string, so that would then be "oppositesex", without underscore.

We basically always have a concept for the species in general as well, so the other solution would be to have "female" and "male" as relations from the general species to the female and male concepts, respectively.

LinguList:
Yes, female and male is okay. Or female_of and male_of, as this would
also work for German "Katze" male_of "Kater", right?

LinguList:
BUT, this would mean: male_of is only in one direction, male_of is not
the opposite of female_of, only in cases like katze and Kater, etc. but
not in "cattle" male_of bull, and "cattle" female of cow

Schweikhard:
> only in cases like katze and Kater, etc. 

And we would likely not have such an ambiguous concept as Katze ("cat in general or female cat") anyway.

> but not in "cattle" male_of bull, and "cattle" female of cow

[...]

Issue #466, opened 2018-04-13
BOAR 1348 needs cleanup


LinguList:
The concepts (reflexes) are mixed: some denote boar (wild pig) annd some denote male pig. Needs to be split and refined. We need two concepts with an extra definition and ideally better labels.

PR #753, opened 2020-02-24
Corrected names for male/female animals

Kristina-Pianykh:
I indicated 2475 MALE GOAT as the alias of HE-GOAT (as discussed in [#730](https://github.com/concepticon/concepticon-data/issues/730)) and moved all the mappings from HE-GOAT to MALE GOAT.

I also changed the name of BOAR to BOAR (MALE PIG) (see [#466](https://github.com/concepticon/concepticon-data/issues/466)).

Besides, some of Chinese 野猪 instances were mapped to BOAR instead of WILD BOAR (namely, those in Chen-2012-888 and Sun-1991-1004). It would be nice if @MacyL could have a quick look at these mappings just to make sure they're correct now.


chrzyki:
Hm, but we have to be careful with this, no? Not all boars are necessarily `male` if the concept is used to refer to the species: https://en.wikipedia.org/wiki/Wild_boar. Did you check for that?

Kristina-Pianykh:
Yes, BOAR (MALE PIG) is different from WILD BOAR in that the former is a pig of male sex while the latter - a different species (and thus could be both male and female).

chrzyki:
Good!

Kristina-Pianykh:
I just renamed the concept BOAR that was already mapped in a number of lists. In most of the cases, they specify the concept by adding _male pig_ to it. In regard to Chinese, 公猪 is clearly a domestic pig in contrast to 野猪 (as far as I can judge by Google images :) ).

Schweikhard:
The new concept name makes potential mismappings at least more obvious.

We might need some hyperonym over BOAR (MALE PIG) and WILD BOAR considering the ambiguity of the glosses, though.
Buck makes it clear in the book that he only means male pigs, and in most other cases it's contrasted with either sow or wild pig or both, but Hale and Lewis might both rather mean WILD BOAR from what I see there.

Kristina-Pianykh:
@Schweikhard you're right, I see now. Both Lewis and Hale have only two pig-related concepts: _hog_ and _boar_. I just briefly checked that there could potentially be another dimension of distinction. As far as I can tell, 'hog' refers to a castrated male pig while 'boar' - to an uncastrated male pig. But then I need to refine the current definition of BOAR (MALE PIG) and add a new one (HOG).

Schweikhard:
Oh, didn't notice/expect the hogs there.
There seems to be a UK/US difference.

Merriam Webster says it's "a domesticated swine especially when weighing more than 120 pounds (54 kilograms) " or "any of various animals related to the domesticated swine".

But Cambridge says:
"US: a pig, especially one that is allowed to grow large so that it can be eaten
UK: a male pig with its sexual organs removed, kept for its meat"

Which meaning can be expect Hale and Lewis to have meant? If they mean the US meaning, then hog might be PIG and boar is WILD BOAR. Otherwise, you're right and they make a new distinction.

Kristina-Pianykh:
Currently, within one list (in both Lewis and Hale) 'hog' is mapped to PIG and 'boar' - to BOAR (MALE PIG), which doesn't make much sense as they don't form an opposition. Lewis is affiliated with a university in the US so I'll remap 'boar' in his list to WILD BOAR.

Which meaning is used in Hale, however, is hard to tell.

Schweikhard:
Then I'd suggest to either treat Hale like Lewis, or to leave hog and boar in Hale unmapped.

AnnikaTjuka:
I added @MacyL as a reviewer, see [description](https://github.com/concepticon/concepticon-data/pull/753#issue-378968887) :

| Besides, some of Chinese 野猪 instances were mapped to BOAR instead of WILD BOAR (namely, those in Chen-2012-888 and Sun-1991-1004). It would be nice if @MacyL could have a quick look at these mappings just to make sure they're correct now.

chrzyki:
I'm fine with merging this, then @MacyL can prepare a clean PR.

MacyL:
I will close my PR (Chinese) to the BOAR branch and then prepare another one.

AnnikaTjuka:
I think there needs to be made one more commit concerning the Hale mapping.

LinguList:
Can we merge this?

Issue #665, opened 2019-10-30
Update on Polyglotta Africana/RefLex African List


chrzyki:
I recall this being discussed via mail and @Kristina-Pianykh waiting for input (if I remember correctly). Could we quickly discuss the status here? (Also to have it properly associated with the repository.

tresoldi:
We also need to check if it is the same concept list for `tls`, and how much overlap there is with the mapping I did in the past: https://github.com/lexibank/tls/blob/master/etc/concepts.csv

tresoldi:
This issue might be related to Nurse-1975 as well: https://github.com/concepticon/concepticon-data/issues/272

Kristina-Pianykh:
All the lists could be found [here](https://docs.google.com/spreadsheets/d/10VrKzLbB65kSm7JYT4S7uJLqNR1jJZ6sESGECApMECg/edit#gid=554505756). Polyglotta Africana (see Polyglotta retyped), the lists by Donzo (see Donzo Appendices B-C, Donzo Appendices 1-4), Marchese (see Marchese Kru updated) and Smeltzer (see Smeltzer Bozo updated) were all retyped. I tried my best linking the French glosses but I'm afraid my A0 language level in French requires the assistance of someone more proficient.

Check also Snider's list, @chrzyki kindly looked through my mappings and left some comments. I resolved some of the issues but others still need to be discussed.

Kristina-Pianykh:
The data from Snider's list was taken from [concepticon-dev](https://github.com/concepticon/concepticon-dev/blob/master/concepticondata/conceptlists/Snider-2004-1700.tsv), the source is on [comparalex](http://comparalex.org/?page=stdlist&id=13).

LinguList:
Do you also have the official PDF? It's called "SIL COMPARATIVE AFRICAN WORDLIST", and this should be easy to find online. The problem is: it is PDF, and comparalex did some work of themselves to get the list into their format, so it is not clear how regular the lists match (but we may also say we take it from comparalex and check, etc.).

Kristina-Pianykh:
Here the official publication [announced](https://www.sil.org/resources/publications/entry/7882) and here's the [PDF](https://www.sil.org/system/files/reapdata/88/54/65/88546558467011236880873045745569794523/silewp2006_005.pdf).

tresoldi:
[...]

PR #760, opened 2020-02-26
Smeltzer-1995-300

Kristina-Pianykh:
Another RefLex list (#665). Since it's entirely in French, I had to add English translations manually. The list includes many ambiguous glosses and there is no additional information in the source that could shed light on their exact meanings. For example:

1. _sentir_ can refer to feel, smell, taste.
2. _voler_ is used twice in the list. It has two meanings: 'to fly' and 'to steal'. Again, these two instances are not disambiguated.
etc.

In such cases, I listed the most probable meanings, separated with "/" since I can't leave them empty.

Two words seemed to be mistyped (_pèr_ and _battir_). They resemble closely 'per' (='father') and 'battre' (='to hit, beat'). Again, since I could not leave the corresponding cells in the ENGLISH column, I added the translations of the mistyped French glosses, marked them with a "?" and left unmapped.


concepticondata/conceptlists/Smeltzer-1995-300.tsv
```diff
@@ -0,0 +1,301 @@
...
+Smeltzer-1995-300-12	12	genou	1371	KNEE	knee
+Smeltzer-1995-300-13	13	peau	763	SKIN	skin
+Smeltzer-1995-300-14	14	plume	1201	FEATHER	feather
+Smeltzer-1995-300-15	15	corne	1393	HORN (ANATOMY)	horn
+Smeltzer-1995-300-16	16	queue	1220	TAIL	tail
+Smeltzer-1995-300-17	17	oeuf	744	EGG	egg
+Smeltzer-1995-300-18	18	lait	635	MILK	milk
+Smeltzer-1995-300-19	19	homme	1554	MAN	man
+Smeltzer-1995-300-20	20	femme	962	WOMAN	woman
+Smeltzer-1995-300-21	21	pèr			father?
```
AnnikaTjuka: I found beau-père [french] and  père [french]  in Key-2016. But I'd leave it unmapped if we cannot be sure that it is a typo.

concepticondata/conceptlists/Smeltzer-1995-300.tsv
```diff
@@ -0,0 +1,301 @@
...
+Smeltzer-1995-300-42	42	éléphant	1290	ELEPHANT	elephant
+Smeltzer-1995-300-43	43	chèvre	1502	GOAT	goat
+Smeltzer-1995-300-44	44	oiseau	937	BIRD	bird
+Smeltzer-1995-300-45	45	serpent	730	SNAKE	snake
+Smeltzer-1995-300-46	46	pou	1392	LOUSE	louse
+Smeltzer-1995-300-47	47	arbre	906	TREE	tree
+Smeltzer-1995-300-48	48	écorce	1204	BARK	bark (of tree)
+Smeltzer-1995-300-49	49	feuille	628	LEAF	leaf
+Smeltzer-1995-300-50	50	racine	670	ROOT	root
+Smeltzer-1995-300-51	51	graisse	323	FAT (ORGANIC SUBSTANCE)	fat, grease
```
AnnikaTjuka: graisse [french] was mappet to ORGANIC FAT OR OIL in  Grollemund-2015-100-28 but to FAT (ORGANIC SUBSTANCE) in Samarin-1967-100-32. Not sure which one is correct.

concepticondata/conceptlists/Smeltzer-1995-300.tsv
```diff
@@ -0,0 +1,301 @@
...
+Smeltzer-1995-300-67	67	siffler	1025	WHISTLE	to whistle
+Smeltzer-1995-300-68	68	dormir	1585	SLEEP	to sleep
+Smeltzer-1995-300-69	69	pousser	1452	PUSH	to push
+Smeltzer-1995-300-70	60	tirer	1455	PULL	to pull
+Smeltzer-1995-300-71	71	lancer	1456	THROW	to throw
+Smeltzer-1995-300-72	72	chanter	1261	SING	to sing
+Smeltzer-1995-300-73	73	sentir			to feel / to smell / to taste
+Smeltzer-1995-300-74	74	entendre	1408	HEAR	to hear
+Smeltzer-1995-300-75	75	nager	1439	SWIM	to swim
+Smeltzer-1995-300-76	76	s'asseoir	1649	SIT DOWN	to sit down
```
AnnikaTjuka: s'asseoir [french] was mapped to SIT in Key-2016-1310-118.

concepticondata/conceptlists/Smeltzer-1995-300.tsv
```diff
@@ -0,0 +1,301 @@
...
+Smeltzer-1995-300-77	77	se coucher	215	LIE DOWN	to lie down
+Smeltzer-1995-300-78	78	se lever	1512	RISE (ONESELF)	to rise (get up)
```
AnnikaTjuka: se lever [french] was mapped to RISE (MOVE UPWARDS) in Key-2016-1310-248

concepticondata/conceptlists/Smeltzer-1995-300.tsv
```diff
@@ -0,0 +1,301 @@
...
+Smeltzer-1995-300-209	209	bâton	1295	STICK	stick
+Smeltzer-1995-300-210	210	épine	124	THORN	thorn
+Smeltzer-1995-300-211	211	fleur	239	FLOWER	flower
+Smeltzer-1995-300-212	212	fruit	1507	FRUIT	fruit
+Smeltzer-1995-300-213	213	mil	931	MILLET	millet
+Smeltzer-1995-300-214	214	maïs	506	MAIZE	maize
+Smeltzer-1995-300-215	215	banane	868	BANANA	banana
+Smeltzer-1995-300-216	216	arachide	2364	PEANUT	peanut
+Smeltzer-1995-300-217	217	herbe	606	GRASS	grass
+Smeltzer-1995-300-218	218	“la brousse“	1683	BUSH	bush
```
AnnikaTjuka: Is there a reason why this in quotation marks? If I do a google search I get pictures of grass landscapes (German: Steppe).

concepticondata/conceptlists/Smeltzer-1995-300.tsv
```diff
@@ -0,0 +1,301 @@
...
+Smeltzer-1995-300-229	229	flèche	977	ARROW	arrow
+Smeltzer-1995-300-230	230	charbon	2658	COAL	coal
+Smeltzer-1995-300-231	231	matin	1339	MORNING	morning
+Smeltzer-1995-300-232	232	foudre	1654	LIGHTNING	lightning
+Smeltzer-1995-300-233	233	houe	284	HOE	hoe
+Smeltzer-1995-300-234	234	argile	1399	CLAY	clay
+Smeltzer-1995-300-235	235	vingt	1710	TWENTY	twenty
+Smeltzer-1995-300-236	236	cent	1634	HUNDRED	hundred
+Smeltzer-1995-300-237	237	profond	1593	DEEP	deep
+Smeltzer-1995-300-238	238	grand	711	TALL	tall, large
```
AnnikaTjuka: This also seems to be an ambiguous case. I found some list that mapped grand to BIG (e.g., Meillet-1921-16-13 | grand [french])

concepticondata/conceptlists/Smeltzer-1995-300.tsv
Kristina-Pianykh: I guess the reason why Grollemund-2015-100-28 is mapped to ORGANIC FAT OR OIL is because its corresponding English gloss (already given in the source) is 'fat, oil'.
But judging by the wikipedia article, it refers more to fats (https://fr.wikipedia.org/wiki/Graisse).

concepticondata/conceptlists/Smeltzer-1995-300.tsv
Kristina-Pianykh: This is one of the cases where I suspect there is a typo in père ('father').

concepticondata/conceptlists/Smeltzer-1995-300.tsv
Kristina-Pianykh: It can mean both but here I made a decision based on the surrounding concepts, all denoting a change in spatial position in regard to a person. Besides, I assume that RISE (MOVE UPWARDS) would then have a counterpart like GO DOWN (DESCEND) but this is not the case.

concepticondata/conceptlists/Smeltzer-1995-300.tsv
Kristina-Pianykh: I found it confusing too since it's the only concept enclosed in quotation marks in the source. I checked with Aymeric and he said it's quite comparable to English 'bush'. The Google images, however, seem to refer to a certain type of vegetation, as you said. I think I'll unmap it for now.

AnnikaTjuka: I think "bush" is also ambiguous in English. It can mean the plant or wild grass land.

concepticondata/conceptlists/Smeltzer-1995-300.tsv
Kristina-Pianykh: This one is extremely ambiguous. As far as I understood from Aymeric's explanation, its proximity to 'gros' points to a certain kind of 'big' in the vertical dimension which is tall in this case. On the other hand, 'gros' refers to a big size more in the horizontal sense.

concepticondata/conceptlists/Smeltzer-1995-300.tsv
Kristina-Pianykh: Hm, in Key-2016-1310 it might be the case that it's the English gloss 'sit' that was taken as the source concept.
I looked up what verbal 'se' in French can mean and it turns out that the verbs with 'se' acquire the reflexive meaning (I guess comparable to German _sich setzen_) while without the reflexive 'se' it could be used transitively (=to sit down someone).

concepticondata/conceptlists/Smeltzer-1995-300.tsv
Kristina-Pianykh: You're right but I guess the current concepticon definition points to an individual plant.

concepticondata/conceptlists/Smeltzer-1995-300.tsv
```diff
@@ -0,0 +1,301 @@
...
+Smeltzer-1995-300-290	290	essuyer	1454	WIPE	to wipe
+Smeltzer-1995-300-291	291	balayer	230	SWEEP	to sweep
+Smeltzer-1995-300-292	292	ouvrir	1156	OPEN	to open
+Smeltzer-1995-300-293	293	casser	2558	BREAK (CLEAVE)	to break
+Smeltzer-1995-300-294	294	couper	1432	CUT	to cut
+Smeltzer-1995-300-295	295	percer	398	PIERCE	to pierce
+Smeltzer-1995-300-296	296	lutter	1423	FIGHT	to fight
+Smeltzer-1995-300-297	297	nouer (attacher)	1917	TIE	to tie
+Smeltzer-1995-300-298	298	délier	352	UNTIE	to untie
+Smeltzer-1995-300-299	299	battir			to hit, beat?
```
Schweikhard: There's currently an issue open on HIT etc, but based on the discussion I would say:
Is it to hit intentionally a single time? Then it's STRIKE. If intentional several times, POUND.

Kristina-Pianykh: In this case, it doesn't matter because this word is very likely to be mistyped and my translation given here corresponds to the word 'battre' that in its turn means to hit or beat. The idea was to leave such ambiguous glosses unmapped just to be on the safe side.

Issue #655, opened 2019-10-10
Pot, clay pot, kettle, pan, wok, jug, earthen jug


Schweikhard:
POT https://concepticon.clld.org/parameters/1462
CLAY POT https://concepticon.clld.org/parameters/2055
KETTLE https://concepticon.clld.org/parameters/854
PAN https://concepticon.clld.org/parameters/31
WOK https://concepticon.clld.org/parameters/2056
JUG https://concepticon.clld.org/parameters/270
EARTHEN JUG https://concepticon.clld.org/parameters/3409

I think currently the English glosses are mostly where they should be (except for the ceramic pots at POT - but see below), but I'm not sure about some of the Chinese ones (e.g. there is something looking like a WOK linked to PAN).
The Hausa large water pot/ràndā likely should move to EARTHEN JUG but I don't know for sure.

Also, there may be some POTs which are actually not used for cooking but for storage (but there are no clear cases from what I see).
The unambiguous cooking pots could be moved to a new concept.
And CLAY POT is only for storage pots (if the definition is reliable - only thing linked is 土罐), but we also have some clay cooking pots, so those would need their own concept in that case.

MacyL:
In Huang-1992-1820 罐子 pot/jar/tin should be linked to CAN [1051] (the same gloss in Chen-2012-888)

Schweikhard:
Thanks for fixing the wok, @MacyL - could you check the Chinese glosses of the other concepts to see whether they are correct now?

In general, as _pot_ does not necessarily imply a cooking vessel but just about any vessel, I would suggest to add a new concept COOKING POT and map all pots that are specified to be used for cooking to that.
We could also add a new concept for WATER POT (for Kraft and Othaniel).

MacyL:
My current principle is that "鍋" or "鍋子" will not be mapped to any concept. Because the image in my mind is like this ---> https://images.app.goo.gl/F1r8BGXqCeppdsjL9
罐子 is linked to CAN 
炒菜鍋 or 鍋(炒菜鍋）is linked to WOK ----> https://images.app.goo.gl/sJqUik8voTproU2w6
luckily, we don't have the concept  "鍋" or "鍋子" yet, but it's good to have a new concept "COOKING POT" for the future.

Schweikhard:
To summarize: I'll move  "鍋" or "鍋子" to COOKING VESSEL, 炒菜鍋 or 鍋(炒菜鍋) to WOK, and 罐子 to POT?

[...]

PR #761, opened 2020-02-26
Fixing oar, bad and river

### LINKED ISSUE #566: PADDLE/OAR
The definition of OAR needs to be made more precise, and the lists linked to PADDLE need to be checked for accurate linking.
There is now a hyponym.

### LINKED ISSUE #567: RIVER
There is a new concept FLOWING BODY OF WATER, which is a hyperonym to RIVER (large river) and STREAM (small river).
Some lists linked to RIVER may need to be moved to the hyperonym.
https://concepticon.clld.org/parameters/666

### LINKED ISSUE #580: BAD/EVIL
I changed the definition of EVIL from "Intending to harm" to "Intending to harm or being/acting ethically wrong." but some things linked might be moved to new hyponyms of either meaning.
BAD also contains some mislinkings or unclear cases.

Schweikhard:
Fixes  #566
Fixes #567
Fixes #580

This one introduces one new concept (and changes two definitions), so please proceed with care regarding additional PRs while this is open.

Btw, Libre Office changed 0.000 to 0, and e.g. 0.50 to 0.5 in Bowern and adds quotation marks in two cases in Castro - should I fix that or is it not an issue?


Schweikhard:
> Not sure about the quotation marks though.

I guess best would be if I'd just turn " into &quot;, wouldn't it?

> I only use a text editor to open other lists (especially with Chinese letters) from now on because you never know what sneaky changes LibreOffice will do ;-)

I just set all the columns to Text, then it won't do too much nasty stuff - but I don't always bother to do so.

Schweikhard:
Thanks, will do so!

> Also, if the main distinction between paddle and oar is whether the pole is attached to the boat, maybe we should make that difference obvious in the definition of PADDLE, too? Just to avoid confusion, because if someone thinks of the oars used in Portuguese galleys, they are usually used with two hands as well and could be mapped to PADDLE under the current definition.

Ah, didn't think of that interpretation.

Schweikhard:
Updated the definitions and quotes, and now also took care of the pots, adding two new concepts.

Fixes #655

concepticondata/conceptlists/Castro-2015-608.tsv
```diff
@@ -260,7 +260,7 @@ Castro-2015-608-258	258	home	2406	HOME	家	home, household
 Castro-2015-608-259	259	garden	586	GARDEN	园（菜园）	vegetable plot, garden
 Castro-2015-608-260	260	shadow	1388	SHADE	影子	shadow
 Castro-2015-608-261	261	dream	2374	DREAM	梦	dream
-Castro-2015-608-262	262	ghost	53	SPIRIT	鬼（鬼神）	"ghost, god (a ""large"" spirit - can possess people etc.)"
+Castro-2015-608-262	262	ghost	53	SPIRIT	鬼（鬼神）	ghost, god (a &quot;large&quot; spirit - can possess people etc.)
```
chrzyki: Problematic quoting escapes.

concepticondata/conceptlists/Castro-2015-608.tsv
```diff
@@ -600,7 +600,7 @@ Castro-2015-608-598	598	a while	726	WHILE	一会儿（坐一会儿）	a while
 Castro-2015-608-599	599	for a moment	374	ONE TIME	一下（过来一下）	for a moment
 Castro-2015-608-600	600	identical	2276	IDENTICAL	一样（一模一样）	the same (two objects are identical, e.g. two stools)
 Castro-2015-608-601	601	look alike	1777	RESEMBLE	像（他像他爸爸）	to be alike (e.g. he looks like his father)
-Castro-2015-608-602	602	together	2482	TOGETHER	一起，同（“我们一起去”）	"together (""We're going together"")"
+Castro-2015-608-602	602	together	2482	TOGETHER	一起，同（“我们一起去”）	together (&quot;We're going together&quot;)
```
chrzyki: Problematic quoting escapes.

concepticondata/conceptlists/Castro-2015-608.tsv
chrzyki: just " " around large here should be fine.

concepticondata/conceptlists/Castro-2015-608.tsv
chrzyki: Likewise, "We're going together" should be fine.

concepticondata/conceptlists/Castro-2015-608.tsv
Schweikhard: Ah I thought it would be better that way - and there are already other files in concepticon that escape quotes.

chrzyki: In the end all of them are converted for the CLLD app anyway, but I think it makes for easier readable lists.

Schweikhard: Makes sense. It's not in columns that are read very often anyway, but I can change it back. Escaping the quotes only has the advantage that Libre Office doesn't mess it up.

Issue #506, opened 2018-08-18
shout vs. shriek


Schweikhard:
I defined SHOUT as "To utter in a loud voice." 
But it's actually not clear whether that's what's meant in all cases.
Some entries linked to SHOUT might mean instead SHRIEK, defined as "To make a very high, loud sound.", but there is no completely clear case.
And some of the entries linked to CALL might also rather fit one of the other two.

shout https://concepticon.clld.org/parameters/715
shriek https://concepticon.clld.org/parameters/298
call https://concepticon.clld.org/parameters/1084

AnnikaTjuka:
Do you mean the ones with "cry out" should be mapped to SHRIEK ?

In case of CALL, I am not sure which could be mapped to either SHOUT or SHRIEK.

Schweikhard:
Yes.

Of CALL, "call out" rather seems like either SHOUT or SHRIEK - or it means something different altogether.

AnnikaTjuka:
For me, "call out" means something like "call out her name" to get somebody to the dinner table or similar. Therefore, I'd leave "call out" mapped to CALL although one might raise her voice for that purpose.

Schweikhard:
To me, "call out" means "aufschreien" if it's used intransitively, but the elicitation glosses don't provide any information regarding that.

It seems transitively it can also mean "aufrufen" or "verlesen" (the latter being one of the two meanings Leo provides besides "herausrufen").

Cambridge has for call out (something):
"to speak loudly
to officially order that a military unit become ready for action"

And Merriam Webster provides for call out:
"1: to summon into action
call out troops
2 : to challenge to a duel
3 : to order on strike
call out the workers
4 : to publicly criticize or fault (someone) "

So, it seems really ambiguous and could be argued to fit any of the three (mostly SHOUT, though).

AnnikaTjuka:
I get your point. But for me (from a user's perspective) it is a bit confusing if we map a gloss with a verb "call (out)" that we have as a concept CALL to another concept/verb SHOUT. 

Maybe there are conventions in Concepticon that I am not aware of and @LinguList could decide in this case.

Schweikhard:
Therefore I would argue to unmap those two instances of "call out" as they are just too uncertain.

AnnikaTjuka:
That would be a solution. I'd suggest you'll make a PR with the proposed changes so that the others can join the discussion.

PR #770, opened 2020-02-28
call out

Schweikhard:
As discussed in #506 I unmapped 'call out' from CALL as it seems too ambiguous.

In the same issue, we are discussing SHOUT and SHRIEK, as there might be some on the former that would better fit the latter. Input is appreciated!

### DIFF
--- a/concepticondata/concepticon.tsv
+++ b/concepticondata/concepticon.tsv
@@ -296,7 +296,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 295	WHETSTONE	Basic actions and technology	Before cutting the meat, he sharpened his knife with a whetstone.	Person/Thing	
 296	CROUCH	Motion	To lower one's body close to the ground by bending one's knees completely.	Action/Process	
 297	CROOKED	Spatial relations	Opposite of straight.	Property	
-298	SHRIEK	Speech and language	To make a very high, loud sound.	Action/Process	
+298	SHRIEK	Speech and language	To make a very high, loud sound in reaction to something.	Action/Process	
 299	SCOLD	Speech and language	To rebuke.	Action/Process	
 300	RATTLE	Speech and language	A wooden instrument that makes a loud knocking noise [used by people watching football games].	Person/Thing	
 301	CHIEFTAIN	Social and political relations	The leader/ruler of a tribe/clan.	Person/Thing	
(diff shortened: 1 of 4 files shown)


Schweikhard:
The definitions seemed clear enough to me, but we can certainly rephrase them if that makes them easier to understand.

I only turned to dictionaries because the elicitation glosses were ambiguous, not because the concept definitions seemed to be.

So perhaps:
SHRIEK: To make a very high, loud sound in reaction to something.
SHOUT: To say something in a loud voice.

AnnikaTjuka:
Maybe you could add the changes concerning SHRIEK and SHOUT so that the whole issue #506 is solved in one PR.

Schweikhard:
I would but I'm not sure which things to move from SHOUT to SHRIEK - most of the ones in question are in Chinese (喊 and 喊叫).

Buck contains both SHOUT and SHRIEK (and even some things that additionally mean CALL). I think SHOUT is the main meaning, but it's not really clear.

The only other one that I think might be worth moving would be
Pallas: Crier [french]; Kreißchen [german]; Clamare [latin]; Кричать [russian]
But the German and the Latin are ambiguous, and I don't know about the French and Russian.

AnnikaTjuka:
My guess is that the Chinese cases can stay where they are because only the English translations differ. But maybe @MacyL can check? 

I'd also leave the ones with "shout, cry out". 

And if Кричать means "shout", I'd also leave the other cases with different language glosses mapped to SHOUT. @Kristina-Pianykh could you confirm?

I think the only confusing gloss is the German "Kreißchen" which would better fit to SHRIEK. But it is also a very old list. Maybe "Kreißchen" had the meaning "schreien" at that time?

MacyL:
> My guess is that the Chinese cases can stay where they are because only the English translations differ. But maybe @MacyL can check?
> 
> I'd also leave the ones with "shout, cry out".
> 
> And if Кричать means "shout", I'd also leave the other cases with different language glosses mapped to SHOUT. @Kristina-Pianykh could you confirm?
> 
> I think the only confusing gloss is the German "Kreißchen" which would better fit to SHRIEK. But it is also a very old list. Maybe "Kreißchen" had the meaning "schreien" at that time?

Short answer : The mappings of 喊叫 and 喊, from what I have seen so far is correct. 
Long answer : 喊叫 is SHOUT. Also, it is correct to unmap the concept 喊(~人开会) in Huang's data, because it includes a lot of various cases. For example,  我喊他叫叔叔 I call him uncle. 我喊他來,  I call him to come.

Schweikhard:
> I think the only confusing gloss is the German "Kreißchen" which would better fit to SHRIEK. But it is also a very old list. Maybe "Kreißchen" had the meaning "schreien" at that time?

The examples here sound like that at least at Grimm's time, it already meant mainly SHRIEK http://woerterbuchnetz.de/cgi-bin/WBNetz/wbgui_py?sigle=DWB&mode=Vernetzung&lemid=GK13207#XGK13207

> Short answer : The mappings of 喊叫 and 喊, from what I have seen so far is correct.
Long answer : 喊叫 is SHOUT. Also, it is correct to unmap the concept 喊(~人开会) in Huang's data, because it includes a lot of various cases. For example, 我喊他叫叔叔 I call him uncle. 我喊他來, I call him to come.

Thanks! So, Huang's 喊叫 [chinese]; yell / shout [english] being mapped to SHOUT is correct?

AnnikaTjuka:
In sum, the proposed changes to solve #506 are:

**Unlink**:
- Pallas-1786-442-62	Crier [french]; Kreißchen [german]; Clamare [latin]; Кричать [russian]
- Zgraggen-1980-380-367	call out [english]
- Robinson-2012-398-46	call out [english]

**Change definitions:**
- SHRIEK: To make a very high, loud sound in reaction to something.
- SHOUT: To say something in a loud voice.

Correct @Schweikhard ?

Schweikhard:
Thanks for summarizing. I guess that's what we can agree on. I'll add the remaining changes.

Issue #758, opened 2020-02-25
Gauchat-1925-480


LinguList:
I have started to map the concepts [here](https://docs.google.com/spreadsheets/d/1rabxVu7JK7ux0StvjGwS3QnvTtPWI9YPXl-nIjdWU7M/edit?usp=sharing). 

The linking is not optimal and many fields have to be left empty, but it is still useful to add this list, as it is a frequent type of questionnaire in dialect studies.

LinguList:
Bibtex is in [Evobib](https://digling.org/evobib/?bibtex=Gauchat1925):

```
@book{Gauchat1925,
    address = {Neuchâtel},
    editor = {Gauchat, Louis and Jeanjaquet, Jules and Tappolet, Ernest},
    keywords = {tppsr},
    publisher = {Attinger},
    title = {Tableaux phonétiques des patois suisses romands.},
    year = {1925}
}
```

PR #772, opened 2020-03-02
Gauchat-1925-480

Kristina-Pianykh:
Refers to #758.


laiyunfan:
I think the mapping is alright.

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-101	101	We have	On a	hŏmo habet	34		
+Gauchat-1925-480-102	102	lost	perdu	perdūtu	34	1660	LOSE
+Gauchat-1925-480-103	103	the key	la clef	clave	36	1472	KEY
+Gauchat-1925-480-104	104	of the bell tower	du clocher	cloccariu	36		
+Gauchat-1925-480-105	105	It (feminine) is	Elle est	ĭlla est	36		
+Gauchat-1925-480-106	106	lost (feminine)	perdue	perdūta	36		
+Gauchat-1925-480-107	107	Who has	Qui a	cui / qualis habet	36		
+Gauchat-1925-480-108	108	done that?	fait cela?	factu	36		
+Gauchat-1925-480-109	109	I go	Je vais	vado	38		
+Gauchat-1925-480-110	110	to the market	au marché	mercatu	38	633	MARKET
```
Schweikhard: Contains a preposition (but so do all the genitives), not sure whether to include.

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-196	196	some cabbage	de chou	caule	66	774	CABBAGE
+Gauchat-1925-480-197	197	a wreath	une gerbe	garba	66	2589	WREATH
+Gauchat-1925-480-198	198	some straw	de paille	palea	66		
+Gauchat-1925-480-199	199	a bag	un sac	saccu	68	2468	BAG
+Gauchat-1925-480-200	200	some oat	d’avoine	avēna	68	60	OAT
+Gauchat-1925-480-201	201	some flour	de farine	farina	68	1594	FLOUR
+Gauchat-1925-480-202	202	an armful	une brassée	braciata	68		
+Gauchat-1925-480-203	203	some armfuls	(des) brassées	braciatas	68		
+Gauchat-1925-480-204	204	of wood	de bois	bŏscu	68	1803	WOOD
+Gauchat-1925-480-205	205	(a) shovel	(une) pelletée	palata	70
```
Schweikhard: 1901 SHOVEL

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-340	340	I cry	Je pleure	plōro	114		
+Gauchat-1925-480-341	341	slowly	doucement	-mĕnte	114	2165	SLOWLY
+Gauchat-1925-480-342	342	so that we don’t hear	pour qu’on n’entende pas	hŏmo audiat	114		
+Gauchat-1925-480-343	343	(I) can	(Je) sais	sa(p)io	114		
+Gauchat-1925-480-344	344	some	un peu	paucu	114	1241	SOME
+Gauchat-1925-480-345	345	read	lire	lĕgere	114	724	READ
+Gauchat-1925-480-346	346	I say that to nobody	Je ne le dis	dīco	116		
+Gauchat-1925-480-347	347	than to them	qu’à eux	illōru	116		
+Gauchat-1925-480-348	348	It beats	Il frappe	fĕrit	116		
+Gauchat-1925-480-349	349	one o'clock	une heure	hōra	116	1353	HOUR
```
Schweikhard: That's the time, not the noun, I think.

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-418	418	in one eye	à un œil	ŏclu	140		
+Gauchat-1925-480-419	419	in the eyes	aux yeux	ŏclos	140		
+Gauchat-1925-480-420	420	(in the) tongue	(à la) langue	lĭngua	140	1205	TONGUE
+Gauchat-1925-480-421	421	(in the) head	(à la) tête	tĕsta	140	1256	HEAD
+Gauchat-1925-480-422	422	(in one) nail	(à un) ongle	ŭngla	140	1258	FINGERNAIL
+Gauchat-1925-480-423	423	(in one) ear	(à une) oreille	aurĭcla, -ŭcla	140	1247	EAR
+Gauchat-1925-480-424	424	He fell	Il est tombé	cadūtu, -ēctu	142		
+Gauchat-1925-480-425	425	(she) fell	(elle est) tombée	cadūta, -ēcta	142		
+Gauchat-1925-480-426	426	on his/her buttocks	sur son derrière	cūlu	142	834	BUTTOCKS
+Gauchat-1925-480-427	427	He falls	Il tombe	cadit	142
```
Schweikhard: 1280 FALL (or 2894 TUMBLE)

concepticondata/conceptlists/Gauchat-1925-480.tsv
Schweikhard: Ah there's already another SHOVEL above - dunno which fits better.

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
+ID	NUMBER	ENGLISH	FRENCH	LATIN	Page	CONCEPTICON_ID	CONCEPTICON_GLOSS
+Gauchat-1925-480-1	1	It is	Il fait	facit	2		
+Gauchat-1925-480-2	2	hot	chaud	caldu	2	1286	HOT
+Gauchat-1925-480-3	3	today	aujourd‘hui	hŏdie	2	1283	TODAY
+Gauchat-1925-480-4	4	This is	C’est	est	2		
+Gauchat-1925-480-5	5	a good	un bon	bŏnu	2	1035	GOOD
```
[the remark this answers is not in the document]
Kristina-Pianykh: The thing with list is that each gloss here should not be treated individually but rather in a context of surrounding words because obviously this list is just a bunch of sentences broken down into phrases. Here the sentence goes 'This is a good weather' (the noun 'weather' being apparently countable in French) so it just happened that the article is joined to the adjective.

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-69	69	are very close	sont tout près	prĕssu	24		
+Gauchat-1925-480-70	70	of the village	du village	villaticu	24	930	VILLAGE
+Gauchat-1925-480-71	71	Let's go	Allons-y	ambulemus	24		
+Gauchat-1925-480-72	72	together	ensemble	insĭmul	24	2482	TOGETHER
+Gauchat-1925-480-73	73	Be careful	Prends garde	cavica	26		
+Gauchat-1925-480-74	74	to stumble	de t‘“encoubler“	incopulare	26		
+Gauchat-1925-480-75	75	The bridge	Le pont	pŏnte	26	1011	BRIDGE
+Gauchat-1925-480-76	76	is	est en	est ĭn	26		
+Gauchat-1925-480-77	77	stone	pierre	pĕtra	26	857	STONE
+Gauchat-1925-480-78	78	iron	en fer	fĕrru	26	621	IRON
```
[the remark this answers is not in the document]
Kristina-Pianykh: I'm not sure I understood what you mean. I just checked that French 'fer' in other lists is also mapped to IRON 621.

concepticondata/conceptlists/Gauchat-1925-480.tsv
Kristina-Pianykh: I think you're right here, I'll unmap it.

concepticondata/conceptlists/Gauchat-1925-480.tsv
Kristina-Pianykh: I suspect the mapping strategy here was to attempt to map infinitives rather than inflectional verb forms. But since there is no other instance of 'fall', this one can be mapped.

concepticondata/conceptlists/Gauchat-1925-480.tsv
Kristina-Pianykh: The French pelletée refers more to quantity, comparable to 'a handful' or 'a spoonful'. The context here would then be something like 'a spadeful of coal', I guess.

concepticondata/conceptlists/Gauchat-1925-480.tsv
Kristina-Pianykh: As this list represents a set of sentences divided very roughly into phrases, I would ignore function words, especially with nouns, and try to map them. What is important to remember here, is that it's clearly not a standard kind of list with base word forms, and addressing only the base forms would not work.

concepticondata/conceptlists/Gauchat-1925-480.tsv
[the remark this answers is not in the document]
Schweikhard: True, but this would mean that the elicited forms also contain all the parts in that segment. So this one also contains the article if the elicited dialect uses indefinite articles.

concepticondata/conceptlists/Gauchat-1925-480.tsv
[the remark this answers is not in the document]
Schweikhard: But "en fer" means "made of iron", "eisern", doesn't it?

It's basically the same issue as with the au marché below I guess.

concepticondata/conceptlists/Gauchat-1925-480.tsv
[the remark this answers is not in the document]
Schweikhard: And that is less of a problem with nouns, but since this one is an adjective, if this segment then stands alone, it looks as if the article would nominalize it. So if anyone would use that data in wordlist format, this would not only change the grammatical aspects of the word but turn it into a different part of speech.

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-40	40	snow	de la neige	nĭve	14	784	SNOW
+Gauchat-1925-480-41	41	ice	de la glace	glacia	14	617	ICE
+Gauchat-1925-480-42	42	It is (2)	Il fait	facit	14		
+Gauchat-1925-480-43	43	some wind	du vent	vĕntu (aura)	16	960	WIND
+Gauchat-1925-480-44	44	Winter	L’hiver	hibĕrnu	16	1471	WINTER
+Gauchat-1925-480-45	45	arrives	arrive	adrīpat	16	1387	ARRIVE
+Gauchat-1925-480-46	46	It is there	Il est là	illac	16		
+Gauchat-1925-480-47	47	I must	Je dois	debēo	16		
+Gauchat-1925-480-48	48	go down	descendre	descĕndere	16	2014	GO DOWN (DESCEND)
+Gauchat-1925-480-49	49	in the city	en ville	vīlla	18
```
chrzyki: Question: If function words should generally be ignored, why can't this be CITY then? "from the thunder" above is mapped to THUNDER.

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-93	93	of the roof	du toit	tēctu	32	769	ROOF
+Gauchat-1925-480-94	94	He closes	Il ferme	claudit	32	1046	SHUT
+Gauchat-1925-480-95	95	the window	la fenêtre	fenĕstra	32	1375	WINDOW
+Gauchat-1925-480-96	96	of the room	de la chambre	pē(n)sile	32	1450	ROOM
+Gauchat-1925-480-97	97	He sweeps	Il balaie	scōpat	34	230	SWEEP
+Gauchat-1925-480-98	98	in front of	devant	de + abante	34	1883	IN FRONT OF
+Gauchat-1925-480-99	99	the door	la porte	pŏrta	34	1567	DOOR
+Gauchat-1925-480-100	100	of the barn	de la grange	granica	34	3025	BARN OR SHED
+Gauchat-1925-480-101	101	We have	On a	hŏmo habet	34		
+Gauchat-1925-480-102	102	lost	perdu	perdūtu	34	1660	LOSE
```
chrzyki: Isn't this the state (something is lost) instead of the action?

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-104	104	of the bell tower	du clocher	cloccariu	36		
+Gauchat-1925-480-105	105	It (feminine) is	Elle est	ĭlla est	36		
+Gauchat-1925-480-106	106	lost (feminine)	perdue	perdūta	36		
+Gauchat-1925-480-107	107	Who has	Qui a	cui / qualis habet	36		
+Gauchat-1925-480-108	108	done that?	fait cela?	factu	36		
+Gauchat-1925-480-109	109	I go	Je vais	vado	38		
+Gauchat-1925-480-110	110	to the market	au marché	mercatu	38	633	MARKET
+Gauchat-1925-480-111	111	to search	chercher	circare	38	1468	LOOK FOR
+Gauchat-1925-480-112	112	a shovel	une pelle	pala	38	1901	SHOVEL
+Gauchat-1925-480-113	113	(a) scythe	(une) faux	falce	38	2132	SICKLE OR SCYTHE
```
chrzyki: 20 SCYTHE?

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-114	114	(a) scythe handle	(un) manche de faux	falcariu	38		
+Gauchat-1925-480-115	115	(a) plague	(un) fléau	flagĕllu	40		
+Gauchat-1925-480-116	116	(a) van	(un) van	vannu	40		
+Gauchat-1925-480-117	117	(a) boiler	(une) chaudière	caldaria	40		
+Gauchat-1925-480-118	118	(a) plow	(une) charrue	carrūca	40	1921	PLOUGH
```
chrzyki: 1921 is the action, I think this should be 2154?

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-175	175	is sold	est vendu	vendūtu	60		
+Gauchat-1925-480-176	176	The cow	La vache	vacca	60	1007	COW
+Gauchat-1925-480-177	177	(is) sold (feminine)	(est) vendue	vendūta	60		
+Gauchat-1925-480-178	178	The goats	Les chèvres	capras	60	1502	GOAT
+Gauchat-1925-480-179	179	(are) sold	(sont) vendues	vendūtas	60		
+Gauchat-1925-480-180	180	It is expensive	Il est cher	caru	60		
+Gauchat-1925-480-181	181	(she is) expensive	(Elle est) chère	cara	62		
+Gauchat-1925-480-182	182	(they are) expensive	(elles sont) chères	caras	62		
+Gauchat-1925-480-183	183	The pig	Le porc	pŏrcu	62	1337	PIG
+Gauchat-1925-480-184	184	is fat	est gros	grŏssu	62	2449	FAT (OF MEAT AND FOOD)
```
chrzyki: This is the state of 'being big' I think?

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-226	226	(the apple is) eaten	(la pomme est) mangée	manducata	76		
+Gauchat-1925-480-227	227	The two halves	Les deux moitiés	dŭas medietates	76		
+Gauchat-1925-480-228	228	(are) eaten	(sont) mangées	manducatas	76		
+Gauchat-1925-480-229	229	I needed	Il m’a fallu	fallūtu	78		
+Gauchat-1925-480-230	230	to drink	boire	bĭbere	78	1401	DRINK
+Gauchat-1925-480-231	231	some milk	du lait	lacticĕllu	78	635	MILK
+Gauchat-1925-480-232	232	some whey	du petit-lait	lactata	78		
+Gauchat-1925-480-233	233	The wine	Le vin	vīnu	78	1524	WINE
+Gauchat-1925-480-234	234	is clear	est clair	claru	78	250	LIGHT (COLOR)
+Gauchat-1925-480-235	235	lukewarm	tiède	tĕpidu	80	1232	WARM
```
chrzyki: lukewarm is not warm.

concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -0,0 +1,481 @@
...
+Gauchat-1925-480-447	447	The hunter	Le chasseur	captiatōre	148	1053	HUNTER
+Gauchat-1925-480-448	448	saw	a vu	vidūtu	148		
+Gauchat-1925-480-449	449	a hare	un lièvre	lĕpore	150	1190	HARE
+Gauchat-1925-480-450	450	(a) wolf	(un) loup	lŭpu	150	522	WOLF
+Gauchat-1925-480-451	451	(a) wild boar	(un) sanglier	singulare	150	3154	WILD BOAR
+Gauchat-1925-480-452	452	(an) eagle	(un) aigle	aquila	150	1905	EAGLE
+Gauchat-1925-480-453	453	a nice	un bel	bĕllu	150		
+Gauchat-1925-480-454	454	green bird	oiseau vert	aucĕllu vĭride	150		
+Gauchat-1925-480-455	455	The bird	L‘oiseau	aucĕllu	152	937	BIRD
+Gauchat-1925-480-456	456	makes	fait	facit	152
```
chrzyki: 734?

concepticondata/conceptlists/Gauchat-1925-480.tsv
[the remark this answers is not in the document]
LinguList: @Schweikhard, you are right in one sense, but it doesn't matter much, as the data is in such a clear form that we will be able to separate all the words, in fact, in our lexibank version, we already have them separated. But for the sake of consistency, we could leave it unmapped here and later see to which degree we can make another list of extracted concepticon concept sets, where articles etc. are deleted and we have the "raw words" with their occurrence and different forms (in fact, also something nice). In any case: given how specific this dataset is, there is no real "wrong" and "right" in this case.

LinguList:
Hi all, I am thinking, while we can discuss many cases, the best approach apart from those clearcut cases with article, or verbs in plain infinitive, is to leave many things unmapped, but also to leave the cases, unless they are plain wrong, for future discussion. I think, we'll modify the lexibank dataset soon in such a way that we can extract paradigms and the like where applicable from the data, and we will then make another list, in which we do a better mapping, as we know the true meanings in the context. 

For now, @Kristina-Pianykh, you can either leave it as is, or reduce it to all cases of "to VERB" and "the NOUN", so we have consistency, of what we leave unmapped, if this is fine with the others.

concepticondata/conceptlists/Gauchat-1925-480.tsv
LinguList: Yes, this should be modified.

concepticondata/conceptlists/Gauchat-1925-480.tsv
LinguList: It's the sentence "The pig is fat", so yes, it is not really "fat of food", but rather "fat (obese)". Obelix is also called "gros" in the comics.

concepticondata/conceptlists/Gauchat-1925-480.tsv
LinguList: Yes, true. Tiède means lukewarm, so this should be unmapped.

concepticondata/conceptlists/Gauchat-1925-480.tsv
LinguList: rather leave it unmapped. We tried to avoid to link to third singular.

LinguList:
@Kristina-Pianykh, I added three comments, where I follow the suggestions of others. these are my original linking errors. If you could correct these, I think we can link this.

concepticondata/conceptlists/Gauchat-1925-480.tsv
Schweikhard: There are more such already mapped, though: arrive, c'est,  il ferme, il balaire...

LinguList: Then, I'd say, they should be unmapped.

concepticondata/conceptlists/Gauchat-1925-480.tsv
LinguList: should be unmapped.

concepticondata/conceptlists/Gauchat-1925-480.tsv
LinguList: Yes, it can be mapped. However, I think now, we should not get too much into detail and later make an updated list where we map the concrete words in inter-linear glosses of this text. So whether we map "in the city" now, or not, is not that important.

concepticondata/conceptlists/Gauchat-1925-480.tsv
LinguList: Yes, @Kristina-Pianykh. I think now, however, we may have less problems with our critical reviewers, if we unmap all cases with "il VERB", and really only leave infinitives, "the NOUN", and "a NOUN". Sorry that I did not see this earlier.

concepticondata/conceptlists/Gauchat-1925-480.tsv
LinguList: "une heure" is in french the same as "l'heure", so if you isolate it, you have "hour", but if we stick to the strategy of mapping only the most clear-cut cases, we can unmap it. It would be mapped as "hour" in an updated list in which all words are isolated from the phrases.

LinguList:
Summarizing one more time. To avoid too many discussions on consistency, let's map it as this:

* the NOUN then a NOUN then NOUN (if this occurs)
* to VERB (with verb in isolation, no preposition)

In this way, we avoid all problems. Sorry again, @Kristina-Pianykh, for not seeing this more clearly earlier.

concepticondata/conceptlists/Gauchat-1925-480.tsv
Kristina-Pianykh: Alright, will do.

LinguList:
Many thanks. Please merge, when you think it is ready.

Issue #504, opened 2018-08-18
Fireplace, oven etc.


Schweikhard:
I think we might want to distinguish between a fireplace (i.e. a fire at the bottom of a chimney in a wall), a hearth (i.e. a place for cooking in the mid of a room), a stove (a fire or other kind of heater where things are put onto to heat them) and an oven (something into which put things to heat them). But I'm a bit at a loss which concept lists to link where, how exactly to write the definitions etc.

Currently we have
fireplace http://concepticon.clld.org/parameters/185
stove http://concepticon.clld.org/parameters/859
oven http://concepticon.clld.org/parameters/1143

AnnikaTjuka:
For me the mappings appear consistent. So I'm not sure what could be changed here.

Schweikhard:
It's basically about splitting FIREPLACE into two by adding HEARTH and possibly a hyperonym. The problem is that hearth is an ambiguous term, so it all might just mean FIREPLACE.

AnnikaTjuka:
I'd leave it as it is because the two concepts are too similar and most languages might not even have a clear-cut distinction.

Schweikhard:
I agree that it's not worth adding new concepts. It might be an idea to unmap Marrison who only writes "hearth", though.

AnnikaTjuka:
I'd be ok with FIREPLACE being a "catch-all" category of _hearth_ and _fireplace_ glosses.

AnnikaTjuka:
So, if we agree and no one else has a counterargument, I would suggest to close this issue.

Schweikhard:
I might at some point change the definition of FIREPLACE around a bit to make it more clear that it's catch-all - but yes, can be closed.

PR #779, opened 2020-03-04
RISE

### LINKED ISSUE #579: RISE
RISE (MOVE UPWARDS) currently refers to heavenly bodies rising from below the horizon.
It also contains some lists that instead refer to moving upwards in general (which is lacking a concept), or to a hyperonym of that and RISE (ONESELF).

Then there is also GO UP (which is basically used as to walk uphill or upstairs but is a hyperonym of CLIMB)
and
GO UP OR RISE (ONESELF)
both of which should be checked alongside.

Schweikhard:
fixed #579
and also changed the definition of FIREPLACE as discussed in #504.

Hope I understood 上去 correctly.
The 3 things I unmapped are referring to the rising of water and smoke/fog, respectively.

There are some things mapped to GO UP (ASCEND) that might only refer to walking up a hill and not to e.g. climbing a hill, but I didn't find clear cases.


concepticondata/conceptlists/Nagano-2013-1256.tsv
```diff
@@ -646,7 +646,7 @@ Nagano-2013-1256-644	644	1159	assemble (vi)
 Nagano-2013-1256-645	645	1160	meet, assemble (v)			
 Nagano-2013-1256-646	646	1162	bounce (v)			
 Nagano-2013-1256-647	647	1164	jump (v)	1829	JUMP	
-Nagano-2013-1256-648	648	1166	rise (v)/rise (上去) (v)	3207	GO UP OR RISE (ONESELF)	
+Nagano-2013-1256-648	648	1166	rise (v)/rise (上去) (v)	568	RISE (MOVE UPWARDS)
```
AnnikaTjuka: Not sure about that one. We have _Yang-2011-1014-713 上去 [chinese] and  Castro-2010-540-358 上去 [chinese]; to go up [english]_ mapped to GO UP (ASCEND).

MacyL:
In that case, maybe keep the GO UP (ASCEND) in all the Chinese datasets (Huang, Yang, and Castro), and expand the GO UP (ASCEND)'s definition?

Schweikhard:
Ah yes, I had intended to map Nagano to GO UP (ASCEND), just confused the concept names.

Is to fly really among the prototypical meanings of 上去? Then those could get their own new hyperonym.

What about the other Chinese glosses on GO UP (ASCEND)?
上(上山) and 上（山) are clearly about walking up or climbing a mountain or similar.
But there's also 上, 上（楼), 上（上坡), and 升起.

Schweikhard:
Or are there any lists among them were the English gloss is more reliable than the Chinese one?

LinguList:
Normally, the Chinese gloss is asked, I think.
null

CarolinHu:
Good job! Only one comment from me as well: given the examples currently in the concept sets (and the ones you unmapped), would it make sense to add a separate concept RISE for water levels, fog from a meadow, smoke from a  fireplace etc.?

Schweikhard:
> Good job! Only one comment from me as well: given the examples currently in the concept sets (and the ones you unmapped), would it make sense to add a separate concept RISE for water levels, fog from a meadow, smoke from a fireplace etc.?

There are only two for water levels, so that might not be enough - and I would say that's a different concept from the rising of vapours, and we only have one for that.

MacyL:
- 上(上山) \ 上（山) \  上（楼)) and ) are GO UP (ASCEND)

- if "sunrise (v)" exists in the concept list, then map 升起 to GO UP (ASCEND). If not, 升起 is RISE (MOVE UPWARDS). 

- 上 treat it cautiously by consulting other sources or unlink. 
> Ah yes, I had intended to map Nagano to GO UP (ASCEND), just confused the concept names.
> 
> Is to fly really among the prototypical meanings of 上去? Then those could get their own new hyperonym.
> 
> What about the other Chinese glosses on GO UP (ASCEND)?
> 上(上山) and 上（山) are clearly about walking up or climbing a mountain or similar.
> But there's also 上, 上（楼), 上（上坡), and 升起.

Schweikhard:
There is only 落(太阳~) | set (of the sun), so I mapped 升起 to RISE (MOVE UPWARDS).
I unmapped 上 (and 下) in Norman.

What about 上（上坡)?

LinguList:
For Norman, shang and xia are very basic terms, but it is correct that they are fuzzy throughout all Chinese dialects, with many meanings. 

However, in this specific case, we have to submit a paper tomorrow, and the data is based on Norman. So I'd prefer to not touch this list now, and I can also assure you that the go up vs. go down is the basic meaning intended in this case, even if the meaning may have changed in the individual dialects.

Schweikhard:
Okay, re-mapped those two. Well if the English glosses are reliable, these mappings are at least not too far off.

LinguList:
Correction: I just saw that Norman is providing English glosses. It is indeed not clear what is meant here, so I agree with unmapping this.

LinguList:
Sorry for not paying more attention earlier.

Schweikhard:
Well the English glosses at least show that those are verbs.
What's the plan now?

LinguList:
sorry again, you were too fast in answering ;)

leave it as is, with go up (ascend) and go down. I think this is the
best we have here. At least in many dialects, the words have these
meanings.

Schweikhard:
Okay, then everything is done as discussed now.

Only unclear case left is 上（上坡).

LinguList:
This [dictionary](http://www.handedict.de/chinesisch_deutsch.php?sid=a4a7f25b439f21e86549664820786028) is quite reliable in definitions. I think the normal reading is nominal, not verbal. [Linguee's](https://www.linguee.com/chinese-english/translation/%E4%B8%8A%E5%9D%A1.html) examples point also to nominal readings. 

But this does not mean that the 上(上坡) gloss is to be interpreted as a noun. It could as well say "the shang that this gloss talks about is the shang that means "to ascend", and not "on top", since 上 also has the nominal meaning "top" (of a table or similar).

LinguList:
So I'd also link it to "go up (ascend)".

Schweikhard:
Thanks - it's already mapped to that.

If there are no objections left I would merge this.

LinguList:
Yes, feel free to merge, and thanks!

Issue #716, opened 2020-01-23
Closing resolved issues


AnnikaTjuka:
I propose to close the following issues:

- [x] #50 
- [ ] #111 
- [x] #301 
- [x] #428 
- [ ] #436 
- [ ] #476 
- [ ] #580 
- [x] #618 
- [x] #681 
- [x] #679 

Some of them seem to be resolved by PR or I found a corresponding dataset in the conceptlists.tsv.

Note that I only skimmed them, so they might not be fully resolved. Please have a look at them and close them if appropriate.

chrzyki:
Thanks @AnnikaTjuka. Closed a couple where I was sure, left some for the others to check.

AnnikaTjuka:
Found some others:
#367 @Schweikhard or @tresoldi is this issue resolved?
#708 #680 @Kristina-Pianykh I think these are resolved with your latest PR :)

tresoldi:
@AnnikaTjuka it is not: https://github.com/concepticon/concepticon-data/blob/master/concepticondata/conceptlists/Buck-1949-1108.tsv

I think the discussion halted because we could add the missing concepts at the end of the list (as in my first suggestion, that @xrotwang accepted), but it would also mean changing the `1108` to `1110`.

AnnikaTjuka:
Thanks everyone for closing some of the issues (6 out of 10 :+1: )! I'll close this issue now because the other issues weren't yet solved. Maybe we should consider a "clean-up" sprint at some point, where we concentrate on solving the open issues.

PR #781, opened 2020-03-05
Buck, duals

### LINKED ISSUE #367: Missing concept in our Buck(et) List
A student (Iván Andrés-Alba) mentioned this: Concept RIB with Number 4.162 is missing in our list, but it occurs in Buck.

### LINKED ISSUE #396: Review grammatical concepts, especially pronouns
There are grammatical concepts, especially pronouns, missing; these are found in some concepts lists as linguists collect them both for typological research and for investigation of language evolution (assumed to be more stable and less subject to borrowing). One of the case is that of Papuan languages, as in Z'graggen's list; see https://github.com/clld/concepticon-data/pull/395 .

The concepts should be expanded to include properties such as different numbers (singular/dual/plural), inclusiveness (inclusive/exclusive), cases, etc. whenever possible (please note that many entries are already there). Other grammatical concepts, such as those related to negation and possession, could be included in this review.

Schweikhard:
fixes #367 
fixes #396


AnnikaTjuka:
Based on the comment by @tresoldi https://github.com/concepticon/concepticon-data/issues/716#issuecomment-580704201 the discussion of issue #367 was not finished yet.

Schweikhard:
> Based on the comment by @tresoldi #716 (comment) the discussion of issue #367 was not finished yet.

I understood his comment to mean that it was not fixed yet, not that the discussion wasn't finished.

> This also needs updates to retired.json to reflect that the Buck list now has a different name/ID.

Ah, didn't know, done.

concepticondata/conceptlists/Zgraggen-1980-380.tsv
```diff
@@ -308,40 +308,40 @@ Zgraggen-1980-380-374	2374	DREAM	374	dream
...
+Zgraggen-1980-380-389	3781	YOUR (MORE THAN TWO)	389	your-pl
 Zgraggen-1980-380-390	970	COCONUT	390	coconut
 Zgraggen-1980-380-391	2111	YOU TWO	391	you-dual
 Zgraggen-1980-380-392	2046	APRON	392	tanket
 Zgraggen-1980-380-393	732	YOUR (SINGULAR)	393	your-sg
 Zgraggen-1980-380-394	2642	HE OR SHE	394	he-she
-Zgraggen-1980-380-395	2400	THEIR	395	theirs
-Zgraggen-1980-380-396			396	your-pl-dual
+Zgraggen-1980-380-395			395	theirs
+Zgraggen-1980-380-396	2274	YOUR (PLURAL)	396	your-pl-dual
```
Kristina-Pianykh: I'm not sure if this mapping is precise enough. YOUR (PLURAL) does not specify the number of persons addressed which is crucial in this case.

concepticondata/conceptlists/Zgraggen-1980-380.tsv
```diff
@@ -308,40 +308,40 @@ Zgraggen-1980-380-374	2374	DREAM	374	dream
...
 Zgraggen-1980-380-400	31	PAN	400	pan
-Zgraggen-1980-380-401			401	you-o2s
-Zgraggen-1980-380-402			402	we-dual
-Zgraggen-1980-380-403	780	OUR	403	our
+Zgraggen-1980-380-401	2304	THEE (OBLIQUE CASE OF YOU)	401	you-o2s
+Zgraggen-1980-380-402	2634	WE TWO	402	we-dual
+Zgraggen-1980-380-403	3779	OUR (MORE THAN TWO)	403	our
 Zgraggen-1980-380-404	984	WORK (LABOUR)	404	work-n
-Zgraggen-1980-380-405			405	you-pl
+Zgraggen-1980-380-405	3777	YOU (MORE THAN TWO OBLIQUE CASE)	405	you-pl
```
Kristina-Pianykh: YOU (MORE THAN TWO OBLIQUE CASE) is already mapped to you-pl-o2p in this list.

concepticondata/conceptlists/Zgraggen-1980-380.tsv
Schweikhard: That mapping is intentional - they also have your-dual, which clearly only refers to dual, so this gloss here seems to imply that it can refer to both dual and plural, which is what YOUR (PLURAL) means.

concepticondata/conceptlists/Zgraggen-1980-380.tsv
Schweikhard: Hm, and it cannot be subject case as there is already _ye_ in the list. I don't know what they mean here.

Schweikhard:
> etc. This can be automatically generated with concepticon rename. The URL in the the metadata file also needs to be updated to the new name.

So I basically undo the renaming changes I did and then run `concepticon rename Buck-1949-1108 Buck-1949-1110`?

because that gives me `
INFO    concepticon/concepticon-data at C:\Users\n_sch\AppData\Local\cldf\cldf\concepticon
Traceback (most recent call last):
  File "c:\program files\python38\lib\runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\program files\python38\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\Users\n_sch\Lexibank\env\Scripts\concepticon.exe\__main__.py", line 7, in <module>
  File "c:\users\n_sch\lexibank\env\lib\site-packages\pyconcepticon\__main__.py", line 70, in main
    return args.main(args) or 0
  File "c:\users\n_sch\lexibank\env\lib\site-packages\pyconcepticon\commands\rename.py", line 51, in run
    assert oid.startswith(args.from_)
AssertionError
`

chrzyki:
Yes, if you undo the changes and then run `concepticon rename` this should take care of things automatically.

Schweikhard:
Ah, there was one change I forgot to undo. Now it worked.

Schweikhard:
Um, the newly automatically created url in the metadata is Buck-1949-1110.tsv-metadata.json - shouldn't it be just Buck-1949-1110.tsv ?

concepticondata/conceptlists/Zgraggen-1980-380.tsv
Schweikhard: One solution would be that _ye_ is YOU (dual or more than 2), you-pl is YOU (MORE THAN TWO) and you-pl-o2p is YOU (MORE THAN TWO OBLIQUE CASE). I changed it to that.

Schweikhard: That makes it inconsistent with their _we_ and _they_, though, which I mapped as more than 2 since they are only contrasted with duals.

chrzyki:
Yes, this is a minor bug in `concepticon rename` ([here](https://github.com/concepticon/pyconcepticon/blob/5886de566efde28ac98d7f56d9a81c78146130b1/src/pyconcepticon/commands/rename.py#L59-L64)). Fix this manually and I'll prepare a PR for `pyconcepticon`.

concepticondata/conceptlists/Zgraggen-1980-380.tsv
Schweikhard: Hm, I looked into the actual book now, and it very much seems to me that the elicitation glosses we have here are not the same as in the book. I don't find most of the pronoun glosses there, only ye, we, thou, they, I, he. Instead it contains, _after the concept list_, paradigms distinguishing between free, possessive, genitive, verbal object, and either between singular, dual and plural (for all 3 persons) or between singular and plural (for all 3 persons) depending on the language.
So I have no idea where these glosses even come from.

Kristina-Pianykh: That was my impression too when I looked at the source. This might be the case that the list on concepticon was taken from elsewhere rather than being typed directly off the source

concepticondata/conceptlists/Zgraggen-1980-380.tsv
Schweikhard: I would propose correcting the concept list and then correcting the mapping.
The data as in the book at least clearly exists, whereas I don't know whether the glosses we have here were ever actually used.

Btw, we could nevertheless retain the new concepts I now added as we might need them in the future.

Schweikhard:
I've changed the content of Zgraggen to what is actually provided in the book. I left the original enumeration (which seems to go back to an unpublished concept list by the author) in a new column ORIGINAL_NUMBER but removed all concepts not provided in the book.

Schweikhard:
Some concepts also needed renaming (e.g. fishing spear to spear) and I remapped them accordingly.

LinguList:
Can you please have a look at the lexibank file of Zgraggen, at
https://github.com/lexibank/zgraggenmadang , and see to which degree the
concept list is important here? What you should now is: since Zgraggen
is important for CLICS, etc., any change on the concept list will
require extended reviews, and removal of data points may lead to a
failure in lexibank. So this is a case for a review that requires more
people to check it, and I recommend that @SimonGreenhill also has a look
at this case, and that @chrzyki checks if this breaks the lexibank dataset.

concepticondata/concepticon.tsv
```diff
@@ -3753,3 +3753,17 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3767	RICE SEEDLING	Agriculture and vegetation	The seedling of a rice plant.	Person/Thing	
 3768	POUND (FOOD)	Motion	To prepare food with a pounding or beating movement, for example, pre-processing of grain or of sago.	Action/Process	
 3769	GRAPEVINE	Agriculture and vegetation	A climbing plant that produces grapes and belongs to the family Vitaceae.	Person/Thing	
+3770	YOU TWO (OBLIQUE CASE)	Kinship	Oblique cases of the second person dual pronoun.	Other	
+3771	THEM TWO	Kinship	Oblique cases of the third person dual pronoun.	Other	
+3772	THEIR (DUAL)	Kinship	Belonging to the third person dual.	Other	
+3773	YOU (MORE THAN TWO)	Kinship	Second person plural excluding dual pronoun.	Other	
+3774	THEM (MORE THAN TWO)	Kinship	Oblique cases of the third person plural excluding dual pronoun.	Other
```
CarolinHu: as you did with 3777, I would add OBLIQUE CASE

Schweikhard:
The lexibank must be based on a different concept list - Zgraggen1980 only contains e.g. coconut, not coconut tree, but the lexibank contains both. And the lexibank also contains a different set of languages.

concepticondata/concepticon.tsv
```diff
@@ -3753,3 +3753,17 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3767	RICE SEEDLING	Agriculture and vegetation	The seedling of a rice plant.	Person/Thing	
 3768	POUND (FOOD)	Motion	To prepare food with a pounding or beating movement, for example, pre-processing of grain or of sago.	Action/Process	
 3769	GRAPEVINE	Agriculture and vegetation	A climbing plant that produces grapes and belongs to the family Vitaceae.	Person/Thing	
+3770	YOU TWO (OBLIQUE CASE)	Kinship	Oblique cases of the second person dual pronoun.	Other	
+3771	THEM TWO	Kinship	Oblique cases of the third person dual pronoun.	Other	
+3772	THEIR (DUAL)	Kinship	Belonging to the third person dual.	Other	
+3773	YOU (MORE THAN TWO)	Kinship	Second person plural excluding dual pronoun.	Other	
+3774	THEM (MORE THAN TWO)	Kinship	Oblique cases of the third person plural excluding dual pronoun.	Other	
+3775	US TWO	Kinship	Oblique cases of the first person dual pronoun.	Other	
+3776	HIM OR HER OR IT	Kinship	Oblique cases of the third person singular pronoun.	Other
```
CarolinHu: as you did with 3777, I would add OBLIQUE CASE
Also, I would go for HE OR SHE OR IT (OBLIQUE CASE) because of 266 (the non-oblique counterpart)

Schweikhard:
Zgraggen1980 contains e.g. the languages Kolom and Suroi which are not in the lexibank.

Schweikhard:
(the lexibank contains Kolom but under the name migum)

LinguList:
But Zgraggen is listed as explicit SOURCE for the lexibank data. In
fact, we are dealing with 4 books here. The lexibank data is based on a
re-mapping of the datasets. I think, we should leave this list untouched
for Version 2.3, and make an explicit new mapping for version 2.4 or
after that. I already filed an issue on the lexibank zgraggen data:
https://github.com/lexibank/zgraggenmadang/issues/5

People involved here are @SimonGreenhill, who provided the data (merged
from four books), @tresoldi as maintainer, and @chrzyki as person
involved with clics and lexibank updates.

concepticondata/concepticon.tsv
```diff
@@ -3753,3 +3753,17 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+3773	YOU (MORE THAN TWO)	Kinship	Second person plural excluding dual pronoun.	Other	
+3774	THEM (MORE THAN TWO)	Kinship	Oblique cases of the third person plural excluding dual pronoun.	Other	
+3775	US TWO	Kinship	Oblique cases of the first person dual pronoun.	Other	
+3776	HIM OR HER OR IT	Kinship	Oblique cases of the third person singular pronoun.	Other	
+3777	YOU (MORE THAN TWO OBLIQUE CASE)	Kinship	Oblique cases of the second person plural excluding dual pronoun.	Other	
+3778	OUR (DUAL)	Kinship	Belonging to the first person dual.	Other	
+3779	OUR (MORE THAN TWO)	Kinship	Belonging to the first person plural excluding dual.	Other	
+3780	THEIR (MORE THAN TWO)	Kinship	Belonging to the third person plural excluding dual.	Other	
+3781	YOUR (MORE THAN TWO)	Kinship	Belonging to the second person plural excluding dual.	Other	
+3782	WE (MORE THAN TWO)	Kinship	First person plural excluding dual pronoun.	Other
```
CarolinHu: I'm not sure whether those will be relevant, but what about the combination of dual/plural, case and inclusiveness all in one concept, as Tiago mentioned in #396 ? 
E.g. WE (MORE THAN TWO, INCLUSIVE, OBLIQUE)

Schweikhard:
I didn't expect this to be a major change when I started this PR.
I can make a new PR with only the Buck changes and we take of this here later.

LinguList:
I only now realized that this touches upon lexibank, which is both
something that was NOT done in the lexibank dataset (where the concept
remappings were never questioned, although they should've been), and
something more general: if we modify concepticon, we need to re-release
lexibank data. So any remapping should be done by having this in mind as
well. Changes in concepticon have consequences in other places, so at
times, it is better to tolerate a problematic mapping than to force a
dramatic change on the other side.

But this is something we could not foresee, it is something we have to
foresee in the future, however, so we learn a nice lesson here from this
issue.

Schweikhard:
Thanks for the comments @CarolinHu !

@LinguList Since the new concepts I introduced are not needed atm but might be necessary in the future, shall I include them when I make a PR with only the Buck changes now?

LinguList:
well, generally not, if they are needed for a future list, as they won't
appear there to be linnked to anything. But if it's too much work, just
leave them. We want to finish, so we can release the data anew.

Schweikhard:
More work to include them ;) So I'll close this one for the time being.

Issue #787, opened 2020-03-09
Davletshin-2012-100: Concept list in Aztecan languages


LinguList:
> Davletshin, Albert (2012): Proto-Uto-Aztecans on their way to the Proto-Aztecan homeland: linguistic evidence. Journal of Language Relationship. 8. 1. 75-92.

It is an open access journal.

LinguList:
[Davletshin-2012-100.txt](https://github.com/concepticon/concepticon-data/files/4305444/Davletshin-2012-100.txt)

LinguList:
[Davletshin2012.pdf](https://github.com/concepticon/concepticon-data/files/4305447/Davletshin2012.pdf)

LinguList:
I typed off the concept list from the article. The PDF should be submitted along with the references, etc. this time.

PR #792, opened 2020-03-12
Add Davletshin-2012-100

CarolinHu:
# Pull request checklist

- [x] add new concept list

## Additional information

List from issue #787 and Lexibank: https://github.com/lexibank/lexibank/issues/206
Pinging @tresoldi so you know when this is done.

This is in fact the basic 100-item Swadesh list, so I went for the same concept mappings except with burn because the author specified the transitive meaning in this list.


LinguList:
Okay, @chrzyki, could you moderate this PR?

LinguList:
And, @chrzyki, I volunteer to review this one.

tresoldi:
Thank you, @CarolinHu ! I'll wait for the review, but everything seems correct.

chrzyki:
@LinguList thanks & will do.

chrzyki:
@AnnikaTjuka, @Kristina-Pianykh and @LinguList I've assigned you as reviewers for this PR and I'll check your reviews accordingly.

@CarolinHu in the future I'd say leave in the whole PR template (even the 'empty') boxes, makes it more apparent what was *not* done and I don't have to check the template.

concepticondata/conceptlists.tsv
```diff
@@ -309,3 +309,4 @@ Zettersten-2020-24a	Zettersten, Martin and Lupyan, Gary	2020	a	24	naming test, c
...
+Davletshin-2012-100	Davletshin, Albert	2012		100	basic, areal, proto-language	English	Uto-Aztecan languages	http://www.jolr.ru/files/(95)jlr2012-8(75-92).pdf	Davletshin2012		The concepts of this list were taken from [Swadesh's 100-item list](:ref:Swadesh-1955-100) and formed the basis for comparing 9 different Uto-Aztec languages. Based on this comparison, the author proposed possible corresponding forms of Proto-Aztecan.
```
Kristina-Pianykh: Maybe paraphrase the first sentence a bit? I suggest: `The concepts for this list were taken from [Swadesh's 100-item list](:ref:Swadesh-1955-100), which formed the basis for comparing 9 different Uto-Aztecan languages.`

It's still not clear to me which lists are to be tagged as 'areal'. If this tag describes lists that were used for specific languages/a specific geographic area, then almost every list on Concepticon (with some minor exceptions) should be tagged as such. I was convinced this tag is used for those lists only that differ from basic ones and include a significant share of 'areal' concepts (e.g., certain plants or cultural artifact, typical of this area/languages).

concepticondata/conceptlists/Davletshin-2012-100.tsv
```diff
@@ -0,0 +1,101 @@
...
+Davletshin-2012-100-68	68	root 	670	ROOT
+Davletshin-2012-100-69	69	round 	1395	ROUND
+Davletshin-2012-100-70	70	sand 	671	SAND
+Davletshin-2012-100-71	71	say 	1458	SAY
+Davletshin-2012-100-72	72	see 	1409	SEE
+Davletshin-2012-100-73	73	seed 	714	SEED
+Davletshin-2012-100-74	74	sit 	1416	SIT
+Davletshin-2012-100-75	75	skin 	763	SKIN
+Davletshin-2012-100-76	76	sleep 	1585	SLEEP
+Davletshin-2012-100-77	77	small 	1246	SMALL
```
Kristina-Pianykh: I didn't find 'small' in the PDF list.

LinguList:
@Kristina-Pianykh, the list is not areal, since it was explicitly taken
from Swadesh's list, which is not areal. Areal lists are shorter or
longer lists with explicitly areal items, like, e.g., "banana", and
where authors explicitly mentioned that they adjusted the list to study
a specific language family or languages in a specific area.

CarolinHu:
Thanks for all of your comments! Working on the changes. Two things are not clear to me yet:

- I used the version of the list from https://github.com/lexibank/davletshinaztecan/blob/master/etc/concepts.tsv because I was assigned to the issue over on Lexibank and because I thought problems occur when the lists in Concepticon and Lexibank differ. If I delete "small" now ("smoke" is in the list, at least if we talk about the PDF linked in issue #787), won't that cause problems in the Lexibank set?

- How exactly do I add PDFs to /sources? I tried doing that via the normal GitHub procedure (see document at the very end), but this seems to not be the right way.

LinguList:
@CarolinHu, you did completely right, but the original Proto-Aztecan forms should not have been deleted, since they are part of the lexibank workflow as well. @tresoldi, did you delete the ProtoAztecan forms in the lexibank script? If so, can you please provide an updated list of them here?

As to the 99 vs. 100 forms, the list you have is correct, I just want you to add in the NOTE on `conceptlists.tsv` that one concept is missing in the list on page X, while the data for all langauges on pages A-B (please check page numbers with the original) show this concept clearly. 

As to adding a PDF: you just add it in the folder `sources` as `Davletshin2012.pdf`, and then add `Davletshin2012` to the `conceptlists.tsv` where it says "PDF".

LinguList:
@tresoldi, never mind, I just saw that they are all there. So @CarolinHu, you just have to copy the PROTO_AZTECAN back into the tsv-file.

CarolinHu:
> As to the 99 vs. 100 forms, the list you have is correct, I just want you to add in the NOTE on `conceptlists.tsv` that one concept is missing in the list on page X, while the data for all langauges on pages A-B (please check page numbers with the original) show this concept clearly.
Ah, now I get it, thank you! I deleted the proto forms because I thought it was enough to have them in Lexibank, but I'll insert them again. Sorry about this.

LinguList:
no problem, it was maybe not specific enough from my side. But you see: we will soon delete the list on lexibank and take it directly from concepticon, so the forms need to be there, as lexibank will no longer use the `concepts.tsv`. For this reason, it is essential to keep them.

AnnikaTjuka:
Regarding the adding of PDFs: Unfortunately, this is not so trivial because of the cdstar server. I talked to @chrzyki recently about this and he prefers that he adds the PDFs to the server.

So @CarolinHu , you only have to add the Bibtex key in the column "PDF" in conceptlists.tsv and send the PDF to @chrzyki . You don't need to save it in the folder. This will produce a warning when you type `concepticon test` but you can ignore that.

LinguList:
Yes. Our practice in the past was pushing them as pdf, and then
@xrotwang would put them on to the CDSTAR server, but if this practice
has changed by now, it is useful that we have a precedence case now
where we can discuss it, as I find we can offer more PDF snippets, if
the datasets are trivial and small.

If this practice of sending a PDF to @chrzyki or @xrotwang and asking
them to place them on the cdstar server is not yet in our contribution
guidelines, it would be nice if one of you could make an issue on this.

AnnikaTjuka:
Sure, I'll open an issue.

chrzyki:
> 
> Yes. Our practice in the past was pushing them as pdf, and then @xrotwang would put them on to the CDSTAR server, but if this practice has changed by now, it is useful that we have a precedence case now where we can discuss it, as I find we can offer more PDF snippets, if the datasets are trivial and small. If this practice of sending a PDF to @chrzyki or @xrotwang and asking them to place them on the cdstar server is not yet in our contribution guidelines, it would be nice if one of you could make an issue on this.

Yes, good to talk about this. I think its best if PDFs aren't pushed to the repository since they can get quite heavy and bloat the repository. Happy to do this whenever it's needed.

CarolinHu:
Thank all of you again for your comments! Unfortunately, I have a train to catch now. If everything is fine after the reviewers had another look, perhaps someone else could merge this for me? If there are still things to correct, I will gladly take care of them on Tuesday. Have a nice weekend!

LinguList:
The moderator will merge this. Thanks! 
null

chrzyki:
Looks good. Thanks for the contributions, all! Will take care of the PDF.

chrzyki:
Resolved the conflicts & waiting for a 'positive' review to merge.

concepticondata/conceptlists.tsv
```diff
@@ -309,4 +309,5 @@ Zettersten-2020-24a	Zettersten, Martin and Lupyan, Gary	2020	a	24	naming test, c
...
+Davletshin-2012-100	Davletshin, Albert	2012		100	basic, proto-language	English	Uto-Aztecan languages	https://doi.org/10.31826/9781463235406-005	Davletshin2012	Davletshin2012	The concepts for this list were taken from [Swadesh's 100-item list](:ref:Swadesh-1955-100) and formed the basis for comparing 9 different Uto-Aztecan languages. Based on this comparison, the author proposed possible corresponding forms of Proto-Aztecan. The concept “small” is actually missing from the proto-word list but was included in the Aztecan word list on pp. 84-90.	91
```
AnnikaTjuka: Is it only one page number? Or shouldn't it be "75-92" instead of 91?

LinguList:
Only one is correct for me.
null

LinguList:
Just a brief explanation, @AnnikaTjuka: the page number refers usually
to where you find the table. You could give the whole page range, but
one would then have to provide two glosses, as the glosses are all
capitalized there, etc., and since the table on that single page is in
some sense the core of the endeavor in the paper, I find it sufficient
to say that we provide only this, also since we provide only the PDF
with this page, not the full etymological records with cognates.

PR #803, opened 2020-03-25
Update uppercase, page numbers, relation

AnnikaTjuka:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [x] refine Concepticon concept relations
- [ ] retire data

## Additional information

I did the following changes:
- Uppercase: I tried to unify the descriptions of the concepts so that all descriptions start with a capital letter. I also corrected a few spellings and added a description for PULL UP (RICE SEEDLING).
- Page numbers: For two norm data sets, I included the pages of the article but the lists occur in the additional supplementary material. Thus, I deleted the numbers in the column "PAGE". In case of Zettersten-2020, I added the page numbers for the lists which appear within the article.
- Relation: I added a concept relation for VINE and GRAPEVINE.


AnnikaTjuka:
@chrzyki Would you mind moderating this PR?

chrzyki:
> 
> 
> @chrzyki Would you mind moderating this PR?

With pleasure. @Schweikhard, @Kristina-Pianykh, and @CarolinHu, can you have a look at this please as reviewers?

concepticondata/concepticon.tsv
```diff
@@ -963,7 +963,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 962	WOMAN	Kinship	An adult human member of the sex that produces ova and bears young.	Person/Thing	
 963	BOOK	Speech and language	A collection of leaves of paper, parchment or other material, usually bound or fastened together within covers, containing writing of any type or blank pages for future inscription.	Person/Thing	
 964	WOOL	Clothing and grooming	The soft, curly hair that forms the fleece of sheep, lamas and other animals and which is used to produce clothing.	Person/Thing	
-965	WORLD	The physical world	The Earth with all its inhabitants and all things upon it.	Person/Thing	
+965	WORLD	The physical world	The earth with all its inhabitants and all things upon it.	Person/Thing
```
CarolinHu: I would leave Earth capitalized here as it refers to the name of the planet.

concepticondata/concepticon.tsv
```diff
@@ -1128,8 +1128,8 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1127	SEND	Motion	To make something (an object, a message) go from one place to another.	Action/Process	
 1128	ORDER	Basic actions and technology	To order someone or several persons by a superior authority to do something, or to arrange something in a tidy manner.	Action/Process	
 1129	PAINFUL	Sense perception	Full of pain; causing uneasiness or distress, either physical or mental.	Property	
-1130	WE (EXCLUSIVE)	Kinship	exclusive “we”: I and at least one other person, but not the addressed person (you) (pronoun 1st person plural without 2nd person)	Other	
-1131	WE (INCLUSIVE)	Kinship	inclusive “we”: I and at least one other person, including the addressed person(s) (pronoun 1st person plural with 2nd person)	Other	
+1130	WE (EXCLUSIVE)	Kinship	Exclusive “we”: I and at least one other person, but not the addressed person (you) (pronoun 1st person plural without 2nd person)	Other	
+1131	WE (INCLUSIVE)	Kinship	Inclusive “we”: I and at least one other person, including the addressed person(s) (pronoun 1st person plural with 2nd person)	Other
```
CarolinHu: Perhaps you could modify WE (EXCLUSIVE) so that it also says "but not the addressed person(s)" (instead of just one person)?

concepticondata/concepticon.tsv
```diff
@@ -1150,12 +1150,12 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1149	ORPHAN	Kinship	A person one or both of whose parents have died, especially a minor.	Person/Thing	
 1150	THUNDER	The physical world	Sound produced by fast air expansion induced by a lightning.	Person/Thing	
 1151	VEIL	Clothing and grooming	A head covering.	Person/Thing	
-1152	BEGET	The body	(especially of a male parent) to procreate or generate.	Action/Process	
+1152	BEGET	The body	(Especially of a male parent) to procreate or generate.	Action/Process
```
CarolinHu: I would capitalize the "to" after the parentheses as well.

concepticondata/concepticon.tsv
```diff
@@ -1717,7 +1717,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 1716	FORTY	Quantity	The natural number forty (40).	Number	
 1717	FIFTY	Quantity	The natural number fifty (50).	Number	
 1718	SIXTY	Quantity	The natural number sixty (60).	Number	
-1719	BURY	The body	to put in the ground and cover with earth.	Action/Process	
+1719	BURY	The body	To put in the ground and cover with earth.	Action/Process
```
CarolinHu: Question about this concept in general (I am not sure about the extent of the words already mapped in Concepticon): is it only "bury" as in "bury dead people"? Because if yes, the description could be narrowed down to avoid confusion/uncertainty, "to put a dead body in the ground". If no, and e.g. a buried treasure would fit into this concept as well, "The body" might not be the appropriate semantic category.

concepticondata/concepticon.tsv
AnnikaTjuka: Ah, I see.

concepticondata/concepticon.tsv
AnnikaTjuka: I think WE (EXCLUSIVE) is mainly used if one talks to another person about something that she is doing without the addressed person (singular). For example, "we go to the concert (but not you)".

concepticondata/concepticon.tsv
AnnikaTjuka: All the other cases with this sequence are lowercase: "(....) to".

concepticondata/concepticon.tsv
AnnikaTjuka: Based on the mappings (see https://concepticon.clld.org/parameters/1719) it is not specific for "bury people". We could make a "narrower" concept but that should be discussed in an issue first.

AnnikaTjuka:
Thanks for the remarks. I changed "earth" back to "Earth" and left comments for the other suggestions.

concepticondata/concepticon.tsv
CarolinHu: I thought it works analog to WE (INCLUSIVE) which is why I argue for equal definitions. In both cases, you may be talking to more than one person and include or exclude that group which is being spoken or referred to. 
Then you'd have "We go to the concert (but not you, Dirk and Linda, to which I am currently talking)" and "We go to the concert (including you, Dirk and Linda, to which I am currently talking)".
It's not a big thing at all and we can leave it as it is, just wanted to explain where I'm coming from.

concepticondata/concepticon.tsv
CarolinHu: It's fine if the concept refers to the broader sense. I would just modify the semantic field in that case because "The body" does not really fit if it's not solely about burying corpses. Perhaps "Basic Actions and Technology"?

concepticondata/concepticon.tsv
AnnikaTjuka: Sure, I get the idea. The tricky part is that "you" is also under-specified (see https://en.wikipedia.org/wiki/Clusivity).

AnnikaTjuka:
I changed the semantic field of BURY. If there are no objections, I'd leave the definition of WE (EXCLUSIVE) as it is.

concepticondata/concepticon.tsv
```diff
@@ -2064,7 +2064,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2063	CLOSE (EYES)	Basic actions and technology	Move ones eyelids in such a way that the eyes are no longer visible.	Action/Process	
 2064	SCATTER (SEEDS)	Basic actions and technology	To distribute seeds over a field.	Action/Process	
 2065	STEP (VERB)	Basic actions and technology	To move the foot in walking; to advance or recede by raising and moving one of the feet to another resting place, or by moving both feet in succession.	Action/Process	
-2066	PULL UP (RICE SEEDLINGS)	Basic actions and technology		Action/Process	
+2066	PULL UP (RICE SEEDLINGS)	Basic actions and technology	To remove the roots of a rice plant.	Action/Process
```
Schweikhard: That sounds strange, isn't it rather something like "to pull a rice seedling out of the ground"?
I remember @MacyL and me talked about that concept before but I don't remember the details. Only thing mapped to it is 拔（秧）.

concepticondata/concepticon.tsv
```diff
@@ -2078,19 +2078,19 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+2085	STICK (CLASSIFIER)	Quantity	Classifier for stick-like objects	Classifier	
+2086	GRAIN (CLASSIFIER)	Quantity	Classifier for tiny objects, like rice	Classifier	
+2087	DROP (CLASSIFIER)	Quantity	Classifier for small units of liquid, like water	Classifier	
+2088	CLOTHING (CLASSIFIER)	Quantity	Classifier for things that one wears	Classifier	
+2089	PAIR (CLASSIFIER)	Quantity	Classifier for objects that occur in pairs	Classifier	
 2090	MY	Possession	Belonging to me.	Other	
-2091	YOUR	Possession	belonging to the second person that is addressed by the first person	Other	732
-2092	HIS OR HER	Possession	belonging to the third person that is addressed	Other	
+2091	YOUR	Possession	Belonging to the second person that is addressed by the first person	Other	732
+2092	HIS OR HER	Possession	Belonging to the third person that is addressed	Other
```
Schweikhard: Shouldn't these two and the classifiers also end in a dot?

concepticondata/concepticon.tsv
```diff
@@ -2346,7 +2346,7 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2345	LEPORID (RABBIT OR HARE)	Animals	Mammal of the family hares and rabbits (Leporidae) with long ears, short tail and hindlegs which are shorter than the forelegs and permit running quickly.	Person/Thing	
 2346	CRESTED MYNA	Animals	A species of starling native to southeastern China and Indochina.	Person/Thing	
 2347	YELLOW CROAKER	Animals	A species of croaker native to the western Pacific, generally in temperate waters such as the East China Sea and the Yellow Sea.	Person/Thing	
-2348	CRAB	Animals	"Decapod crustaceans of the infraorder Brachyura, which typically have a very short projecting ""tail"" (abdomen) (Greek: βραχύς / brachys = short, οὐρά / οura = tail), usually entirely hidden under the thorax."	Person/Thing	
+2348	CRAB	Animals	Decapod crustaceans of the infraorder Brachyura, which typically have a very short projecting "tail" (abdomen) (Greek: βραχύς / brachys = short, οὐρά / οura = tail), usually entirely hidden under the thorax.	Person/Thing
```
Schweikhard: Office is going to mess up that one every time - can't we just escape the quotes?

concepticondata/concepticon.tsv
AnnikaTjuka: I used the definition of [pull up](https://www.collinsdictionary.com/dictionary/english/pull-up) in the Collins dictionary as a basis.

Schweikhard: "to remove by the roots"? Ah, now I see how you mean your definition.
"To remove the roots of a rice plant." sounds to me like "To remove the roots from a rice plant, and doing whatever with the remainder.", but you mean "To remove a rice plant including the roots.".

Rice seedlings are nothing usually just discarded, though, so I would guess this is rather about harvesting them for food or replanting them where there is more space. @MacyL could you help?

concepticondata/concepticon.tsv
AnnikaTjuka: I found some others where "dots" are missing and added them.

concepticondata/concepticon.tsv
AnnikaTjuka: I'll open a separate issue to discuss the "quotes". Because some descriptions include "&quote".

concepticondata/concepticon.tsv
AnnikaTjuka: Suggestion: "To remove a rice plant including its roots (for harvesting)."

AnnikaTjuka:
I added some missing dots and unified the quotes.

AnnikaTjuka:
If @MacyL and @Schweikhard agree with the following definition, I'll add it as well:

拔（秧）PULL UP (RICE SEEDLING) "To remove a rice plant including its roots (for harvesting)."

Please also check the mappings for PULL UP WEEDS (https://concepticon.clld.org/parameters/3548)

concepticondata/concepticon.tsv
MacyL: Here is my thought : remove PULL UP (RICE SEEDLINGS), and move Allen-2007-500-296 to PULL UP WEEDS.
Reason : 
1. Only one concept, 拔（秧）(Allen-2007-500-296) map to this concepticon concept.
2. The form "ma²¹" (Allenbai, Jianchuan language variety) is equal to the form "ma21" in (Wangbai, doculact DS). And Wang-2004-471-311 pull up (weeds) maps to PULL UP WEEDS (along with many others)
3. In many concept list with Chinese glosses, the object in the brackets is an example. So we should focus on defining the action 拔  "pull up". 
4. When harvesting rice (稻米）, Chinese don't pull up the rice plants （稻株）, the farmers cut the plants instead of pull up. The method varies between countries.
5. 拔秧 (not brackets are used here), if there is a concept like this,  it means pull up rice seedlings including the roots (as Annika suggested) . But harvesting is not the main reason, it's for spacing rice seedlings, so that they have room to grow (and subsequently, maximising the quantity of rice).

Schweikhard: Thanks! Then I'd suggest we remap Allen to PULL UP WEEDS.
We could either retire this concept here, or we leave it open for future mappings and change the definition to something like "To pull out rice seedlings from the soil including their roots in order to separate seedlings growing too closely.".

AnnikaTjuka:
I changed the mapping of Allen-2007 and added the description. But I think we should retire PULL UP (RICE SEEDLING) because it seems rather specific and PULL UP WEEDS seems to be sufficient enough.

@chrzyki How can I retire a concept? The retired.json seems to list only retired conceptlist IDs.

Schweikhard:
I agree, if we get such a new concept, it's likely to be just PULL UP (SEEDLING) or such like.

I think for retiring a concept you just need to put the ID of the concept which it replaces into the last column of concepticon.tsv. It only works if the replacing ID is lower, though.

chrzyki:
> 
> 
> I changed the mapping of Allen-2007 and added the description. But I think we should retire PULL UP (RICE SEEDLING) because it seems rather specific and PULL UP WEEDS seems to be sufficient enough.
> 
> @chrzyki How can I retire a concept? The retired.json seems to list only retired conceptlist IDs.

The final column in https://github.com/concepticon/concepticon-data/blob/master/concepticondata/concepticon.tsv (`REPLACEMENT_ID`) allows replacing concepts with new ones.

Schweikhard:
Which isn't the case here.

chrzyki:
See for example how 264 HE-GOAT was replaced by 2475 MALE GOAT.

Schweikhard:
Ah, so the ID numbers don't matter anymore? Good to know!

AnnikaTjuka:
I added the ID of PULL UP WEEDS in the "Replacement" column of PULL UP (RICE SEEDLINGS) so that all suggestions were edited accordingly to the reviews.

concepticondata/concepticon.tsv
```diff
@@ -2338,22 +2338,22 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 2337	OF THAT KIND (SUCH)	Miscellaneous function words	In that way, such, etc.	Other	
 2338	HOW LONG	Time	Asking for the duration of an event.	Other	
 2339	NEED (SOMETHING)	Cognition	Feeling the desire or the urge for something.	Other	
-2340	SUNNY	The physical world	Full of sunshine (mostly referring to &quot;a sunny day&quot;).	Property	
-2341	CLOUDY	The physical world	Full of clouds (mostly referring to &quot;a cloudy day&quot;).	Property	
+2340	SUNNY	The physical world	“Full of sunshine (mostly referring to “a sunny day”).”	Property	
+2341	CLOUDY	The physical world	“Full of clouds (mostly referring to “a cloudy day”).”	Property
```
Schweikhard: For crab you use "instead of “ for the inner quotes. And from what was discussed in the issue, I thought it's supposed to be "x" around, and ""x"" within?

AnnikaTjuka: Yes, your right. I changed the quotes accordingly.

AnnikaTjuka:
I found a few more quotes with the "LibreOffice" type of quotes (**“**) and changed them. Now all descriptions should have the same quotation style.

chrzyki:
Looks like this can be merged! Thanks for the many fixes and the helpful hands of the reviewers.

PR #805, opened 2020-03-30
Add trailing zeroes to SRCIDs

chrzyki:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This adds back trailing zeroes which carry semantics for the STEDT lookup and made the Lexibank fail. See also this issue for the dataset: https://github.com/lexibank/halenepal/issues/14.


LinguList:
Thanks! @AnnikaTjuka, @Schweikhard, @Kristina-Pianykh, if you look at
the modifications, you can see that this happened involuntarily in the
list during a PR, because when using LibreOffice to open a text file,
things like zeros at the end of a floating number are often
automatically deleted. Nobody spotted this, so our review process wasn't
attentive, but it caused the dataset to fail being created in lexibank.
What we can learn from this is to pay also attention to these small
details when reviewing a PR, and I think our "editor" system is the best
way to advance this.

AnnikaTjuka:
We were actually aware of the deleted zeros, see #749 . But we didn't discuss it further. 

I have gotten into the habit of opening the concept lists only in Sublime, so that something like this doesn't happen again.

LinguList:
Ah, interesting, so the ones (myself specifically) who SHOULD have been
aware of them should have paid more attention. As a general rule, as
these numbers are often reported from sources, they are like page
numbers, which means one should not modify them unless with a reason.

chrzyki:
I missed it as well and should have been more careful. However, it is a good lesson and it shows that everything is connected (which is a good thing I think!) and that our pipeline gets better at catching these things. Can I get an approval to merge this?

Issue #810, opened 2020-04-02
Lists mapped to retired Concepticon_IDs


AnnikaTjuka:
We have a few Concepticon_IDs which are retired but still have lists mapped to them, e.g., 1328 SCISSORS (Replacement_ID: 356 SHEARS). 

Should we move all lists that are mapped to a retired concept to the replacement ID?

LinguList:
Yes, this should be done. In fact, I thought there was even a check that
would prevent this to happen upon PRs. If there is no check, we need
one, @chrzykie, do you know of a test in this direction?

chrzyki:
Heya,

I don't think we explicitly test for that, but I can add it!

This is the list of retired concepts still in use:

```
Savelyev-2020-254-122 uses 2309 which is replaced by 1422
Savelyev-2020-254-176 uses 2275 which is replaced by 1388
Bodt-2019-664-621 uses 1785 which is replaced by 549
Sagart-2019-250-159 uses 1099 which is replaced by 363
Sagart-2019-250-212 uses 2307 which is replaced by 2249
Castro-2010-511-218 uses 2275 which is replaced by 1388
Huang-1992-1820-452 uses 2050 which is replaced by 404
Huang-1992-1820-1452 uses 1099 which is replaced by 363
Huang-1992-1820-1453 uses 1317 which is replaced by 501
Huang-1992-1820-1746 uses 2563 which is replaced by 29
Hale-1973-1798-256 uses 1785 which is replaced by 549
Bowern-2012-644-453 uses 1785 which is replaced by 549
BeijingDaxue-1964-905-527 uses 1317 which is replaced by 501
Chacon-2014-142-85 uses 2138 which is replaced by 2641
List-2014-1280-923 uses 1785 which is replaced by 549
Borin-2015-1532-1038 uses 1785 which is replaced by 549
Key-2016-1310-617 uses 1785 which is replaced by 549
Haspelmath-2009-1460-348 uses 1785 which is replaced by 549
Buck-1949-1110-224 uses 1785 which is replaced by 549
Luniewska-2016-299-116 uses 1328 which is replaced by 356
Luniewska-2019-299-116 uses 1328 which is replaced by 356
Othaniel-2017-300-140 uses 1099 which is replaced by 363
Othaniel-2017-300-205 uses 2275 which is replaced by 1388
Tolmie-1884-211-50 uses 2394 which is replaced by 301
Bremer-2016-200-47 uses 2275 which is replaced by 1388
[...]

PR #816, opened 2020-04-07
Remapping of retired IDs

AnnikaTjuka:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [X] refine existing Concepticon concept set mappings
- [ ] retire data

## Additional information

This PR is based on the discussion in issue #810 . I exchanged the IDs of retired concepts with their replacements.


AnnikaTjuka:
@chrzyki  I used the list which you provided in the issue. But when I changed LEGENDARY CREATURE 2138 to MYTHICAL BEING 2641, I got an error message:
`ERROR: Chacon-2014-142: deprecated concept set 2641 linked for Chacon-2014-142-85`

AnnikaTjuka:
@CarolinHu Would you mind moderating this PR?

chrzyki:
Heya,

`2641 MYTHICAL BEING` uses the now deprecated `sameas` relation in https://github.com/concepticon/concepticon-data/blob/master/concepticondata/conceptrelations.tsv#L547 which is triggering the error message. Removing/changing this relation fixes this.

LinguList:
THAT was the TEST!!!

The sameas relation was what tested and prevented to USE those concepts! 

When it was deleted, the tests were also deleted in this sense.

So we urgently need a test to check for retired concept ids, as we saw where this leads if tests are lacking.

Or we go for back to sameas.

xrotwang:
Why was the `sameas` relation deleted? I admit that using an otherwise experimental aspect of the data model for internal bookkeeping wasn't the best idea, but I'd also think that we must make backwards compatibility a higher priority now.

AnnikaTjuka:
The "sameas" relation had only one entry (LEGENDARY CREATURE 2138 - MYTHICAL BEING 2641). And I followed the above comment by @chrzyki to update the mapping in Chacon-2014.

LinguList:
Sameas was our test to check that only the lowest ID was used. When
those were deleted (but I honestly don't know, when), they yielded the
situation that there were not tests anymore whether one uses the wrong
identical concept, like wood instead of forest. This lead to the current
PR. So we need to decide how to handle this: adding tests, or going back
to sameas (by adding the old relations). And yes, @xrotwang is right, we
need to be careful with too many changes, so the compatibility with
lexibank and backwards will always have to be considered now.

xrotwang:
Maybe we could use a `Conceptset`  key in `retired.json` to replace the functionality of the `sameas` relation. But we probably have to dig through the git history to figure out all `sameas` relations we had, right?

xrotwang:
Just checked: [Concepticon 1.2](https://github.com/concepticon/concepticon-data/releases/tag/v1.2.0) is the last with a bigger set of `sameas` relations. So I propose to start re-implementing a marker for prefered Conceptsets (out of a set of identical ones) from 1.2 as baseline.

chrzyki:
@AnnikaTjuka apologies for the confusion, I didn't mean to mislead you and simply forgot about this, sorry! This also makes it clearer to me what @Schweikhard meant with the lower replacement IDs.

@LinguList Yes, that partly solves the mystery I guess. I also vaguely recalled that this was checked somewhere at some point but couldn't exactly remember when and were.

@xrotwang https://github.com/concepticon/concepticon-data/commit/f52696f851fd9298763e1b1ae4a93d8e0e910ccd#diff-63bb61c43e529bc566c18911fe6be401 is the last commit before there was only the `MYTHICAL BEING` `sameas` left.

chrzyki:
> 
> 
> Maybe we could use a `Conceptset` key in `retired.json` to replace the functionality of the `sameas` relation. But we probably have to dig through the git history to figure out all `sameas` relations we had, right?

Personally I'd find this much more transparent than using the relations for this purpose.

Concerning the `sameas` removals:

```
git log -p -S 'sameas' -- concepticondata/conceptrelations.tsv
```

LinguList:
Good news is: I know when that happened. We just have to check the history of conceptrelations.tsv. I bet it was 6-8 months ago. @schweikhard reorganized concept relations back then.
null

CarolinHu:
I am sorry, @AnnikaTjuka. With the unexpected turn this has taken and the commits yet to come, I don't think I'm the most suitable candidate for moderating this PR, as I am in no position to judge the consequences that a change in code may have for the other databases. I propose one of the people already active in this thread take on the role of moderator instead.
As an aside: thank you for this "decluttering" work :)

AnnikaTjuka:
No worries, @CarolinHu !

@chrzyki @LinguList @xrotwang I am not sure what to do next. Should I add the deleted "sameas" relations in the `retired.json`?

LinguList:
@AnnikaTjuka, not in the retired, but in conceptrelations.tsv. If you add them, and run the test, it should even show you if your remappings were correct (!).

AnnikaTjuka:
Ok, I'll do that.

In Chacon-2014, I replaced the retired ID with its replacement ID and got the error message as long as the "sameas" relation was still in the `conceptrelations.tsv`. But I think this was the case because the relation needs to be the other way around:
SOURCE | SOUCRE_GLOSS | RELATION | TARGET | TARGET_GLOSS
--| -- | -- | -- | --
2138 | LEGENDARY CREATURE | sameas | 2641 | MYTHICAL BEING

MYTHICAL BEING is the replacement for LEGENDARY CREATURE. Is that correct?

AnnikaTjuka:
I added the old "sameas" relations in `conceptrelations.tsv` and got error messages for some of the concepts, namely 2165, 2475, 2641. The relations are as follows:

SOURCE | SOUCRE_GLOSS | RELATION | TARGET | TARGET_GLOSS
--| -- | -- | -- | --
701 | SLOW | sameas | 2165 | SLOWLY
264 | HE-GOAT | sameas | 2475 | MALE GOAT
2641 | MYTHICAL BEING | sameas | 2138 | LEGENDARY CREATURE

Note that I didn't even remap lists with SLOW and SLOWLY in my first commit.

LinguList:
The rule of the retiring process is (or should be, was) to use the lower
ID as the active one, and retire the older one. I assume swapping around
the relations will change the error message or have it go away.

AnnikaTjuka:
I tried that for the MALE GOAT vs. HE-GOAT relation, but I still get the same error messages.

AnnikaTjuka:
But the error message complains about lists that are already mapped to MALE GOAT.

LinguList:
Look, the rule is: the CONCEPTICON_ID needs to be the smaller one. As
sameas is defined as all other relations, swapping should have an
effect. Can you please share all error messages when swapping and when
not doing this? Otherwise it is hard to debug.

AnnikaTjuka:
Error messages after inserting the deleted "sameas" relations:

`ERROR:Bodt-2019-664: deprecated concept set 2165 linked for Bodt-2019-664-443
ERROR:Chacon-2014-142: deprecated concept set 2641 linked for Chacon-2014-142-85
ERROR:Allen-2007-500: deprecated concept set 2165 linked for Allen-2007-500-496
ERROR:Kraft-1981-434: deprecated concept set 2475 linked for Kraft-1981-434-191
ERROR:List-2014-1280: deprecated concept set 2475 linked for List-2014-1280-741
ERROR:Borin-2015-1532: deprecated concept set 2475 linked for Borin-2015-1532-438
ERROR:Key-2016-1310: deprecated concept set 2475 linked for Key-2016-1310-799
ERROR:Haspelmath-2009-1460: deprecated concept set 2475 linked for Haspelmath-2009-1460-167
ERROR:Buck-1949-1110: deprecated concept set 2475 linked for Buck-1949-1110-122
ERROR:Janhunan-2008-235: deprecated concept set 2475 linked for Janhunan-2008-235-269
ERROR:Yang-2011-1014: deprecated concept set 2165 linked for Yang-2011-1014-1005
ERROR:Snider-2004-1700: deprecated concept set 2475 linked for Snider-2004-1700-966
ERROR:Xiao-2012-213: deprecated concept set 2165 linked for Xiao-2012-213-127
ERROR:List-2018-1105: deprecated concept set 2475 linked for List-2018-1105-109
ERROR:Rzymski-2020-1624: deprecated concept set 2475 linked for Rzymski-2020-1624-670
ERROR:Gauchat-1925-480: deprecated concept set 2165 linked for Gauchat-1925-480-341
ERROR:Grierson-1928-168: deprecated concept set 2475 linked for Grierson-1928-168-92
`

AnnikaTjuka:
Changing the order:
SOURCE | SOUCRE_GLOSS | RELATION | TARGET | TARGET_GLOSS
--| -- | -- | -- | --
2165 | SLOWLY | sameas | 701 | SLOW
2475 | MALE GOAT | sameas | 264 | HE-GOAT

Error messages:
`ERROR:Bodt-2019-664: deprecated concept set 2165 linked for Bodt-2019-664-443
ERROR:Chacon-2014-142: deprecated concept set 2641 linked for Chacon-2014-142-85
ERROR:Allen-2007-500: deprecated concept set 2165 linked for Allen-2007-500-496
ERROR:Kraft-1981-434: deprecated concept set 2475 linked for Kraft-1981-434-191
ERROR:List-2014-1280: deprecated concept set 2475 linked for List-2014-1280-741
ERROR:Borin-2015-1532: deprecated concept set 2475 linked for Borin-2015-1532-438
ERROR:Key-2016-1310: deprecated concept set 2475 linked for Key-2016-1310-799
ERROR:Haspelmath-2009-1460: deprecated concept set 2475 linked for Haspelmath-2009-1460-167
ERROR:Buck-1949-1110: deprecated concept set 2475 linked for Buck-1949-1110-122
ERROR:Janhunan-2008-235: deprecated concept set 2475 linked for Janhunan-2008-235-269
ERROR:Yang-2011-1014: deprecated concept set 2165 linked for Yang-2011-1014-1005
ERROR:Snider-2004-1700: deprecated concept set 2475 linked for Snider-2004-1700-966
ERROR:Xiao-2012-213: deprecated concept set 2165 linked for Xiao-2012-213-127
ERROR:List-2018-1105: deprecated concept set 2475 linked for List-2018-1105-109
ERROR:Rzymski-2020-1624: deprecated concept set 2475 linked for Rzymski-2020-1624-670
ERROR:Gauchat-1925-480: deprecated concept set 2165 linked for Gauchat-1925-480-341
ERROR:Grierson-1928-168: deprecated concept set 2475 linked for Grierson-1928-168-92
`

LinguList:
Okay, thanks, this helps. Since the problem is that the concept sets are
simply linked wrong. It seems that the sameas-relation is not depending
on the order, but on the CONCEPTICON_ID, as I mentioned before. If you
link Bodt-2165, while you should have linked 701 (701 is smaller than
2165), this is obvious that he error is on the link itself, right?

LinguList:
And it is also important to note here that the same holds for MALE GOAT
and HE-GOAT: HE-GOAT is the smaller number, so it should be given
preference. If one wants to change the aesthetics of the glosses, this
would mean one would have to replace HE-GOAT by MALE GOAT.

It is obvious that while there were many of these post-edits on the
lists, in order to arrive at a more "logical-feeling" mapping, with
better definitions and the like, the underlying rules were forgotten,
also since the sameas relations were deleted, so there was no way to check.

It is good that this happens now, as it is still not too much work to
get back, but it also shows that it was about time to get to our refined
review process, and to remind ourselves that changes in concepticon have
consequences somewhere else. If we go for linked data, we need to think
about the consequences of linking. The only way to cope with this is to
go through all of the individual links and correct them. 

The questions of what SHOULD be our preference, if it is he-goat or male
goat, would have to be resolved later. I would not advise to change any
gloss names now. But I'd suppose that all those who were also involved
in discussin the renaming of concepticon glosses now have a look and see
how they want to change them back.

@xrotwang, would this be trivial, or would we have to retire them things
again?

AnnikaTjuka:
When I did the remapping for this PR, I solely relied on the data given in concepticon.tsv and the `REPLACEMENT_ID` column:

ID | GLOSS | SEMANTICFIELD | DEFINITION | ONTOLOGICAL_CATEGORY | REPLACEMENT_ID
-- | -- | -- | -- | -- | --
264 | HE-GOAT | Animals | A male goat. | Person/Thing | 2475

Note that I did not change anything in some of the lists where an error message occurred, e.g., Allen-2007.

LinguList:
I know that, @AnnikaTjuka, it is not your error, it is errors
accumulated over the time, since the original principle: remapping can
only happen for the SMALLER ID, was broken at some point, after we lost
the same-as relations. It's not your fault. Its fault of me not paying
attention to changes that were done, and it is the fault of those doing
bulk-changes that they did not think of potential consequences of
touching the glosses and the concepts we have there. You could say it is
much more my fault than the bulk changes, since I should've paid more
attention.

We have two possibilities now:

## First one: go back to original

1. we go back to where we were, which means: replacment IDs that are
WRONG, as they are going from large number to small number must be fixed
(swapped around)

2. fixing the mappings to wrong sameas relations

3. all things discussed for the glosses, etc. need to be re-done then,
but this time with a PR under surveillance of all major editors

## Second one: new tests ignoring sameas relations

1. apply new tests that only check for replacment IDs, not sameas

2. fix those parts in the mapping process that MAKE USE of the sameas
relation as well (!)

3. make the PR for these checks

4. check with the PR by @AnnikaTjuka


@xrotwang, @chrzykie, any other alternatives?

chrzyki:
I'd vote for the second option, if this includes the the removal of the `sameas` relation and moves this mechanic either completely into the `REPLACEMENT` column and/or `retired.json`.

LinguList:
I would, however, insist on sticking to the lowest identifier option. This may seem silly and just a preference by myself to keep the identifers "nice" in some way, but in fact, what this does is forcing us all to rethink if we want to sacrifice a gloss (that is in fact just a gloss, another way of providing an identifier for an object) and moving it to the retired.json or doing other things. I emphasize that this is practical concern, not technical, as was my intention to keep IDs consecutive, so that each line would reflect the line number. Technically, you can discard this, but the curation of data has also an aesthetic aspect, and this aspect actually prevents us from making errors. So solution 2 would ALSO require to move back to taking the LOWEST ID and only retiring the older ones. The REASON for this is also: usually, we retire concepts or mark them as same as, when we create them but FORGET that they were ALREADY in the data. This is also what happened in our cases a lot, and you can see that the MALE DOG was added, because somebody did not see that there is already a he-dog (how should one?). So retiring or marking as "sameas" or "aliasing", however you call this, should follow the principle of taking the lowest ID. 

In that case, I would say: we have to write the tests first, before we check the data. So we put this PR on hold until then. And the test should test then:

1. is a concepticon ID used that is just an alias for another one?
2. are concepticon IDs defined as an alias for another one that should NOT be defined as such, since they contradict the order of IDs (and therefore the history of the project)?

AnnikaTjuka:
I added the result of the discussion about the test in issue #810 . 
I will close this PR if there are no objections.

chrzyki:
Yes, can in my opinion be closed. We can take of this then separately.

Issue #820, opened 2020-04-15
Unconformed remappings for "to know" in Bodt-2019-664 and Lieberherr-2017-100


LinguList:
Those two lists are both made by colleagues, representing languges they know very well, so the entries should have never been changed from "know (something)" to "KNOW", without consulting Tim directly, as he could just tell us.

I am raising this issue to emphasize, that:

1. we need to be extremely careful with bulk-edits, that involve the remapping of many different concept lists, as often, one may think one knows what is going on, but the lists have their specific history, and there are even almost "perfect" solutions here

2. always inform on the given lexibank repository about a bulk edit that affects a given list

In the future, I will further restrict these edits much more radically. An edit should only involve one remapping of a concept, and it should be agreed upon by three senior editors to whom this is brought up. Otherwise, we run danger of losing control.

This issue can be closed when:

- [ ] we have a definite answer from Tim on the meaning of "know" in both datasets (as it is the same languages, and he can tell us)
- [ ] the mapping is updated (unless Tim confirms the unspecific version of KNOW)

I have assigned this issue to almost everybody to bring this to your attention. I think it was very good we had all these discussions and also refinements of the data, don't understand me wrongly. But with the PRs on concepticon growing at such a paste, I have less chances to check what is being done, and sometimes, I (and other senior editors) still have the better overview, especially when the data comes from colleagues, so we must make the review process stricter, but this means I will need more time to also check everything thoroughly myself. 

If there are volunteers to ask Tim, I'll be glad, but I'll write to him as well, otherwise.

MacyL:
I can write him a message to confirm

LinguList:
Yes, thanks.

tresoldi:
CC or ping me later, @MacyL , so I'll fix it there if necessary or prepare a PR.

MacyL:
short answer : the author suggests to map "know" (Bodt-2019-664) and "to know" (Lieberherr-2017-100) to KNOW (SOMETHING) [1410].

long answer : the words **dɛn** or **dan** in Kho-Bwa languages means "know something". When it is referring as "know how", the word is in a specific descriptive compounds or loans.

MacyL:
I will make a PR as soon as possible.

CarolinHu:
Closing this since the KNOW issue was addressed and probably everyone had a chance to read the note by now.

PR #822, opened 2020-04-16
Add Urban-2016-160

CarolinHu:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

A pretty straightforward list, I think. @MacyL, could you perhaps moderate this PR?


LinguList:
I volunteer to review, @macyl.

concepticondata/conceptlists.tsv
```diff
@@ -316,3 +316,4 @@ Hartmann-2013-162	Hartmann, Iren and Haspelmath, Martin and Taylor, Bradley	2013
...
 Lee-2015-246	Lee, Sean	2015		246	basic, areal	English	Koreanic	https://doi.org/10.1371/journal.pone.0128448	Lee2015		The basic concepts in this list are selected to &quot;quantitatively describe the evolutionary characteristics&quot; (of Koreanic).	e0128448	
+Urban-2016-160	Urban, Matthias	2016		160	basic	English	Global	http://doab-test.huma-num.fr/handle/20.500.12657/23584	Urban2016		This list of 160 nominal items was compiled in order to compare morphological complexity across 78 languages of different language families. The concepts were organized into four semantic categories relating to nature, artifacts, the body and temporal concepts (plus some miscellaneous words). The list is not directly based on another, although Brown 1999 and [Buck 1949](:ref:Buck-1949-1110) provided some inspiration.	571-572
```
LinguList, line 319:
```
The list is not directly based on another
```
rather:
```
The list is not directly based on another concept list
```
English doesn't allow to use another with out "one" or some filling word, and this is more explicit.

concepticondata/references/references.bib
```diff
@@ -3163,3 +3163,13 @@ @article{Lee2015
   journal      = {{PLOS} {ONE}}
 }
 
+@incollection{Urban2016,
+    author     = {Urban, Matthias},
+    year       = {2016},
+    title      = {Motivation by formally analyzable terms in a typological perspective: An assessment of the variation and steps towards explanation},
+    booktitle  = {The lexical typology of semantic shifts},
+    pages      = {555–576},
+    publisher  = {De Gruyter},
```
LinguList, line 3172: I think they tend to spell it now "DeGruyter". And you should add teh address as well. This would be `address = {Berlin and New York}`.

concepticondata/conceptlists.tsv
```diff
@@ -316,3 +316,4 @@ Hartmann-2013-162	Hartmann, Iren and Haspelmath, Martin and Taylor, Bradley	2013
...
 Lee-2015-246	Lee, Sean	2015		246	basic, areal	English	Koreanic	https://doi.org/10.1371/journal.pone.0128448	Lee2015		The basic concepts in this list are selected to &quot;quantitatively describe the evolutionary characteristics&quot; (of Koreanic).	e0128448	
+Urban-2016-160	Urban, Matthias	2016		160	basic	English	Global	http://doab-test.huma-num.fr/handle/20.500.12657/23584	Urban2016		This list of 160 nominal items was compiled in order to compare morphological complexity across 78 languages of different language families. The concepts were organized into four semantic categories relating to nature, artifacts, the body and temporal concepts (plus some miscellaneous words). The list is not directly based on another concept list, although Brown 1999 and [Buck 1949](:ref:Buck-1949-1110) provided some inspiration.	571-572
```
AnnikaTjuka, line 319: You could add "Brown 1999" as a reference in the reference file and use the same linking as for Buck.

concepticondata/conceptlists/Urban-2016-160.tsv
```diff
@@ -0,0 +1,161 @@
+ID	NUMBER	ENGLISH	CATEGORY	CONCEPTICON_ID	CONCEPTICON_GLOSS
+Urban-2016-160-1	1	animal	Nature-related and topological concepts	619	ANIMAL
+Urban-2016-160-2	2	ashes	Nature-related and topological concepts	646	ASH
+Urban-2016-160-3	3	bark	Nature-related and topological concepts	1204	BARK
+Urban-2016-160-4	4	bay	Nature-related and topological concepts	663	BAY
+Urban-2016-160-5	5	beak	Nature-related and topological concepts	73	BEAK
+Urban-2016-160-6	6	bird	Nature-related and topological concepts	937	BIRD
+Urban-2016-160-7	7	bloom (blossom, flower)	Nature-related and topological concepts
```
AnnikaTjuka, line 8: Maybe 239 FLOWER ? Urban-2011-160-7 was also mapped to this concept.

concepticondata/conceptlists/Urban-2016-160.tsv
```diff
@@ -0,0 +1,161 @@
...
+Urban-2016-160-84	84	knife	Artifacts	1352	KNIFE
+Urban-2016-160-85	85	ladder	Artifacts	1464	LADDER
+Urban-2016-160-86	86	mirror	Artifacts	1171	MIRROR
+Urban-2016-160-87	87	needle	Artifacts	1382	NEEDLE (FOR SEWING)
+Urban-2016-160-88	88	paper	Artifacts	648	PAPER
+Urban-2016-160-89	89	pen	Artifacts	794	PEN
+Urban-2016-160-90	90	rope	Artifacts	1218	ROPE
+Urban-2016-160-91	91	scissors	Artifacts	1328	SCISSORS
+Urban-2016-160-92	92	shoe	Artifacts	1381	SHOE
+Urban-2016-160-93	93	road/street/way	Artifacts	2457	PATH OR ROAD
```
AnnikaTjuka, line 94: Urban-2011-160-93 was mapped to 667 ROAD

concepticondata/conceptlists/Urban-2016-160.tsv
```diff
@@ -0,0 +1,161 @@
...
+Urban-2016-160-103	103	bladder	Body Parts and Body Fluids	1746	BLADDER
+Urban-2016-160-104	104	blood	Body Parts and Body Fluids	946	BLOOD
+Urban-2016-160-105	105	bone	Body Parts and Body Fluids	1394	BONE
+Urban-2016-160-106	106	brain	Body Parts and Body Fluids	1583	BRAIN
+Urban-2016-160-107	107	breast	Body Parts and Body Fluids	1402	BREAST
+Urban-2016-160-108	108	buttocks	Body Parts and Body Fluids	834	BUTTOCKS
+Urban-2016-160-109	109	calf	Body Parts and Body Fluids	498	CALF OF LEG
+Urban-2016-160-110	110	cheek	Body Parts and Body Fluids	1730	CHEEK
+Urban-2016-160-111	111	chin	Body Parts and Body Fluids	1510	CHIN
+Urban-2016-160-112	112	eyeball	Body Parts and Body Fluids
```
AnnikaTjuka, line 113: Other lists including (Urban-2011-160-112) mapped "eyeball" to EYE.

concepticondata/conceptlists/Urban-2016-160.tsv
```diff
@@ -0,0 +1,161 @@
...
+Urban-2016-160-117	117	fingernail	Body Parts and Body Fluids	1258	FINGERNAIL
+Urban-2016-160-118	118	guts	Body Parts and Body Fluids	1334	GUTS
+Urban-2016-160-119	119	heart	Body Parts and Body Fluids	1223	HEART
+Urban-2016-160-120	120	jaw	Body Parts and Body Fluids	798	JAW
+Urban-2016-160-121	121	kidney	Body Parts and Body Fluids	489	KIDNEY
+Urban-2016-160-122	122	lip	Body Parts and Body Fluids	478	LIP
+Urban-2016-160-123	123	liver	Body Parts and Body Fluids	1224	LIVER
+Urban-2016-160-124	124	lungs	Body Parts and Body Fluids	688	LUNG
+Urban-2016-160-125	125	mouth	Body Parts and Body Fluids	674	MOUTH
+Urban-2016-160-126	126	mucus	Body Parts and Body Fluids
```
AnnikaTjuka, line 127: This is commonly mapped to 208	NASAL MUCUS (SNOT) but I see that "snot" is also in the list. Because "snot" is in brackets, I'd propose to map only "mucus". We might also want to update the mapping in Urban-2011-160 because both glosses are mapped to the same concept.

LinguList:
Yes, it is fine. @Macyl, could you please just answer and let us know if
you are willing to manage this PR or not? If not, I can also do it, or
we can ask somebody else. You can also just say that you are too busy
with the Peiros profile to do it this week, this is fine, we just need a
short answer ;-)

CarolinHu:
Thank you for reviewing, @AnnikaTjuka! 
Concerning the comments you left with regard to the Urban-2011 list, I’ll answer them as a bundle here. I don’t want to reinvent the wheel, but I think there’s an argument for remapping them in the 2011 list (unless there is some background I am not aware of, as was recently discussed in #820). In any case, I think it will be best to wait for another review before settling this.

**-road/street/way**: The 2011 paper looked specifically at polysemy and marking and found patterns relating road/street/way to the Milky Way. I think ROAD on its own, without the less paved PATH/WAY meaning, would just be too narrow for the pattern Urban wanted to explore.

**-bloom (blossom/flower)**: This is a tricky one, given that bloom/blossom and flower are pretty different concepts. However, since bloom as a noun is closer to blossom than to flower, I left it unmapped. It could also be the verb meaning, although that seems unlikely, given the noun focus of the list.

**-eyeball**: In the 2016 list, Urban specifically states: “Regarding body-part terms, which have been rather well studied from a cross-linguistic point of view, care was taken that parts are included that
have hitherto received relatively little attention.” (p. 557) I would argue that eye would be such a well-studied body part. So I interpreted in the way that he really went for eyeball (i.e. apple of the eye) instead of the whole organ. In addition, he seems to have split up the organ in the list, seeing that pupil, eyelid and eyelash are part of the list, too, so eyeball would be more in line with the other concepts. Likewise, there is no nose in the list, but nostrils. Eyeball/apple of the eye could be a new concept for Concepticon, though?

**- mucus and snot**: It’s hard to say here what the author meant, but SNOT is much narrower and, in my opinion, closer to NASAL MUCUS than MUCUS as the latter one can be an umbrella term for any mucus that the human body possibly produces (cervical mucus, phlegm etc.)

concepticondata/conceptlists.tsv
CarolinHu: Is that something that is usually done, i.e. adding references that do not have a Concepticon list counterpart (yet)? If so, I'll add it. Good point.

LinguList: Yes, if you reference something, you need to write as [Brown (1998)](:bib:Brown1998) and add Brown to the references.bib. I missed that, good catch @AnnikaTjuka .

concepticondata/conceptlists/Urban-2016-160.tsv
Kristina-Pianykh: Yes, this is a rather ambiguous case. I would probably check with the kind of languages (most importantly, cultures where these languages are spoken) when making a decision on this one.

concepticondata/conceptlists/Urban-2016-160.tsv
Kristina-Pianykh: Hale-1973-1798-122 eyeball is also mapped to EYE. Perhaps, this instance can be linked to EYE too also because it doesn't contrast in this list with anything other eye-related, and the concept of EYE seems logical among other body parts concepts.

concepticondata/conceptlists/Urban-2016-160.tsv
Kristina-Pianykh: I checked with wikipedia on 'mucus' and it turns out to be a broader term than what is commonly understood by 'snot':

> Mucus serves to protect epithelial cells in the linings of the respiratory, digestive, and urogenital systems, and structures in the visual, and auditory systems from pathogenic fungi, bacteria[2] and viruses. Most of the mucus in the body is produced in the gastrointestinal tract.

So I'd leave this gloss unmapped. And yes, the corresponding mappings in Urban-2011-160 need to be modified.

Kristina-Pianykh:
Thanks @CarolinHu for a such a detailed account on the problematic cases! I agree with you on the matter of 'mucus vs. snot', 'path/way/street' and I'm now more inclined to think 'eyeball' should be unmapped (the strategy also preserved for ambiguous cases). And the fact that the list also lacks such basic body part concepts as 'nose' and 'ear' seems to support this point.

But as far as 'bloom (blossom, flower)' is concerned, I would want to know what others think as I would also map it to FLOWER.

LinguList:
Does seem to be dead the link. You can drop it: we have the article
referenced, so it is clear where it appears. I would propose to include
a PDF, however. From the pages with the concepts.

I attach it with this post (if it doesn't appear, let me know). @chrzyki
usually uploads PDF to cdstar...

LinguList:
Okay, here's the PDF again, attachments don't seem to work.

[Urban2016.pdf](https://github.com/concepticon/concepticon-data/files/4498473/Urban2016.pdf)

LinguList:
As to the debate regarding the "eye-ball": the Augapfel seems to be a specific body part with vertebrates, and some lists distinguish it, so we may want to remap that in the future (but only, once we have a save account on how to deal with retirements and changing too many lists in bulk edits). I added issue #285 

@CarolinHu, you can leave eyeball unmapped, and we modify it later. The good thing is: going from <NA> mappings to a positive mapping can be done easily, when new concept sets emerge, but changing existing mappings is a bit more problematic.

CarolinHu:
Thank you all for your helpful reviews! So, in the next commit, I will (this is more a reminder for myself and a checklist for @MacyL to see whether anything was forgotten):
- delete the link
- add Brown reference
- map bloom (blossom/flower) to flower since two reviewers agreed on it
- leave the other mappings as they are, including no mapping for eyeball (as this will be handled in a separate issue)


What about the Urban 2011 list and concepts other than the eyeball, @LinguList? Should I remap the concepts discussed there as well or rather leave that list for now? (I could also do a separate PR to avoid confusion of the two lists.)
That would include:
- changing the mapping of road/street/way from ROAD to PATH OR ROAD
- deleting the mapping NASAL MUCUS (SNOT) for mucus, as snot is already mapped to the same concept
- changing the mapping of river/stream from RIVER to FLOWING BODY OF WATER
- change the mapping of calf from CALF to CALF (OF LEG), as calf is in the body parts category

LinguList:
Please only work on this list, no modification of any other lists. This
must be done in a separate issue, to avoid confusion. Thanks all for
your work. @MacyL, can you please finalize this and also merge, once all
is okay?

MacyL:
From my side, it looks pretty good. I apologise for my late reponses, my mailbox somehow didn't send me notifications until Saturday. 
I would like to confirm the reference with @CarolinHu and then I will merge this PR.  The REFS, PDFs are all given the bib keys as `Urban2016`. You uploaded the PDFs to cloud storage, and then mention Brown1998 as an "inspiration" in the detail. So this is the final solution that everyone agreed, correct?

CarolinHu:
No worries! Yes, since the PDF was provided above, I added Urban2016 as PDF shorthand in conceptlists.tsv. The actual uploading to the server will probably be done by @chrzyki. As for Brown, it was agreed to add the book to the reference file.

PR #827, opened 2020-04-22
Add Koelle-1854-335 concept list (Polyglotta Africana)

tresoldi:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR adds a concept list for Polyglotta Africana (lexibank dataset https://github.com/lexibank/polyglottaafricana), as mapped mostly by @Kristina-Pianykh .

While some glosses would be candidates for new concept sets (such as IVORY and NOSE RING), this PR is *not* proposing new concepts.

It has been a while since I made my last Concepticon PR and I can see some practices changed, sorry for any mistake. :smiley:


tresoldi:
Pinging @LinguList , @chrzyki and @Kristina-Pianykh

LinguList:
Okay, @tresoldi, the review process has changed. You are allowed to
suggest an EDITOR of a pull request, not assign any reviewers. The
editor will then assign reviewers and also close the PR and moderate
discussions. So please assign somebody as an editor, but be aware that
we have another PR ongoing, although the two should not clash, I think.

tresoldi:
So it is supposed to work like a paper? I'll suggest @chrzyki as editor, then, so that you @LinguList and @Kristina-Pianykh can review.

As for potential clashes, no problem on waiting for the Zhang-2019-122 PR to finish. I'll leave it to the editor.

chrzyki:
@Schweikhard, @Kristina-Pianykh, @LinguList can you review this PR? I'll have another look at it as well and compare with my notes and comment after you've had time to do so.

tresoldi:
To facilitate, here is the discussion for my changes: https://github.com/lexibank/polyglottaafricana/pull/5

concepticondata/conceptlists.tsv
```diff
@@ -317,3 +317,4 @@ Lundgren-2020-275	Lundgren, Olof	2020		275	areal	English	The Tupí–Guaraní la
...
 Lee-2015-246	Lee, Sean	2015		246	basic, areal	English	Koreanic	https://doi.org/10.1371/journal.pone.0128448	Lee2015		The basic concepts in this list are selected to &quot;quantitatively describe the evolutionary characteristics&quot; (of Koreanic).	e0128448	
 Urban-2016-160	Urban, Matthias	2016		160	basic	English	Global		Urban2016	Urban2016	This list of 160 nominal items was compiled in order to compare morphological complexity across 78 languages of different language families. The concepts were organized into four semantic categories relating to nature, artifacts, the body and temporal concepts (plus some miscellaneous words). The list is not directly based on another concept list, although [Brown 1999](:bib:Brown1999) and [Buck 1949](:ref:Buck-1949-1110) provided some inspiration.	571-572	
+Koelle-1854-335	Koelle, Sigismund W.	1854		335	basic, areal	English	African languages	http://reflex.cnrs.fr/	Koelle1854		This list of lexical and phrasal items reflects the comparative vocabulary collected for more than a hundred African languages in the 19th century the Church Missionary House. It includes basic vocabulary, cultural and areal items.		Polyglotta Africana
```
LinguList, line 320: "in the 19th century **by** the Church Missionary House"
"cultural**,** and areal items."

concepticondata/conceptlists/Koelle-1854-335.tsv
```diff
@@ -0,0 +1,336 @@
...
+Koelle-1854-335-70	70	Egg	744	EGG
+Koelle-1854-335-71	71	Eight	1705	EIGHT
+Koelle-1854-335-72	72	Eighteen	1713	EIGHTEEN
+Koelle-1854-335-73	73	Elbow	981	ELBOW
+Koelle-1854-335-74	74	Elder Brother (My, Thy)	1759	OLDER BROTHER
+Koelle-1854-335-75	75	Elder Sister	1758	OLDER SISTER
+Koelle-1854-335-76	76	Elephant	1290	ELEPHANT
+Koelle-1854-335-77	77	Eleven	1706	ELEVEN
+Koelle-1854-335-78	78	Ewe	1345	EWE
+Koelle-1854-335-79	79	Ewe (Sheep)	1345	EWE
```
Schweikhard, line 80: EWE is mapped twice.

concepticondata/conceptlists/Koelle-1854-335.tsv
```diff
@@ -0,0 +1,336 @@
...
+Koelle-1854-335-82	82	Farm	201	FARM
+Koelle-1854-335-83	83	Father (My Father, Thy Father)	1217	FATHER
+Koelle-1854-335-84	84	Female Breast	1402	BREAST
+Koelle-1854-335-85	85	Female Slave		
+Koelle-1854-335-86	86	Fifteen	1709	FIFTEEN
+Koelle-1854-335-87	87	Finger	1303	FINGER
+Koelle-1854-335-88	88	Finger (thumb)	1781	THUMB
+Koelle-1854-335-89	89	Fire	221	FIRE
+Koelle-1854-335-90	90	Firewood	10	FIREWOOD
+Koelle-1854-335-91	91	Firewood (dry wood)	10	FIREWOOD
```
Schweikhard, line 92: FIREWOOD is mapped twice.

concepticondata/conceptlists/Koelle-1854-335.tsv
```diff
@@ -0,0 +1,336 @@
...
+Koelle-1854-335-128	128	House	1252	HOUSE
+Koelle-1854-335-129	129	I bathe (wash myself)	138	BATHE
+Koelle-1854-335-130	130	I beg	3534	BEG
+Koelle-1854-335-131	131	I break a stick	2558	BREAK (CLEAVE)
+Koelle-1854-335-132	132	I breathe	1407	BREATHE
+Koelle-1854-335-133	133	I buy	1869	BUY
+Koelle-1854-335-134	134	I call a slave		
+Koelle-1854-335-135	135	I catch a fish	2638	FISHING
+Koelle-1854-335-136	136	I come	1446	COME
+Koelle-1854-335-137	137	I cook meat
```
Schweikhard, line 138: COOK (SOMETHING)?

concepticondata/conceptlists/Koelle-1854-335.tsv
```diff
@@ -0,0 +1,336 @@
...
+Koelle-1854-335-148	148	I eat corn		
+Koelle-1854-335-149	149	I eat rice		
+Koelle-1854-335-150	150	I eat yam		
+Koelle-1854-335-151	151	I eat yams		
+Koelle-1854-335-152	152	I fall	2894	TUMBLE (FALL DOWN)
+Koelle-1854-335-153	153	I flog a child		
+Koelle-1854-335-154	154	I give thee	1447	GIVE
+Koelle-1854-335-155	155	I go	695	GO
+Koelle-1854-335-156	156	I hear	1408	HEAR
+Koelle-1854-335-157	157	I kill a fowl
```
Schweikhard, line 158: SLAUGHTER?

concepticondata/conceptlists/Koelle-1854-335.tsv
```diff
@@ -0,0 +1,336 @@
...
+Koelle-1854-335-179	179	I take	1749	TAKE
+Koelle-1854-335-180	180	Itch	148	ITCH
+Koelle-1854-335-181	181	Ivory		
+Koelle-1854-335-182	182	I weep	1839	CRY
+Koelle-1854-335-183	183	King	1508	KING
+Koelle-1854-335-184	184	Knee	1371	KNEE
+Koelle-1854-335-185	185	Knife	1352	KNIFE
+Koelle-1854-335-186	186	Kuskus (bearing like Oats)		
+Koelle-1854-335-187	187	Leaf	628	LEAF
+Koelle-1854-335-188	188	Leaf (grass)	628	LEAF
```
Schweikhard, line 189: LEAF is mapped twice.

concepticondata/conceptlists/Koelle-1854-335.tsv
```diff
@@ -0,0 +1,336 @@
...
+Koelle-1854-335-322	322	Water	948	WATER
+Koelle-1854-335-323	323	Water (fresh)		
+Koelle-1854-335-324	324	Well	954	WELL
+Koelle-1854-335-325	325	Wet	1726	WET
+Koelle-1854-335-326	326	White	1335	WHITE
+Koelle-1854-335-327	327	White Man	2839	WHITE MAN
+Koelle-1854-335-328	328	witch	824	WITCH
+Koelle-1854-335-329	329	Woman	962	WOMAN
+Koelle-1854-335-330	330	Yam	410	YAM
+Koelle-1854-335-331	331	Yam (wild)	410	YAM
```
Schweikhard, line 332: YAM is mapped twice.

concepticondata/conceptlists/Koelle-1854-335.tsv
```diff
@@ -0,0 +1,336 @@
...
+Koelle-1854-335-332	332	Yesterday	1174	YESTERDAY
+Koelle-1854-335-333	333	young	1207	YOUNG
```
Schweikhard, line 334: YOUNG is mapped twice.

tresoldi:
@LinguList it is actually from RefLex... I guess we should make this more explicit, and perhaps even rename the list? (Not sure to what... It would feel wrong to have something like Segerer-19XX...)

concepticondata/conceptlists/Koelle-1854-335.tsv
```diff
@@ -0,0 +1,336 @@
...
+Koelle-1854-335-122	122	Heel	980	HEEL
+Koelle-1854-335-123	123	Hell	878	HELL
+Koelle-1854-335-124	124	Hoe	284	HOE
+Koelle-1854-335-125	125	Honey	942	HONEY
+Koelle-1854-335-126	126	Horse	615	HORSE
+Koelle-1854-335-127	127	Hot	1286	HOT
+Koelle-1854-335-128	128	House	1252	HOUSE
+Koelle-1854-335-129	129	I bathe (wash myself)	138	BATHE
+Koelle-1854-335-130	130	I beg	3534	BEG
+Koelle-1854-335-131	131	I break a stick	2558	BREAK (CLEAVE)
```
Schweikhard, line 132: Either this should be unmapped or some of the other phrase concepts mapped.

Schweikhard:
I suppose the duplicate mappings are due to the combination of different concept lists that you assume in the other PR?

concepticondata/conceptlists/Koelle-1854-335.tsv
tresoldi: I decided to do so (as in the case of FIREWOOD), because for some languages the gloss is `Ewe` and for others it is `Ewe (Sheep)`. No language has both and exploring the data (particularly for West African languages) indicated that they are cognates.

We could keep only the item 78, far more frequent, if it is not allowed to have two glosses mapped to the same concept set.

concepticondata/conceptlists/Koelle-1854-335.tsv
tresoldi: Se comments for EWE, same thing applies.

concepticondata/conceptlists/Koelle-1854-335.tsv
tresoldi: The entries are an actual translation of "I cook meat", with the object and for some languages that don't drop the pronoun also the subject.

concepticondata/conceptlists/Koelle-1854-335.tsv
tresoldi: I suppose we could map it... It is not entirely clear from the data if in all cases it is a "single" entity, but the case is analogous to "catch a fish"...

concepticondata/conceptlists/Koelle-1854-335.tsv
tresoldi: Same case for EWE and FIREWOOD. Once more, if needed we can drop the mapping for 188 (in this case, I think there is only one language using it).

concepticondata/conceptlists/Koelle-1854-335.tsv
tresoldi: As LEAF and others above. 331 could be dropped if we decide for a single mapping.

concepticondata/conceptlists/Koelle-1854-335.tsv
tresoldi: Again, same as LEAF, FIREWOOD, etc.

concepticondata/conceptlists/Koelle-1854-335.tsv
tresoldi: Oh you are right, I missed this. Will unmap.

LinguList:
@tresoldi, @Kristina-Pianykh was comparing with the original, and I find it essential, even if we have a provider like Reflex, that we stick to the original, similar to what we did with halenepal (which was tedious, but also more concepts), or suntb, where we also have the list with Chinese glosses, and the like. So the main list should be Koelle's, then reflex items should be placed in their corresponding rows in an extra column.

LinguList:
This is also, what @Kristina-Pianykh annotated on the google sheet, I believe. 

So let us search for the original, so we can compare here.

concepticondata/conceptlists/Koelle-1854-335.tsv
Schweikhard: I agree that it then makes sense to map it to both. But I guess if we switch to the origin list anyway this problem doesn't apply anymore.

tresoldi: They are actually in the source. My guess is that it indicates that the language has no different word for EWE and SHEEP (only a different one for RAM, or perhaps for the castrated one -- how is it called in English?), so that in theory a concept EWE OR SHEEP.

tresoldi: Found it, castrated male sheep is "wether". Luckily it is not in this list.

LinguList:
Okay, who is looking into the PDF? In the worst case, we have to order it via library? 

I think it is dangerous to discuss this too early in the detail, when we don't have a real scan of the book, since the mappings will have to conform to that, right? pinging @tresoldi, @chrzyki, @Kristina-Pianykh and @Schweikhard : as long as we lack the PDF, we don't need to discuss details in mapping.

chrzyki:
The book should still be in our office.

LinguList:
Okay, we have the book? That is excellent. I will visit the office at some point, I can just grab it and scan parts of it.

chrzyki:
I *think*. I was heading for the office today anyway (after lunch). I'll check and keep you updated.

LinguList:
Even better, just found the email by @Kristina-Pianykh. She already made the comparison, @chrzyki and @tresoldi, you were in CC of that email. Look for August 9. There's the csv below. An email from 20th shows a small scan of the polyglotta text.
And she already made a first comparison list:

[Original_Polyglotta.csv.txt](https://github.com/concepticon/concepticon-data/files/4521668/Original_Polyglotta.csv.txt)

So I propose, but this is up to @chrzyki as editor, to close this PR and get back to the original list. I tried to make this clear from the beginning, but it seems you were not really following me, or I did not express myself correctly: The situation is the same as with STEDT data. We need to aim for the best, and we HAVE been working on the original book, to map it, already in August last year (as I realized only now). So the lexibank dataset should be based on a combined, revised concept list that contains reflex customary glosses which are often different, and could also be placed in the same cell (if they refer to identical concepts) and the original resource, the book, in scan from 1854.

Ideally, I'd also ask for making a short PDF that shows the glosses, how they look in the book. 

so the list should double check with the original, and then provide:

* number (even if added by us)
* english (as provided in source, variants of concepts inside the individual chapters can be excluded, unless reflex lists them)
* reflex_gloss (multiple items per cell possible, serves for identification only)
* page (I think, which page in the source which item can be found is useful to know)

tresoldi:
So, a list with the original gloss and the gloss in Reflex (as used in the `polyglottaafricana` dataset), with the dataset matching the latter.

Something like the following?

| NUMBER ORIGINAL | BASELIST | REFLEX | CONCEPICON_ID | CONCEPICON_GLOSS |
| --- | --- | --- | --- | --- |
| 1 | One | One | 1493 | ONE |
| 5a | (then they begin again) // (then they generally begin again from one) | | 
| 75 | Nail (of finger & toe) | Nail (of Finger and Toe) | 1258 | FINGERNAIL |

tresoldi:
For something like the table above, I could take care of it (with the exception of the page).

LinguList:
I want to have the PDF first, and have a look at this myself, now that
we have it almost perfectly lexibankified. You received the TXT file
that @KristinaPianykh provided, right? I uploaded it in my comment
before. This yes can be the basis, but I want to have the PDF to compare
(should've done this already half a year ago, but well, now it must be
done. We can't annotate data that we haven't seen in original. At least
this was always the ambition of Concepticon etc. to go beyond this.

tresoldi:
Ok, I will wait. We can always reopen the same PR in the future, if you prefer to close it now.

PR #828, opened 2020-04-23
redefine mappings

LinguList:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR modifies the mappings in concordance with the rule of not using concept sets with a replacment_ID in concepticon.tsv. It was created with help of the modifed code in pyconcepticon that adds these checks [see here](https://github.com/concepticon/pyconcepticon/pull/28)

I think there is no need to review this a long time, it just shows that the problem of aliases can be seen as fixed now. It is in the check of an individual list, it is in the mappings, where the replacments don't occur anymore, so we can ignroe the sameas relations from the past.


mappings/map-af.tsv
```diff
@@ -180,12 +180,12 @@ ID	GLOSS	PRIORITY
 787	SAIL///seil	2
 1948	SAW///saag	4
 2783	SCALES (MEASURE INSTRUMENT)///skaal	2
-1328	SCISSORS///skêr	2
```
LinguList, line 183: Please check specifically this cases, where SCISSORS was still preferred in our mapping files, but should not be.

Issue #886, opened 2020-07-21
please add what's required for 2.4b1


xrotwang:

LinguList:
The good thing is: we have this already. And we have all authors, editors, etc. I might only want to have a check on the .zenodo.json (or add it if it is not there yet).

LinguList:
So the most important part were the new datasets added by @AnnikaTjuka in the past.

AnnikaTjuka:
I have pushed all the necessary changes with regard to NoRaRe data sets in the last weeks. So now more open issues from my side.

PR #887, opened 2020-07-21
pre-release 2.4.0-rc.1

xrotwang:
addresses #886


xrotwang:
As far as I can see, the concepticon app data has never been under version control (it's excluded in `.gitignore`). Do you think this should be changed?

LinguList:
Ah, I see, then we could in fact also delete the concepticon,json file? I tend to update it on https://digling.org/calc/concepticon/, and it goes into norare, but there I already have a recent version (and I added the full app, just for completeness, although we have to find a common policy then anyway, when we merge both datasets). 

And even more important, I just saw it: the app data is in `app/data`, so no need to delete, it has never been there.

xrotwang:
Ok, I'll flesh out `concepticon citation` to also write a `.zenodo.json`, based on `CONTRIBTORS.md`, and add the result here.

LinguList:
```
{
    "title": "Concepticon. A resource for the linking of concept lists",
    "creators": [
        {
            "name": "Johann-Mattis List"
        },
        {
            "name": "Christoph Rzymski"
        },
        {
            "name": "Simon Greenhill"
        },
        {
            "name": "Nathanael E. Schweikhard"
        },
        {
            "name": "Kristina Pianykh"
        },
        {
            "name": "Annika Tjuka"
        },
        {
            "name": "Mei-Shin Wu"
        },
        {
            "name": "Annika Tjuka"
        },
        {
            "name": "Robert Forkel"
        }
    ],
    "description": "<p>Cite the library as:</p>\n\n<blockquote>\n<p>List, J., C. Rzymski, S. Greenhill, N. Schweikhard, K. Pianykh, A. Tjuka, M. Wu, and R. Forkel (2020): Concepticon. A resource for the linking of concept lists. Version 2.4.0.beta. Max Planck Institute for the Science of Human History: Jena. https://concepticon.clld.org/.</p>\n</blockquote>",
    "access_right": "open",
    "keywords": [
        "linguistics",
        "cldf:Wordlist",
        "concept list"
    ],
    "license": {
        "id": "CC-BY-NC-4.0"
    },
    "contributors": [],
    "upload_type": "dataset",
    "communities": [
        {
            "identifier": "dighl"
        },
        {
            "identifer": "calc"
        }
    ]
}
```

LinguList:
I just prepared this one. We could, of course, also list all contributors in contributors.md to this, may be even cooler for the future, so new contributors would then be added there as well to our zenodo.json?

AnnikaTjuka:
@LinguList I'm listed twice in the "creators:" list. Should we add @CarolinHu instead:
```
 {
            "name": "Mei-Shin Wu"
        },
        {
            "name": "Carolin Hundt"
        },
        {
            "name": "Robert Forkel"
        }
```

LinguList:
Carolin can be discussed for 2.4, your name should only appear once, I
copy-pasted the passages in json, so I mistakenly added you twice. For
this version, I'd leave it at the editors we had so far, we only take
new editors with a full release.

xrotwang:
@LinguList where's the NC clause in the license coming from? We always had CC-BY, no?

xrotwang:
@LinguList regarding @CarolinHu : I consider this a "release candidate" (hence the `-rc.1` suffix) for the 2.4 release. So technically, this **is** a full release. Now we could regard not adding new editors right now an oversight which will be fixed with the proper release - but if we already know she should be added then I'd do it now.

LinguList:
sorry, delete it, this is from lingpy, but even there, we can probably
discard it. I just copypasted an older .zenodo.json, sorry

LinguList:
That's exactly the discussion I wanted to avoid: that we openly discuss
these things, which we normally do in a quiet email session, where we
set up the order of authors (which may have to change also), and the like.

What is important to mention is that we may even consider other people
to be added to the editor's board this time, and we don't have the time
to discuss all of this now (and I may want to talk with people and
encourage them to help, etc.).

So I really would say: please let us just take the order and authors
from last time. If that's not possible, then we rather do not do a
release and ask people to check out a specific branch or git hash: all
after all it is just a paper we submit, if the paper was accepted, it
would be a different question, but we need to pass reviewers, and
editors, and we'll likely want to have another release anyway in three
months from now...

xrotwang:
`.zenodo.json` is now created via `concepticon citation`.

LinguList:
very cool!!!

xrotwang:
ok, so then this is probably complete.

LinguList:
so we can merge?

PR #900, opened 2020-08-04
Fix problematic double quotes

chrzyki:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This fixes (only breaking, illegal) quotation marks. @AnnikaTjuka, would you mind having a quick look at this?


concepticondata/conceptlists/Gauchat-1925-480.tsv
```diff
@@ -59,14 +59,14 @@ Gauchat-1925-480-57	57	the meadow	le pré	pratu	20	2819	GRASSLAND
 Gauchat-1925-480-58	58	to pass	passer	passare	20	558	PASS BY
 Gauchat-1925-480-59	59	to the cross	vers la croix	crŭce	20		
 Gauchat-1925-480-60	60	the well	le puits	pŭteu	20	954	WELL
-Gauchat-1925-480-61	61	the "chable"	le „chable“	catabolu	22
```
AnnikaTjuka, line 62: It might be more consistent if we replace the quotation marks in the french column with << also?

chrzyki: Yep, I think so too. I'll do that in a next step, this PR was just for the (potentially) breaking `"`.

Issue #905, opened 2020-08-07
Missing column in Carling-2019-100


Kristina-Pianykh:
The data in Concepticon seem to lack the column 'Classification' found in the [source list](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0223588). I would vote to add it if it wouldn't have drastic consequences on other data in case the list is used somewhere else.

LinguList:
It doesn't have many consequences, so I think it is nice, if this can be done.

chrzyki:
Done in https://github.com/concepticon/concepticon-data/pull/906.

PR #906, opened 2020-08-07
Add additional column to Carling-2019-99, rename concept list

chrzyki:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [x] retire data

## Additional information

This adds the `CLASSIFICATION` column as outlined in #905. It also renames the list because the last item (`TOTAL`) was, as far as I can tell, not part of the original list but an artifact of the sheet layout.


concepticondata/retired.json
```diff
@@ -28779,6 +28779,506 @@
...
+    },
+    {
+      "id": "Carling-2019-100-99",
+      "comment": "renaming",
+      "replacement": "Carling-2019-99-99"
+    },
+    {
+      "id": "Carling-2019-100-100",
+      "comment": "renaming",
+      "replacement": "Carling-2019-99-100"
```
Kristina-Pianykh, line 29281: Has the policy of retiring concepts that are not in the list changed? I think I used `"replacement": null `once but as far as I can tell it's already outdated.

chrzyki: Nope, I was just missing a commit. Thanks.

Kristina-Pianykh:
The data they provided contain 99 concepts. I honestly have no idea where the 100th concept TOTAL came from in the first version of the list. Here's what the authors says o the semantic categories:

> Our data includes words for predator and game animals, domestic animals, farming, metallurgy, tools, weapons, implements, and so forth. The basis for our list is a selection of core meanings, which are of primary importance to subsistence and economy (see Table 1), and which can be reconstructed to the proto-languages of the families, in particular Indo-European [18, 19]. We have extracted the dataset from the lexical subsection of the database DiACL [20], which uses a taxonomic classification of concepts (including activities, agricultural tools, crops & fruit, farm animals, products, seasons, trees, weapons, wild animals, etc.) based on the standard of the Intercontinental Dictionary Series [4, 21]. For the current study, we use a semantic classification of our data that is based on clustering of colexifications (co-occurrence of meanings of concepts) and meaning changes [22].

chrzyki:
The list was already released so retiring is necessary. Categories come from this supplement: https://doi.org/10.1371/journal.pone.0223588.s003.

The 'TOTAL' came from one of the supplements where it was the last row, being the TOTAL of the cells but not an individual row/concept.

Issue #789, opened 2020-03-10
Refactor concept_set_meta data layout


xrotwang:
- Each metadata set gets it's own directory in `concept_set_meta`.
- References for metadata sets are curated in the global `references.bib`.
- Each metadata set may have code to regenerate the concepticon linking

xrotwang:
superseded by NoRaRe

PR #913, opened 2020-08-18
Move conceptlist metadata to individual folders

chrzyki:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

Only a draft PR for talking about the general idea behind concepticon/concepticon-data#789.


LinguList:
I suppose we do this by including norare for once. Norare is already giving all the structure needed here, and covers most datasets (remaining ones may need to be added for completeness).

LinguList:
I suppose we close this for now, is this okay, @chrzyki? I see it is a draft PR, but it is always giving me the impression that I need to do something. We can tackle the integration of NoRaRe in October, but would need a discussion with @xrotwang, since this will add an additional curation workflow to the whole of concepticon (which is certainly doable, but should be discussed, and maybe better to be done AFTER @AnnikaTjuka and myself have added more datasets for the revision of the norare study).

chrzyki:
Sure, sounds like a plan!

Issue #918, opened 2020-08-30
 Typo in BeijingDaxue-1964-905


MacyL:
please change 閑淡 to 闲谈, and I would suggest the concept mapping changes to CHAT (WITH SOMEBODY) [2903]

LinguList:
Okay, can you make a PR on this?

MacyL:
sure, I will make a PR on Monday :)

chrzyki:
This is done, thanks.

PR #920, opened 2020-08-31
relate to #918

MacyL:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information
relate to issue #918 
...

### DIFF
--- a/concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
+++ b/concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
@@ -525,7 +525,7 @@ BeijingDaxue-1964-905-523	523	摘	262	verb	body-part related actions	三、肢�
 BeijingDaxue-1964-905-524	524	擦	262	verb	body-part related actions	三、肢體動作	動詞	wipe, rub	cā	1449	RUB	擦
 BeijingDaxue-1964-905-525	525	搓	263	verb	body-part related actions	三、肢體動作	動詞	to rub one's hands	cuō			搓
 BeijingDaxue-1964-905-526	526	提	263	verb	body-part related actions	三、肢體動作	動詞	hold (in hand), lift (in hand)	tí	501	LIFT	提
-BeijingDaxue-1964-905-527	527	舉	264	verb	body-part related actions	三、肢體動作	動詞	raise (something with hand)	jǔ	501	LIFT	举
+BeijingDaxue-1964-905-527	527	舉	264	verb	body-part related actions	三、肢體動作	動詞	raise (something with hand)	jǔ	1317	RAISE	举
 BeijingDaxue-1964-905-528	528	托	264	verb	body-part related actions	三、肢體動作	動詞	support, rely on	tuō			托
 BeijingDaxue-1964-905-529	529	端	265	verb	body-part related actions	三、肢體動作	動詞	carry horizontally	duān	2763	CARRY HORIZONTALLY	端
 BeijingDaxue-1964-905-530	530	捧	265	verb	body-part related actions	三、肢體動作	動詞	hold in two hands	pěng	412	CARRY IN HAND	捧
@@ -609,7 +609,7 @@ BeijingDaxue-1964-905-607	607	拌	304	verb	daily actions	四、日常操作	動�
 BeijingDaxue-1964-905-608	608	選擇	304	verb	daily actions	四、日常操作	動詞	select	xuǎn zé	1750	CHOOSE	选择
 BeijingDaxue-1964-905-609	609	藏(收存)	305	verb	daily actions	四、日常操作	動詞	safe (money), collect (money)	cáng (shōu cún)			藏(收存)
 BeijingDaxue-1964-905-610	610	說	305	verb	communication	五、交際、事務、人事	動詞	speak	shuō	1623	SPEAK	说
-BeijingDaxue-1964-905-611	611	閑淡	306	verb	communication	五、交際、事務、人事	動詞	talk, make smalltalk	xián dàn	2708	TALK	闲澹
+BeijingDaxue-1964-905-611	611	闲谈	306	verb	communication	五、交際、事務、人事	動詞	talk, make smalltalk	xián tán	2903	CHAT (WITH SOMEBODY)	闲谈
 BeijingDaxue-1964-905-612	612	問	306	verb	communication	五、交際、事務、人事	動詞	ask	wèn	1309	ASK (INQUIRE)	问
 BeijingDaxue-1964-905-613	613	理	307	verb	communication	五、交際、事務、人事	動詞	manage	lǐ	3073	MANAGE	理
 BeijingDaxue-1964-905-614	614	叫	307	verb	communication	五、交際、事務、人事	動詞	call	jiào	1084	CALL	叫
--- a/concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
+++ b/concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
@@ -609,7 +609,7 @@ BeijingDaxue-1964-905-607	607	拌	304	verb	daily actions	四、日常操作	動�
 BeijingDaxue-1964-905-608	608	選擇	304	verb	daily actions	四、日常操作	動詞	select	xuǎn zé	1750	CHOOSE	选择
 BeijingDaxue-1964-905-609	609	藏(收存)	305	verb	daily actions	四、日常操作	動詞	safe (money), collect (money)	cáng (shōu cún)			藏(收存)
 BeijingDaxue-1964-905-610	610	說	305	verb	communication	五、交際、事務、人事	動詞	speak	shuō	1623	SPEAK	说
-BeijingDaxue-1964-905-611	611	闲谈	306	verb	communication	五、交際、事務、人事	動詞	talk, make smalltalk	xián tán	2903	CHAT (WITH SOMEBODY)	闲谈
+BeijingDaxue-1964-905-611	611	閒談	306	verb	communication	五、交際、事務、人事	動詞	talk, make smalltalk	xián tán	2903	CHAT (WITH SOMEBODY)	闲谈
 BeijingDaxue-1964-905-612	612	問	306	verb	communication	五、交際、事務、人事	動詞	ask	wèn	1309	ASK (INQUIRE)	问
 BeijingDaxue-1964-905-613	613	理	307	verb	communication	五、交際、事務、人事	動詞	manage	lǐ	3073	MANAGE	理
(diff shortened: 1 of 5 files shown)


concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
```diff
@@ -525,7 +525,7 @@ BeijingDaxue-1964-905-523	523	摘	262	verb	body-part related actions	三、肢
 BeijingDaxue-1964-905-524	524	擦	262	verb	body-part related actions	三、肢體動作	動詞	wipe, rub	cā	1449	RUB	擦
 BeijingDaxue-1964-905-525	525	搓	263	verb	body-part related actions	三、肢體動作	動詞	to rub one's hands	cuō			搓
 BeijingDaxue-1964-905-526	526	提	263	verb	body-part related actions	三、肢體動作	動詞	hold (in hand), lift (in hand)	tí	501	LIFT	提
-BeijingDaxue-1964-905-527	527	舉	264	verb	body-part related actions	三、肢體動作	動詞	raise (something with hand)	jǔ	501	LIFT	举
+BeijingDaxue-1964-905-527	527	舉	264	verb	body-part related actions	三、肢體動作	動詞	raise (something with hand)	jǔ	1317	RAISE	举
```
LinguList, line 528: We had long discussions about this. @MacyL, can you ask @Schweikhard to double check?

concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
```diff
@@ -609,7 +609,7 @@ BeijingDaxue-1964-905-607	607	拌	304	verb	daily actions	四、日常操作	動
 BeijingDaxue-1964-905-608	608	選擇	304	verb	daily actions	四、日常操作	動詞	select	xuǎn zé	1750	CHOOSE	选择
 BeijingDaxue-1964-905-609	609	藏(收存)	305	verb	daily actions	四、日常操作	動詞	safe (money), collect (money)	cáng (shōu cún)			藏(收存)
 BeijingDaxue-1964-905-610	610	說	305	verb	communication	五、交際、事務、人事	動詞	speak	shuō	1623	SPEAK	说
-BeijingDaxue-1964-905-611	611	閑淡	306	verb	communication	五、交際、事務、人事	動詞	talk, make smalltalk	xián dàn	2708	TALK	闲澹
+BeijingDaxue-1964-905-611	611	闲谈	306	verb	communication	五、交際、事務、人事	動詞	talk, make smalltalk	xián tán	2903	CHAT (WITH SOMEBODY)	闲谈
```
LinguList, line 612: @MacyL, you realized we use long and short characters in teh two columns, right? I assume that the first character should then be 閒談, right?

LinguList (changes requested):

concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
MacyL: wait, I didn't change this row. I pulled the data from the Master branch, and then change the 闲谈 characters and mappings. I have no idea why this is different. But thanks for this one, I will make another issue about what I have found.

concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
MacyL: ah true, sorry for not noticing it. The column was "Chinese" and "Chinese simplified". So I was a bit confused the column meaning. I will change it now.

LinguList:
I think, @MacyL, you can add the revisions proposed in #921 here, then we have all in one place. Otherwise, you have to also modify the RAISE. And I'd suggest @Schweikhard as another reviewer.

LinguList:
@MacyL, did you follow up on my comment from 14 days ago? I don't see any changes. You should inform me, if you made them, as I am not always notified of this. This PR has been open for too long a time...

MacyL:
Somehow this comments didn't send to my mailbox, sorry. I will do so.

concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
Schweikhard: I wasn't sure whether my input is needed here as this seemed to have been an unintentional change. As the PR is being discussed again - what is the difference in meaning between RAISE and LIFT?
RAISE includes 舉 and 举(~手), whereas 提, 举起, 揭(~盖子), 抬, 揭（盖子） and 举起, as well as raise, lift and some mismappings referring to raising a child or animal (including 养（养鸡）).

So I would guess that LIFT is about taking an object and increasing its altitude, whereas RAISE is about increasing the altitude of a body part like one's arm or hand.

If that's accurate, then this raise (something with hand) here should be changed back to LIFT (and the raise = bring up mappings fixed) and the definitions should be made clearer.

Schweikhard (changes requested):

concepticondata/conceptlists/BeijingDaxue-1964-905.tsv
Schweikhard: If of course both mean "taking an object and increasing its altitude" (as the English translation here would imply), then everything at RAISE should be moved to LIFT and RAISE be retired.

MacyL: I agree that the RAISE should be body part related, and lift is to increase an object's altitude. However, I believe that BeijingDaxue's 舉 is body part related movement, so it should be mapped to RAISE according to the above two conditions? No?

Schweikhard: So the English translation is unreliable?

MacyL: I can't say that the English translation is unreliable, but I personally believe that 提 is more suitable to map to LIFT than 舉 (both were mapped to LIFT).

LinguList: On what grounds, @MacyL, did you check the dialect readings? The English translation was added by myself, using an automated procedure. The Chinese reading is the important one, so if you have better translations, please also modify those.

laiyunfan:
I think:
提: The action increases the altitude of an object, **with the arm naturally positioned hung down**
舉（举）: The action increases the altitude of an object **to somewhere equal to higher than the head**. For example, weight-lifting is translated as 舉重（舉-weight）.

It is for me impossible to fit it in any of the English concepts, or do you have better suggestions after reading my interpretation?

LinguList:
Thanks, @laiyunfan. I suggest, we do not need to link BOTH concepts, but could link 舉（举）to "lift", as this involves the hand?

MacyL:
At this point, I ask for a clear instruction on how should I map these concepts.
1. what about the 提 in this concept? Unmap or keep the LIFT
2. what about the concepts in the issue #921 

if we now change this 舉  back to LIFT, then the issue #921 has no significant improvement. Because the mapping in LIFT includes 舉 提 揭 養。 what I am doing is that these single words does not have a clear English meaning or translation, so I was moving all the 舉 to RAISE, and 提 to LIFT as a way to regulate the Chinese concept mapping.  That's my argument of making these changes, but I am not a concepticon expert, so I need a clear instruction on how should I map these concepts

LinguList:
@Macyl, if there is no clear counterpart, and another translation seems
to be needed, then you should use a different translation and make a new
concepticon concept set.

LinguList:
So what I suppose is then: let us first get back to the original issue and not modify things here. So we leave the lift parts unmapped, but you can correct the other aspect in your modification to Beijing Daxue. Then, when a decision was reached at in the issue, you can make a new issue. And we ask @laiyunfan to help by finding good glosses, if you consider both words not being covered in their basic meaning by any existing conceptticon concept sett.

concepticondata/conceptlists/Castro-2010-540.tsv
```diff
@@ -410,7 +410,7 @@ Castro-2010-540-408	408	to borrow	借	1823	BORROW
 Castro-2010-540-409	409	to owe	欠	369	OWE
 Castro-2010-540-410	410	to exchange	(交)换	2509	EXCHANGE
 Castro-2010-540-411	411	to suck	吸	1421	SUCK
-Castro-2010-540-412	412	to lift (over head)	举起	501	LIFT
+Castro-2010-540-412	412	to lift (over head)	举起	1317	RAISE
```
LinguList, line 413: Leave this as it was.

concepticondata/conceptlists/Sun-1991-1004.tsv
```diff
@@ -590,7 +590,7 @@ Sun-1991-1004-588			588	炼(钢)	refine
 Sun-1991-1004-589	288	FORGE	589	打(铁)	strike (iron)
 Sun-1991-1004-590	2071	LAY (VERB)	590	砌(砖)	lay (bricks)
 Sun-1991-1004-591	765	MOVE	591	搬	move something to a new place
-Sun-1991-1004-592	501	LIFT	592	抬	carry on a pole (two people)
+Sun-1991-1004-592			592	抬	carry on a pole (two people)
```
LinguList, line 593: Leave this as it was.

concepticondata/conceptlists/Yang-2011-1014.tsv
```diff
@@ -561,7 +561,7 @@ Yang-2011-1014-560	560	555	pitch (throw from side way)	摔 (石头)	555	pitch (t
 Yang-2011-1014-561	561	556	toss (overhand)	投	556	toss (overhand)	1456	THROW
 Yang-2011-1014-562	562	557	stretch; stretch out	伸 (伸手)	557,0 / 557,1 / 557,1	stretch; stretch out / stretch; stretch out / stretch; stretch out	214	STRETCH
 Yang-2011-1014-563	563	558	pound	舂/捣碎	558	pound	354	POUND
-Yang-2011-1014-564	564	559	lift (over head)	举起	559	lift (over head)	501	LIFT
+Yang-2011-1014-564	564	559	lift (over head)	举起	559	lift (over head)	1317	RAISE
```
LinguList, line 564: leave this as was or unlink

LinguList (commented):
Please see my comments, @macyl, it seems we may need a "lift over head" concept set. But we want to discuss this in the issue before, not making a half-cooked PR now here. So if you revert those changes, where I indicated, we can merge this PR and pursue the discussion later in the issue, okay?

MacyL:
I have made the changes accordingly. Please have a look

LinguList (approved):

LinguList:
@Schweikhard or @laiyunfan, can you approve, so we can merge?

laiyunfan (approved):

Schweikhard (approved):

Issue #922, opened 2020-08-31
SCREAM vs SHRIEK


Kristina-Pianykh:
`SCREAM (Action/Process)` seems to be a good candidate for adding to Concepticon (with Hale-1973-1798-1194 and Kibrik-2012-122-98) but since we already have SHRIEK I'm not sure if the distinction between these synonyms is important enough or the 'scream' concepts can simply be mapped to SHRIEK.

Issue #902, opened 2020-08-06
WASH (CLOTHES) (3434) and LAUNDER (2503)


chrzyki:
Can those two be merged/one be replaced?

AnnikaTjuka:
Yes, I think they are duplicates. 洗(衣服) is mapped to WASH (CLOTHES) in Castro-2010-511-188 and to LAUNDER in Castro-2015-608-349.  And the translation for _laver (linge)_ is also 'do laundry/wash clothes'.

PR #923, opened 2020-09-02
Modify the scream concept sets, retire WASH (CLOTHES), refine wash-clothes mappings

Kristina-Pianykh:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [x] refine Concepticon glosses
- [ ] retire data

## Additional information

Refers to #922 and #902.
1. I changed the naming of SCREAMING to SCREAM (NOISE),
2. added SCREAM (PRODUCE A CRY) and mapped two lists to it,
3. modified the definition of SHRIEK to distinguish it from SCREAM (PRODUCE A CRY),
4. retired WASH (CLOTHES) as we have already LAUNDER,
5. changed the naming of LAUNDER to LAUNDER (WASH CLOTHES) (I thought it would be easier to find it given a gloss like `wash (clothes`)),
6. refined some Chinese mappings.

Speaking of the Chinese lists, in Sun-1991-1004 the gloss `洗(衣服)` is mapped to LAUNDER/WASH (CLOTHES) in Castro-2010-511 and Castro-2015-608 (see also the Engish translation) so I remapped it accordingly. In Huang-1992-1820 the gloss `洗(~衣)` also seems to mean 'to wash clothes' as in Chen-2012-888, thus the mapping to LAUNDER (WASH CLOTHES).

I assume some of the lists are involved in Lexibank so it would be nice if somebody in the know could let me know if any of the changes I'm proposing could interfere.


LinguList:
Okay, I would like to ask @MacyL and @Schweikhard to check this first to make reviews. I'll moderate it.

Schweikhard (approved):
The remappings and new concepts/definitions look good to me! The distinction between SHRIEK and SCREAM (NOISE) seems helpful.

Is SCREAM (PRODUCE A CRY) the verb to SHRIEK or to SCREAM (NOISE)? Based on the definition, it seems rather like the former, which is a bit strange based on the name.

MacyL:
Most of them looks good to me. A minor comment, I prefer Huang-1992-1820's concept remains as it was. My reason is that the concepts in the brackets usually are the examples to show people when you will use it. And in Chinese "洗衣" “洗手” “洗澡” are all the same "洗". Unless, there is another "洗(~XXX)" in the concept list, I suggest to keep 洗(~衣) as WASH.

Kristina-Pianykh:
Thank you for your comment @MacyL, I'll unmap Huang.

As far as I understood it, the main difference between 'to scream' and 'to shriek' is the pitch (higher in the latter) and often the length of the sound produced (longer in the former). I can add a sentence to the definition of SCREAM (PRODUCE A CRY) to draw a comparison with SHRIEK if it makes the distinction more clear.

SCREAM (NOISE), in the meantime, is a substantive concept and shouldn't be mixed with the verbal counterparts.

LinguList (approved):

LinguList:
Feel free to merge, thanks!

Schweikhard:
> As far as I understood it, the main difference between 'to scream' and 'to shriek' is the pitch (higher in the latter) and often the length of the sound produced (longer in the former). 

I understood it as, SHRIEK is a screamed sound (no words), SCREAM (NOISE) is screamed words. But the definitions aren't completely clear.

Issue #927, opened 2020-09-14
3715 BELLYBUTTON and 1838 NAVEL


AnnikaTjuka:
Only one item (Ombligo [spanish]) is mapped to 3715 BELLYBUTTON. I guess the two concepts are duplicates. If @Schweikhard, @Kristina-Pianykh, @CarolinHu, and @LinguList agree, I'll retire 3715 BELLYBUTTON.

Schweikhard:
I agree.

CarolinHu:
Me too, they very much look like duplicates.

Kristina-Pianykh:
I would definitely retire BELLYBUTTON.

Issue #924, opened 2020-09-03
471 UPPER LEG and 800 THIGH


CarolinHu:
Based on the concept descriptions of those two sets and the fact that UPPER LEG only has two mappings which may be subsumed under THIGH even though their forms are different, I think one could be retired?

LinguList:
I would think so, what do the others think, @Schweikhard, @AnnikaTjuka and @MacyL and @Kristina-Pianykh ?

AnnikaTjuka:
@CarolinHu Thanks for pointing this out! I'm in the process of setting up a list of body parts in Concepticon. I'll put it on my list of possible changes.

Kristina-Pianykh:
I'm pretty sure they are duplicates. It looks like 'upper leg' is not a scientific term but rather just a common name. See also some elaborations [here ](https://www.healthline.com/health/human-body-maps/leg#knee)and [here](https://www.medicinenet.com/script/main/art.asp?articlekey=8456).

MacyL:
> Yes that very much looks like a duplicate to me.

agree.

AnnikaTjuka:
Looking at the mappings, we have 2 lists mapped to UPPER LEG and 49 lists mapped to THIGH. UPPER LEG (471) is the older concept so that THIGH (800) will be retired. 

I'm going to remap the 49 lists with THIGH to UPPER LEG.

@LinguList Is it ok, if I change the CONCEPTICON_GLOSS to UPPER LEG (THIGH) to prevent any further confusion.

LinguList:
But I think it is a good idea.

AnnikaTjuka:
Based on similar PRs (e.g., #923), retiring and renaming of concepts does not seem to involve the retire.json. I'd rename the concept gloss and remap the lists to the refined concept 471 UPPER LEG (THIGH). I'll get to work and will open a PR soonish.

PR #931, opened 2020-09-16
Retire concepts

AnnikaTjuka:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [X] refine existing Concepticon concept set mappings
- [X] refine Concepticon glosses
- [X] retire data

## Additional information

This PR solves #927 and #924 .


AnnikaTjuka:
For BELLY BUTTON  and NAVEL, I retired the former and remapped Izura-2005-499 to NAVEL.
For UPPER LEG and THIGH, I retired THIGH, renamed the concept gloss, and remapped the lists to the refined concept 471 UPPER LEG (THIGH).

Note that there is one list that uses both glosses: _thigh_ and _upper leg_ (Zgraggen-1980-380). I mapped both to  UPPER LEG (THIGH), but maybe only one should be mapped?

AnnikaTjuka:
@Schweikhard Would you mind moderating this PR?

LinguList:
Well, in fact, I did not know that there was a list that distinguishes the two, but this means in fact that we should not have merged the two concepts. The rule in concepticon is: if there is one concept list distinguishing two concepts, we accept the distinction. But on the other hand: it still seems to be synonymous. Can one of the lexibankers have a look at lexibank/zgraggenmadang and try to find out what is going on there? Are there really two different concepts in the list? @MacyL, could you do that?

Schweikhard:
I can moderate - and based on the discussion started, I suggest @LinguList and @MacyL as reviewers.

MacyL:
sure

LinguList (approved):
Okay, I think it is better to unlink Zgraggen "upper leg" and leave only "thigh". Otherwise okay.

MacyL:
A quick check on zgraggenmadang repository. Although, the _upper leg_  is included in the conceptl-ist "Zgraggen-1980-380", it is not shown in the raw data, and it is not used by the repository, eitehr (checked the parameters.csv, the forms.csv, and the script). 
For some languages, there are two words for "leg", I wondered whether the authors were using upper leg to distinguish the two words.

LinguList:
Okay, fine, this shows that the confusion is in the concept list itself ;)

AnnikaTjuka:
Thanks for checking @MacyL ! I'll change the mapping according to @LinguList comment then.

AnnikaTjuka:
I unmapped _upper leg_ in Zgraggen-1980-380. @Schweikhard would you mind taking a final look, so that I can merge the PR?

Schweikhard (approved):
Looks good to me!

AnnikaTjuka:
Great! Thanks, everyone!

PR #968, opened 2020-10-16
added Heggarty-2019-170 used for iecor

Bibiko:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data


LinguList (commented):
Wait, please, we have already mapped the Heggarty List as Heggarty-2020-200. We quote Scarborough, etc. Why does this list now have 170 items? I'd rather add a flag on the 200 item list, so we have all of them in the list, as we may want to use the full data, when more is available. 

https://github.com/concepticon/concepticon-data/blob/master/concepticondata/conceptlists/Heggarty-2020-200.tsv

LinguList (commented):
Instead of the description.md, I'd also rather propose a link to the wiki, if it is online. Furthermore, since the Wiki seems to be highly formalized (illustrative context, Target sense, etc.), if to include in our TSV files in concepticon, this should be separated and put in a single cell to make it more readable.

xrotwang:
I'd say it does make sense to add the second list if the descriptions have changed. Otherwise I'd agree that adding flags like `jena170` or similar to the old list is more appropriately modeling the process.

xrotwang:
Ah, I see the old list doesn't have the descriptions. So I'd say we just augment the old list with descriptions and `jena170` flag.

LinguList:
Yes, augmenting also makes this more consistent, as I'd say the list is
better given as the 200 item list, as in Scarboroughs paper, etc.

Bibiko:
For the analysis the number of meanings was decreased to 170. The underlying IE-COR lexibank CLDF dataset lists these 170 items only. Since it a real subset of Heggarty-2020-200 one could of course this one. Regarding to description_md, the data are living in a repo but I'd argue to spilt it into two new columns `illustrative_context_md` and `target_sense_md` and to remove the column `example_content` (which is the same as illustrative_context)

Bibiko:
but the remaining 30 in Heggarty-2020-200 have no or rudimentary Markdown descriptions

xrotwang:
I think it's ok if there's no description for the non-Jena170 items - in particular if there's also the flag signaling this difference.

LinguList:
And it is important for documentation which concepts got cut.

In the list by Sagart
(https://github.com/concepticon/concepticon-data/blob/master/concepticondata/conceptlists/Sagart-2019-250.tsv)
for example, we used concepts with 0.8 coverage and more, as an
objective criterion, resulting in 180 concepts used in the study.

Bibiko:
in other words, it's better to add the markdown stuff to Heggarty-2020-200 and in addition to add a new column `iecor170` which is true for meanings in iecor?

xrotwang:
exactly.

LinguList:
Yes, but I'd be against complete markdown, as this is really not easy to
handle, if you have \n etc. in there, but rather have normal sentences
in the descriptions. Otherwise, I'd favor a link to the wiki on github.

Bibiko:
ok - I'll do - the wiki lives at https://github.com/lingdb/CoBL-public/wiki - thus I'll leave `illustrative_context` as plain text, add a link to this wiki, and add the column `iecor170`

xrotwang:
But how persistent is the wiki? We should definitely include the full wiki pages here https://github.com/lexibank/iecor/blob/master/cldf/parameters.csv when releasing the lexical data. But we cannot link to this right now.

xrotwang:
Btw. that's what we'll do for grambank - i.e. put the wiki pages into `parameters.csv` - but arguably Grammaticon - or in this case Concepticon - would be the place where I'd expect this info.

LinguList:
It is probably useful to have the info, but not in markdown text, right?
I mean, can't one extract the content per field in a manner similar to
the column "notes" we write in conceptlists.tsv ? We have precedents for
this in Concepticon with the list by Kassian et al. (110), as Heggarty
et al took this technique from them.

xrotwang:
Ah ok. Yes, putting sections of the markdown doc into separate columns makes sense.

xrotwang:
@Bibiko  see https://github.com/concepticon/concepticon-data/blob/master/concepticondata/conceptlists/Kassian-2010-116.tsv

Bibiko:
ok - one can also convert Markdown to HTML to have it more readable(?)

LinguList:
CLLD handles markdown in rendering, as far as I understand, so
constructions like `[links](http...)` should work, but I would ask to
delete newlines and not use the `###` headings, as they won't be needed.

Bibiko:
yes - just doing so

xrotwang:
Yes, markdown is better than HTML. Why delete newlines, though? These are significant in markdown, e.g. to mark blocks, etc. Yes, that makes looking at the CSV file harder - but then you can just do `csvcut -t -C description_md ...tsv` and filter out the markdown.

Bibiko:
@LinguList Where does this list come from? The IE-CoR meanings differ slightly regarding to their mappings to the Concepticon codes as for HAIR (IE-CoR: 1040 vs. Heggarty-2020-200: 2648), HIT (1433 vs. 353), LIVE (2309 vs. 1422), SHADOW (2275 vs. 1388), THIGH (800 vs. 471), TURN (1588 vs. 1445). What can be done here?

In addition the column `NUMBER` in Heggarty-2020-200 does not map against IE-CoR IDs, which then should be added as new column, I think.

LinguList:
It makes it a horrow to check for us (!). And as I suspect, these are
mostly one-line sentences, if it is a novel, I'd say they should NOT be
placed into concepticon, but rather referenced somehwere else.

xrotwang:
I think number comes from Matthew's paper.

Hans-Jörg Bibiko <notifications@github.com> schrieb am Fr., 16. Okt. 2020,
19:02:

LinguList:
@bibiko, we used the PAPER by Scarborough, as this is an offical
reference. If they don't manage to keep things in line there, it is not
on us. Concepticon mappings past our review, so I trust our mappings,
not the ones done by the COBL people.

Bibiko:
But as far as I know the CoBL IE meanings are based initially on Scarborough's list, but they developed it further (and Matthew Scarborough is one of the main authors of IE-CoR). CoBL has now 347 meanings. Is this not an indicator that IE-CoR should have its own concept list?

SimonGreenhill:
Yeah, this is the list for the forthcoming Indo-European paper (Heggarty 2020 I guess now, unless it takes even longer). Paul and Cormac removed the extra words as being complicated.

LinguList:
In that case, @Bibiko, if we discuss forthcoming papers, I propose to NOT include them now, but wait for their publication, then we can do this from scratch and leave other lists untouched. It does not make much sense otherwise, since they have shifted and cut their list so many times that it is impossible to decide which is the correct list. So better: if there's a paper, we link the list, but we only start doing that if the paper is accepted and we have proofs to compare.

PR #986, opened 2020-10-28
Refine SLED definition, remap BirketSmith

### LINKED ISSUE #953: SLED (16) vs SLEDGE (368)
I see that we have retired SLEDGE in favour of SLED. If we only want to use SLED (and not differentiate in the glosses between 'small sleds' and 'larger sledges') I think we should refine the description of SLED.

chrzyki:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

Closes #953.


CarolinHu (approved):
All good, thanks for adding!

PR #992, opened 2020-10-29
Update mappings

chrzyki:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

We should probably refine more of the WALL mappings before release. I think most of the remaining ones (except of course for the very explicit WALL OF HOUSE glosses) should be changed to WALL. What do you think, @LinguList?


concepticondata/conceptlists/Dawson-1887-731.tsv
```diff
@@ -157,7 +157,7 @@ Dawson-1887-731-155	155	Mat	195	MAT	Dwellings	4	29
 Dawson-1887-731-156	156	Bed	1663	BED	Dwellings	4	29
 Dawson-1887-731-157	157	Floor	1632	FLOOR	Dwellings	4	29
 Dawson-1887-731-158	158	Ceiling			Dwellings	4	29
-Dawson-1887-731-159	159	Wall	933	WALL (OF HOUSE)	Dwellings	4	29
+Dawson-1887-731-159	159	Wall	3830	WALL	Dwellings	4	29
```
CarolinHu, line 160: Could this be indeed WALL (OF HOUSE) as opposed to the ceiling above?

CarolinHu (approved):
I can't speak for the Chinese mappings, but the rest looks good.
Regarding your proposal to remap almost all of them, does it make sense to add a relation WALL narrower WALL (OF HOUSE)?

chrzyki:
> I can't speak for the Chinese mappings, but the rest looks good.
> Regarding your proposal to remap almost all of them, does it make sense to add a relation WALL narrower WALL (OF HOUSE)?

Sense relations in general would be good to have for the new concepts. Do you maybe have time to look at this?

concepticondata/conceptlists/Dawson-1887-731.tsv
chrzyki: In this instance in can indeed be both (just going by the lexemes), so the more general mapping is okay I'd say.

CarolinHu:
Yes, I'll have a look.

PR #994, opened 2020-10-29
Some concept relations

CarolinHu:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

Relations for some recently added concepts. There are 3 more where I wasn't sure whether they would be suitable, so I left them out for now:

- STONE OR ROCK narrower/instance of BOULDER ?
- PUT narrower PUT (IN A SITTING/STANDING POSITION) - here I am hesitating because PUT only makes reference to moving objects while the newly added PUTs also include "moving oneself"
- FEELING HOT and FEELING COLD - is there any suitable relation to our normal concepts of HOT or is the "feeling" part too specific as it has, on the surface, nothing to do with observable temperatures?

Welcoming a review from whoever has time now.


concepticondata/conceptrelations.tsv
```diff
@@ -770,3 +770,12 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
 3805	SHY OR ASHAMED	narrower	1894	ASHAMED
 3809	SCREAM (PRODUCE A CRY)	producedby	2522	SCREAM (NOISE)
 298	SHRIEK	similar	3809	SCREAM (PRODUCE A CRY)
+3213	HIRE OR RENT	narrower	3831	RENT (SOMETHING)
```
AnnikaTjuka, line 773: We would also need to add a relation for 856 HIRE, right?

concepticondata/conceptrelations.tsv
```diff
@@ -770,3 +770,12 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
...
+947	STREAM	narrower	3821	BROOK
```
AnnikaTjuka, line 774: STREAM is already an instance of 3311	FLOWING BODY OF WATER. Wouldn't this also be the relation for BROOK?

concepticondata/conceptrelations.tsv
```diff
@@ -770,3 +770,12 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
...
+1208	CAT	narrower	3825	MALE CAT
+1936	DEER	instanceof	3826	ROE DEER
+832	BEAN	instanceof	3839	BROAD BEAN
+3830	WALL	narrower	933	WALL (OF HOUSE)
```
AnnikaTjuka, line 778: Is it `narrower` or `instance of`?

concepticondata/conceptrelations.tsv
```diff
@@ -770,3 +770,12 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
...
+1123	PREGNANT	narrower	3827	PREGNANT (OF ANIMALS)
```
AnnikaTjuka, line 779: Is it `narrower` or `instance of`?

AnnikaTjuka (commented):
These relations are always tricky. I left some questions, maybe @Schweikhard could have a look as well.

concepticondata/conceptrelations.tsv
CarolinHu: Thanks for leaving a review! That relation (HIRE OR RENT narrower HIRE) already exists in l. 623.

concepticondata/conceptrelations.tsv
Schweikhard: The difference between RIVER, STREAM and BROOK seems fuzzy anyway... But I guess a RIVER is defined by not flowing into another FLOWING BODY OF WATER, and a BROOK is a small STREAM. Thereby, a BROOK is a kind of STREAM.

concepticondata/conceptrelations.tsv
CarolinHu: Here I wasn't sure about the distinction between those relations either.

Schweikhard: I guess narrower because a Mauer and a Wand, while having many similarities, are in some sense an artificial grouping of two different things. But I could argue either way.

concepticondata/conceptrelations.tsv
Schweikhard: I can see the point, but with adjectives, that's hard to decide...

concepticondata/conceptrelations.tsv
AnnikaTjuka: Ah yes!

concepticondata/conceptrelations.tsv
CarolinHu: Here I would really go for narrower. Instanceof sounds like animal pregnancy is one of different types of pregnany when it might just be human convention to have separate words for humans and other mammals (with the biological background being pretty much the same).

Schweikhard (commented):
I kinda try and keep away from the question whether something's narrower or instanceof ;) In general, I would tend to use instanceof as a matter of principle, as narrower is there mainly for technical purposes. I'd only clearly use narrower if the concepts in the hyponym are really wide apart, referring to two things that you would normally not group together.

LinguList (approved):
I am fine with the discussions presented here. If they don't break the tests, this is also showing that nothing serious has happened so far!

CarolinHu:
Thanks all for the comments! I'll go ahead and merge then.

PR #1003, opened 2020-11-09
Three new concepts and mappings

CarolinHu:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR adds CRANBERRY, WINDMILL, and ONE THIRD as new concepts;
and also maps two concepts in Hale 1973 to the already existing GRAVEYARD and PEACOCK.
As always I'm happy for pointers to those new concepts in languages I might have missed!

I haven't done this yet because I wanted to get some feedback first, but I think there might be a case for remapping Buck-1949-1110-737 third (sb., fraction) from the current THIRD to the new ONE THIRD. The gloss speaks of a fraction, and Buck-1949-1110-736 is also mapped to THIRD, making it a double mapping at the moment.

(Sorry about the number formatting in Nagano, will fix this in a minute. I also already added BUT as a new concept but then decided against mapping lists in this PR because there are quite a few and it would get a bit confusing, so these mappings will be added later in a new PR if that's okay.)

@chrzyki Do you have time for a review?


chrzyki:
Thanks @CarolinHu! I'll review once the number formatting has been taken care of.

chrzyki (approved):
CRANBERRY: Zalizniak-2020-2591
WINDMILL: Monnier-2014-1031

I agree with Buck (but a second opinion might be good here).

Doing the `BUT` mapping in a separate PR sounds reasonable.

Thanks!

CarolinHu:
Thanks for reviewing and spotting Monnier!
I left Zalizniak unmapped because due to the specification of the botanical name, I think they were rather going for the whole plant, not only for the berry. There are several instances of this further down in the list.

chrzyki:
> Thanks for reviewing and spotting Monnier!
> I left Zalizniak unmapped because due to the specification of the botanical name, I think they were rather going for the whole plant, not only for the berry. There are several instances of this further down in the list.

Ah, makes sense! Good thing that you spotted that.

CarolinHu:
Sorry, the tests were taking a bit and then I had to leave. Is this ready for merging?

chrzyki:
Yep!

Issue #1013, opened 2020-12-29
Study of Sixteen Facial Expressions World-Wide


LinguList:
This study lists the following facial expressions:

1. Amusement
2. Anger
3. Awe
4. Concentration
5. Confusion
6. Contempt
7. Contentment
8. Desire
9. Disappointment
10. Doubt
11. Elation
12. Interest
13. Pain
14. Sadness
15. Surprise
16. Triumph

It may be useful to add these to Concepticon, as they are in some sense "linkable", and may help us to integrate more data.

The study is here: https://www.nature.com/articles/s41586-020-3037-7

PR #1014, opened 2021-01-03
Cowen

Kristina-Pianykh:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

#1013. A tiny list of emotive concepts. I added a couple of new concept sets that are going to be linked to some other lists in concepticon in the next PR.

@LinguList would you be so kind to have a quick look at it if you have a minute?


concepticondata/conceptlists.tsv
```diff
@@ -353,3 +353,4 @@ Zalizniak-2020-2591	Zalizniak, Anna A. and Smirnitskaya, Anna and Russo, Maksim
...
+Cowen-2020-16	Cowen, Alan S. and Keltner, Dacher and Schroff, Florian and Jou, Brendan and Adam, Hartwig and Prasad, Gautam	2020		16	specific	English	Arabic, English, French, German, Indonesian, Italian, Japanese, Korean, Mandarin, Polish, Portuguese, Russian, Spanish, Thai, Turkish, Vietnamese	https://doi.org/10.1038/s41586-020-3037-7	Cowen2020		This list of 16 emotions emerged as a result of a machine-learning analysis of large-scaled video material from 144 countries. The purpose of the study was to find out what facial expressions are associated with various naturalistic social contexts and determine culturally universal patterns of context–expression association.
```
LinguList, line 356: If all these languages are mentioned as "target languages", we need them also in our concept list. If the data cannot be found in the supplement anywhere, the additional languages should be discarded.

LinguList (commented):
Thanks a lot for taking care of it. The only objection I have is the comment I mentioned.

I also wonder if we should consider adding also the other missing concepts?

concepticondata/conceptlists.tsv
Kristina-Pianykh: As far as I understood, the list itself is made only in English while the material analyzed was derived from the videos in the languages listed. But I can see now that they should be removed from here. Shall I then just indicate 'global' in the 'target languages'?

Kristina-Pianykh: And I'll add the rest of the concept sets missing here.

LinguList:
I was confused by teh columns: if it is not "source language", there won't be an error, so "target languages" as "global" is probably best, as this is a supposedly universal test.

LinguList (approved):

PR #1019, opened 2021-01-07
Add Database of paradigmatic semantic relation pairs for German

AnnikaTjuka:
# Pull request checklist

- [X] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR adds the database of paradigmatic semantic relation pairs for German nouns, verbs, and adjectives. I had to merge and edit the datasets to make them fit the Concepticon structure. The list includes various multi-mappings. The list is based on a list of target words that were compiled for two experiments: 1. word naming (column WORD_RESPONSE) and 2. similarity rating (column SIMILARITY_MEAN). Only the target words (column GERMAN) were considered for mapping.


AnnikaTjuka:
[Concepticon-check.txt](https://github.com/concepticon/concepticon-data/files/5781869/Concepticon-check.txt)

AnnikaTjuka:
The list is super long but there are actually not that many mappings because the words occur multiple times (see description above).

AnnikaTjuka:
@LinguList Do you have time to moderate this PR?

LinguList (commented):
sorry I did NOT comment on this earlier, but the inherent network structure of this list is now not evident, right? I mean, you have the response, but the response can itself not function as a headword or a word/concept that could be linked up. So I wonder, if one should not better re-structure the list, similar to the way we did this in multi-simlex? I'd also check the list for potential redundancies, and given that the list is so long, I'd at a last point check if the list is not better listed as part of norare?

LinguList:
That to say: I probably did NOT pay enough attention to the way you described that you wanted to structure the data in our email conversation. That is a bit unfortunate, sorry, as I should've mentioned this, as I see now.

AnnikaTjuka:
Yes, that was the point I was trying to make. But the data is still different from the Multi-SimLex data in that the target words were specifically defined at the outset of the study. In Multi-SimLex no target words were defined (thus making the word pairs a bit more arbitrary and they can be used vice versa).

AnnikaTjuka:
The representation in Multi-Simlex wirh "word1" and "word2" could be adapted to "target" and "response"? But the relations (antonym, hyponym, and synonym) would also need to be accounted for.

AnnikaTjuka:
Not sure, how to put all these relations into one structure/list.

LinguList:
Taking "sweet":

NUMBER | GLOSS | GERMAN | LINKS | RATINGS | RELATIONS | POS
--- | --- | --- | --- | --- | --- | ---
1 | sweet | süß | sauer bitter | 9.6 7.4 | ANT ANT | ADJ 
2 | sour | sauer | süß | 9.6 | ANT | ADJ
3 | bitter | bitter | süß | 7.4 | ANT | ADJ

Question is of course: what do you select as separator? I recommend space if this is possible.

Alternatively, you can sub-separate data:

NUMBER | GLOSS | GERMAN | LINKS | POS
--- | --- | --- | --- | --- 
1 | sweet | süß | sauer:9.6:ANT bitter:7.4:ANT | ADJ 
2 | sour | sauer | süß:9.6:ANT | ADJ
3 | bitter | bitter | süß:7.4:ANT | ADJ

LinguList:
In any case: since this is a network, and there are repeated entries, it seems to be the best to 

1. extract each German word and give it an ID
2. model the network

If you want to, I would -- given the importance -- plan a blog post for this month that discusses how to display and parse in code networks in Concepticon. 

SimLex is an exception, given that they have an artificially larger semantic space due to individual translation decisions, so they force us to list a concept like "agree" multiple times, although it is not clear if this is really multiple senses.

LinguList:
Looking at simlex again (our google sheet), I think now we should exclude Kiswahili and run it again, since this will further reduce the number of concepts.

AnnikaTjuka:
Ok, I'm going to need a little time to better understand this before I can start restructuring the list (a blog post is a great idea and would be very helpful!). I'll close this PR for now.

PR #1024, opened 2021-01-11
patches zalizniak

### LINKED ISSUE #1023: Wrong links in Zalizniak
The list seems to have wrong links. https://github.com/concepticon/concepticon-data/blob/master/concepticondata/conceptlists/Zalizniak-2020-2591.tsv

The numbers do not reference to the correct concepts. (in SHIFTS). The SHIFT URLS are fine. So only those need to be adjusted.

LinguList:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

refine errors in a concept list fixes #1023 

...


LinguList:
@AnnikaTjuka, note that the tag "historical" was removed here, since the list is not "historical", i.e., made from before 1900. I forgot to check this earlier, when the list was originally submitted.

Note also that we don't need to officially "retire" the concepts, although the IDs changed, as this has not been released.

Note also that the NOTE now contains a detailed explanation of the network that underlies the data.

LinguList:
@xrotwang, one quick question regarding the newtork here: we have a field NUMBER and items of this field recur space-separated in the field LINKS. Can this be somehow explicitly modeled in CSVW? Because if this WAS possible, we could handle networks more regularly. But otherwise, I think the current representation should be also clear enough.

AnnikaTjuka (approved):
Ah ok, I see. Thanks for updating the list!

Issue #1018, opened 2021-01-06
ITCH 148 declared as n. but lists a number of verbal concepts


Kristina-Pianykh:
There are currently quite a number of lists mapped to this concept set. If we decide to split it into substantive and verbal 'itch' concept sets, all of the lists have to be examined thoroughly to determine what specific meaning of 'itch' they refer to.

Kristina-Pianykh:
I would suggest adding the concept ITCH (CAUSE ITCHING) for verbal concepts that are currently mapped to ITCH (`Person/Thing`) (and?) ITCHY (`Property`). There is quite a number of Chinese 痒 currently mapped to ITCHY that have several English verbal translations and should not be moved, I assume. Since there are also some verb-like concepts in English (to itch/to be itchy/feel itchy (v)) and German ('Es juckt mich.'), I am not sure if ITCHY was initially meant to cover both property and action concepts and thus should not be thought of in terms of a word category, roughly speaking. This brings me to the question if there is a need to split ITCHY and move some of the unambiguous action-concepts to ITCH (CAUSE ITCHING).

Could I ask you, @LinguList, @CarolinHu, @AnnikaTjuka, @Schweikhard, to have a brief look at the concepts and share your opinion?

LinguList:
Yes, it seems useful to remap the concepts. The Chinese case is tricky, as it is definitely a noun, but can also function as verb. So I'd follow English glosses in identifying the mapping here.

AnnikaTjuka:
So we would have the following mappings:

CONCEPTION_GLOSS | Mapping
--|--
ITCH (Person/Thing) | the itch
ITCHY (Property) | itchy
ITCH (CAUSE ITCHING) (Action/Process) | to itch, to be itchy, feel itchy (v), 'Es juckt mich.'

There are plenty of glosses with only "itch". It might be useful if you list them here so that we can decide whether the lists need to be remapped before you open a PR.

Kristina-Pianykh:
Here is the table of all the lists with the ambiguous reference to 'itch':
| List      | ID           | GLOSS|Comments
| ------------- |:-------------:| -----|-------|
| Bodt-2019-664 |Bodt-2019-664-254| itch| I checked the publication and couldn't find where the data come from
[...]

PR #1028, opened 2021-01-14
Add ITCH (CAUSE ITCHING OR FEEL ITCHY) and remap 'itch' glosses

Kristina-Pianykh:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information
#1018. I added the concept set ITCH (CAUSE ITCHING OR FEEL ITCHY) (in the broadest sense of the verb) and remapped the lists with 'itch' glosses.
The only thing I missed while preparing a preliminary table of ambiguous glosses in #1018 is Sun-1991-1004-900: 痒 [chinese]; itch [english]. I could not disambiguate the meaning so I left it mapped to ITCH.


Kristina-Pianykh:
@AnnikaTjuka would you have time to moderate this?

AnnikaTjuka:
Sure! Based on our discussion in #1018: @LinguList Would you mind taking a look at the remappings?

LinguList (approved):

AnnikaTjuka (approved):
Thanks, @Kristina-Pianykh! I would leave the Sun-1991-1004-900 mapping as is if we can't disambiguate it.

LinguList:
All Chinese concepts could in fact be added to a NEW concept set, that would be `ITCH OR ITCHING`, and the definition would hint that this can either be the process that causes itching or the result of that process, both verbal and nominal. 

In the section in Sun, the concepts selected in this chapter are all adjectives or adjective like, but they have these typical adjective-noun transitions, comprising concepts such as "hungry", "thirsty", etc. So I reckon the intended concept that they want to annotate here is indeed "itchy", the adjectival meaning, and I'd even say for this reason one should replace the PAIN mapping with "PAINFUL", since 痛 901 can also have an adjectival or a verbal reading.

Kristina-Pianykh:
The problem with the Chinese concepts is that, judging by the English translations, they can refer not only to the action of the result of itching but also to the property (i.e. 'itchy'). See the following examples:

| ID      | GLOSS           | 
| ------------- |:-------------:|
|Lama-2012-301-279 | 痒 [chinese]; Itchy [english]
|Castro-2010-540-463 | 痒 [chinese]; itchy [english]
|SatterthwaitePhillips-2011-423-188 | 痒 [chinese]; itchy [english]
|Wang-2015-62-53 | 痒 [chinese]; itchy [english]

Although formally, it is still one and the same Chinese word.

LinguList:
Yes, this is typical for Chinese, that the boundaries between word classes are not that sharp. But you have compounds like "scratch itch" that show that the "itch" is rather to be read as a noun. One could expand the category by saying "ITCH OR ITCHY OR ITCHING", which would then include all kinds of itch.

Kristina-Pianykh:
Alright, then I'll add the new concept set ITCH OR ITCHY OR ITCHING to reflect the ambiguity of the Chinese concepts in this case.

LinguList:
Yes, and adding relations to the three ITCHes would also be useful 
(sorry that we didn't decide to do this earlier....)

LinguList (approved):
Thanks a lot!

PR #1032, opened 2021-01-17
update URLs in Zalizniak add Gerardi

LinguList:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This concept list does not really need a moderator, as it has already been discussed before, so I'll ask for any concepticon participant to approve it, or to also check, if you would rather do that.


concepticondata/conceptlists.tsv
```diff
@@ -1,4 +1,5 @@
 ID	AUTHOR	YEAR	LIST_SUFFIX	ITEMS	TAGS	SOURCE_LANGUAGE	TARGET_LANGUAGE	URL	REFS	PDF	NOTE	PAGES	ALIAS
+Gerardi-2021-414	Gerardi, Fabrício and Stanislav Reichert and Coelho Aragon, Carolina	2021		414	basic	English,Portuguese	Tupían languages	https://tuled.org/parameters	Gerardi2021		This list was compiled for a large lexical database of Tupían languages and beyond. The study we provide as reference is not identical with the concept list which we present here, as it belongs to a later version that was consistently mapped to the Concepticon. As a result, the number of items differs from the number mentioned in the study presenting the database.	1-19
```
chrzyki, line 2: I think I'd prefer this to be added to the end of the list, not the beginning. Not that it hugely matters, but it might save us from problematic merges down the road.

chrzyki (approved):

PR #1033, opened 2021-01-21
New concepts from Snider-Anonby

CarolinHu:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR adds 5 new concepts, mainly derived from Anonby and Snider, and maps them in other lists as well. I'm aware of the Hale "TRUE/FALSE" change and will correct it with the next commit. 

Regarding POSSESSIONS: Lynott, Buck, and Hale list "property" as a concept but I wasn't able to determine whether they meant POSSESSIONS or something like BUILDING, so I left them unmapped.

Regarding FAIL: The concept appears three times in Hill (SimLex). @AnnikaTjuka can you help me out what the policy on mappings was with this list? And would you take a look at the PR since you're already familiar with Anonby? :)


AnnikaTjuka:
In the Hill list, glosses can be mapped multiple times so you can add the mappings of FAIL to all occurrences of "fail."  The Anonby mappings look good :)

CarolinHu:
Thank you! I added the concept in Hill now.

AnnikaTjuka (approved):
Thanks! Because this PR includes new concepts, I think @LinguList should also have a look.

LinguList (approved):

Issue #836, opened 2020-05-05
General concept for BLOW


AnnikaTjuka:
Just saw that we have other Spanish glosses mapped to either BLOW (OF MOUTH) or BLOW (OF WIND). We should consider opening a general BLOW concept. 

_Originally posted by @AnnikaTjuka in https://github.com/concepticon/concepticon-data/pull/834/review_comment/create_

AnnikaTjuka:
We have 102 lists mapped to BLOW (OF WIND) of which most have the gloss 'blow'. In my view, it does not make sense to open a new general BLOW concept now. This would mean that we'd have to transfer too many lists (of which probably some are rightfully mapped to BLOW (OF WIND)).

However, there are some glosses of other languages which do not seem to differentiate between [BLOW (OF WIND)](https://concepticon.clld.org/parameters/175) or [BLOW (WITH MOUTH)](https://concepticon.clld.org/parameters/176): _soplar_, 吹 and дуть

I guess we have the following options:
- Indicate a relation between the two concepts @Schweikhard 
- try to map the three glosses (Spanish, Chinese, Russian) consistently to either one of the concepts @LinguList @MacyL 
- leave it as is

LinguList:
/soplar/, 吹 and дуть are all ambiguous, I'd say, pusten or wehen.

Schweikhard:
I would nevertheless argue for a general concept - it could receive the current definition, and we would move anything there that is not clearly the wind blowing (in a lot of cases it is clear).

We _could_ also treat the current concept as the hyperonym as we seem to be doing currently, but it kinda has the wrong name for that.

AnnikaTjuka:
True, having  BLOW (OF WIND) as a hyperonym is not the most elegant way. But I fear if we make a huge remapping (probably 50 lists) to a new BLOW concept we lose hand-curated instances where a list with 'blow' is rightfully mapped to BLOW (OF WIND).

LinguList:
we could say: blow prototypically means "blow (of wind)", but where we
know, we link it to other concepts. We can also say: the three terms
mean the prototypical wind-blow (in most concept lists, I think this is
what they mean). Having a meta-concept is probably useful, but
re-linking all concepts may be less useful, yes.

AnnikaTjuka:
[...]

PR #1035, opened 2021-01-25
Wittmann-1973-200

CarolinHu:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

A rather simple concept list, based on Swadesh. @Schweikhard would you mind moderating?


Schweikhard:
Sure! That seems like a very standard list, perhaps we only need one reviewer? @AnnikaTjuka would you review?

LinguList (commented):
This is a very interesting list. However, in my opinion, the cognate information should go into a lexibank dataset and can be discarded here. In order to preserve your nice work, I'd recommend to open an issue on github.com/lexibank/lexibank and paste the conceptlist in the form of a txt file or as a zip file there. From there, we can then turn this into a lexibank dataset.

For the concept list, I'd therefore discard the columns indicating cognacy.

chrzyki:
@CarolinHu I've invited you to lexibank/lexibank so that you can open issues there.

CarolinHu:
Thank you @chrzyki and will do @LinguList!

concepticondata/conceptlists/Wittmann-1973-200.tsv-metadata.json
```diff
@@ -0,0 +1,66 @@
...
+                    {
+                        "datatype": "string",
+                        "name": "MAURITIAN"
+                    },
+                    {
+                        "datatype": "string",
+                        "name": "HAITIAN"
+                    },
+                    {
+                        "datatype": "string",
```
LinguList, line 55: @CarolinHu, you can be even a bit more explicit by adding the glottocodes in these cases, e.g., for Portuguese, you could write:

```
                    {
                        "datatype": "string",
                        "propertyUrl": "https://glottolog.org/resource/languoid/id/port1283",
                        "name": "PORTUGUESE"
                    },
```

We have not been doing this consequently, but it may be helpful for completeness.

LinguList (approved):

concepticondata/conceptlists/Wittmann-1973-200.tsv
```diff
@@ -0,0 +1,201 @@
+ID	NUMBER	CONCEPTICON_ID	CONCEPTICON_GLOSS	ENGLISH	MAURITIAN	HAITIAN	FRENCH
+Wittmann-1973-200-1	1	98	ALL	all	tu	tu	tout
+Wittmann-1973-200-2	2	1577	AND	and	ek	é, épi	et
+Wittmann-1973-200-3	3	619	ANIMAL	animal	zanimo	bèt	animal
+Wittmann-1973-200-4	4	646	ASH	ashes	lasan	sân	cendres
+Wittmann-1973-200-5	5	1461	AT	at		a	à
```
AnnikaTjuka, line 6: Is the MAURITIAN form missing in the original data set as well?

concepticondata/conceptlists/Wittmann-1973-200.tsv
```diff
@@ -0,0 +1,201 @@
...
+Wittmann-1973-200-7	7	1292	BAD	bad	move	móvé	mauvais
+Wittmann-1973-200-8	8	1204	BARK	bark	lekors	ékòs	écorce
+Wittmann-1973-200-9	9	1157	BECAUSE	because	akoz	paské	parce que
+Wittmann-1973-200-10	10	1251	BELLY	belly	vât	vât	ventre
+Wittmann-1973-200-11	11	1202	BIG	big	gro, grâ, bel	gró, grâ	gros, grand
+Wittmann-1973-200-12	12	937	BIRD	bird	zozo	zwézó	oiseau
+Wittmann-1973-200-13	13	1403	BITE	to bite	mord-e	mòdé	mordre
+Wittmann-1973-200-14	14	163	BLACK	black	noar	nwa	noir
+Wittmann-1973-200-15	15	946	BLOOD	blood	disâ	sâ	sang
+Wittmann-1973-200-16	16	176	BLOW (WITH MOUTH)	to blow	sufle	suflé	souffler
```
AnnikaTjuka, line 17: _souffler_ seems to be ambiguous. It is mapped to  BLOW (WITH MOUTH) and BLOW (OF WIND). If we take BLOW (OF WIND) as the prototypical one, I suggest to remap (see also #836).

concepticondata/conceptlists/Wittmann-1973-200.tsv
```diff
@@ -0,0 +1,201 @@
...
+Wittmann-1973-200-74	74	1448	HOLD	to hold	tini	kêbé	tenir
+Wittmann-1973-200-75	75	1239	HOW	how	kumâ, ki manyer	kômâ	comment
+Wittmann-1973-200-76	76	1435	HUNT	to hunt	lasas, pus-e	šasé	chasser
+Wittmann-1973-200-77	77	1200	HUSBAND	husband	mari, bonom	mari	mari
+Wittmann-1973-200-78	78	1209	I	I	mo-a	mwê	moi
+Wittmann-1973-200-79	79	617	ICE	ice	laglas	glas	glace
+Wittmann-1973-200-80	80	1459	IF	if	si	si	si
+Wittmann-1973-200-81	81	1460	IN	in	dâ	nâ, lâ	dans
+Wittmann-1973-200-82	82	1417	KILL	to kill	tuy-e	tuyé	tuer
+Wittmann-1973-200-83	83	1410	KNOW (SOMETHING)	to know	kon-e	kôn(ê)	connaître
```
AnnikaTjuka, line 84: _connaître_ is mapped in two lists to KNOW.

concepticondata/conceptlists/Wittmann-1973-200.tsv
```diff
@@ -0,0 +1,201 @@
...
+Wittmann-1973-200-85	85	1355	LAUGH	to laugh	rir	ri	rire
+Wittmann-1973-200-86	86	628	LEAF	leaf	fey	fèy	feuille
+Wittmann-1973-200-87	87	244	LEFT	left (side)	gos	gòš	gauche
+Wittmann-1973-200-88	88	1297	LEG	leg	lazam	žâm	jambe
+Wittmann-1973-200-89	89	1411	LIE (REST)	to lie	alôz-e	kuše, layé	être couché
+Wittmann-1973-200-90	90	1422	BE ALIVE	to live	viv	viv	vivre
+Wittmann-1973-200-91	91	1224	LIVER	liver	lefoa	fwa	foie
+Wittmann-1973-200-92	92	1203	LONG	long	lon	lô(g)	long
+Wittmann-1973-200-93	93	1392	LOUSE	louse	lipu	pu	pou
+Wittmann-1973-200-94	94	2106	MALE PERSON	man-male	zom	nòm, nèg	homme
```
AnnikaTjuka, line 95: Not sure about this one. _homme_ is commonly mapped to MAN.

AnnikaTjuka (commented):
I left a few comments. I'm not a French expert, so my comments are based on already existing mappings.

LinguList:
good catches, @AnnikaTjuka, I agree with all your suggestions!

concepticondata/conceptlists/Wittmann-1973-200.tsv
CarolinHu: In the original article, the missing form is represented as "Ø". I left it blank here so it's not interpreted as the corresponding letter. Should I rather change it to Ø?

concepticondata/conceptlists/Wittmann-1973-200.tsv
CarolinHu: True, I forgot which of those was our prototypical case. Will change it.

concepticondata/conceptlists/Wittmann-1973-200.tsv
CarolinHu: I mapped it to KNOW (SOMETHING) because that's the corresponding mapping in Swadesh-1952-200.

concepticondata/conceptlists/Wittmann-1973-200.tsv
CarolinHu: (see comment above) In Swadesh, it's MALE PERSON as well.

LinguList:
@CarolinHu, your argumentation is also very reasonable. To make this "explicit", you could just add one sentence to the NOTE field in conceptlists.tsv, and say that the mappings follow by and large the traditional Swadesh list of 200 items, but that the French translations may diverge from those at times.

concepticondata/conceptlists/Wittmann-1973-200.tsv
AnnikaTjuka: Ø usually refers to a 'null-morpheme'. You could add "Ø (null-morpheme)" in the cell so that it doesn't look like a letter or missing data point. @LinguList what do you think?

LinguList: It is difficult to say what Ø refers to in lexicostatistical datasets, so I'd rather also leave it blank. What one can also do, since this is supposed to be a lexical dataset, is to delete the language and leave only those which serve as glossing languages? In fact, even the French as a lexifier is potentially something we may not need, since they obviously chose some "translations" in such a way that the words are cognate!

CarolinHu: Maybe someone else wants to chime in to decide this? I would leave all four languages in there, especially French, because while the list is based on the English Swadesh concepts, the description sounded very much like French was the point of departure. It also makes some non-optimal translations of Swadesh's original more visible if Haitian and Mauritian are represented.

LinguList (approved):

CarolinHu:
Thank you all for your comments! With the final approval, I'm going to merge this PR now.

Issue #1044, opened 2021-02-02
gerarditupi concept list now available


LinguList:
The concept list was just published in Diachronica: https://benjamins.com/catalog/dia.18032.fer

LinguList:
This relates to the original lexibannk dataset, which was already published, but without making use of a specific concepticon version: https://github.com/lexibank/gerarditupi

So concepts are mapped, we just need to add them to concepticon now.

Kristina-Pianykh:
I can add the list but I currently don't have the source. It would very much appreciate it if someone who has one could send it to me.

LinguList:
Just sent you the paper via email. Since it is still published before 
appearing in a real volume, page numbers will change later, as well as 
the number info (but the volume is 38 (for 2021).

PR #1045, opened 2021-02-04
Add Gerardi-2021-248

Kristina-Pianykh:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information
#1044. I have lots of questions concerning this list. First of all, quite a few mappings were obviously wrong (e.g., DANCE was linked to BE, SAND - to SEND, BAMBOO - to HARVEST CROPS), while some basic concepts were initially left unmapped (e.g., CHEST and FALL). I took the liberty of correcting those as well as some others (e.g., BURN_intr - to BURNING instead of BURN, YOU (sg) - to THOU instead of YOU). I checked with Gerardi-2021-414 and found out that these exact mismappings occur there too. I want to make sure that correcting them here was the right decision because in this case Gerardi-2021-414 should be fixed accordingly.

Another thing is that they claim the list contains 244 items while the provided data in fact have 248 concepts. The last thing is that there is no explicit indication of what the list is based on.


Kristina-Pianykh:
<details>
<summary>concepticon check</summary>

```
concepticondata\conceptlists\Gerardi-2021-248.tsv
Check: unique_concepticon_gloss
  CONCEPTICON_ID    LINE_NO  ID                      NUMBER  GLOSS
----------------  ---------  --------------------  --------  --------
            1257        244  Gerardi-2021-248-243       243  WING
            1257        245  Gerardi-2021-248-244       244  WING (2)
Check: unique_id
OK
Check: unique_number
OK
Check: matching_concepticon_gloss_and_id
OK
Check: valid_concepticon_gloss
OK
```
</details>

LinguList:
How good you checked that!

In the original mapping, they did not use the  Concepticon glosses, but rather only the IDs, I think that's when all these errors slipped in (dance: 1879, be: 1579, or send vs. sand was then a spelling problem).

LinguList:
Yes, the numbers are again a problem, I think I even checked the number in the appendix, when I saw it in the latex version, but it was not really clear, etc. For the number of conceepts, it is possible that there were four more, but 244 were used for the alignments, something we have to check in the lexibank dataset. 

And yes: in this case, the larger list must be corrected too. I suggest, that we then do both at once? It'll be easier to review, I think.

Kristina-Pianykh:
I checked with the Appendix and there are exactly 244 items. The data I took from Lexibank contains 4 more concepts. 3 of them are marked with the note "Only in Cognate List."

LinguList:
That's it: they are in one list, but not in the other. What I'll do now is: if the lexical items etc. are mixed, I'll reduce the lexical data in the lexibank repository to the items mentioned in the paper, so we have some continuity there. Every other variant would seem to tedious for us to maintain.

Kristina-Pianykh:
I reduced the list to the 244 items presented in the publication and corrected some mappings in Gerardi-2021-414.

concepticondata/conceptlists.tsv
```diff
@@ -354,7 +354,8 @@ LSDO-2019-462	Language and Social Development Organization	2019		462	basic, area
...
 Wittmann-1973-200	Wittmann, Henri Gontran	1973		200	basic, areal	French	Mauritian, Haitian		Wittmann1973		This list, based on [Swadesh’s 200 item list](:ref:Swadesh-1952-200), was used for lexicostatistical analysis of the relatedness between Caribbean Creoles (represented by Haitian in the list) and Mascarene Creoles (represented by Mauritian). The chosen words of both Creole languages were compared among each other and with French with regards to their homosemantic similarity, yielding a greater relatedness of each Creole to French than to each other. Our mappings generally follow the English concepts in Swadesh; however, the French translations may diverge from them in some cases.	94-98	
+Gerardi-2021-244	Ferraz Gerardi, Fabrício and Reichert, Stanislav	2021		244	basic, areal	English	Tupi-Guarani languages	https://doi.org/10.1075/dia.18032.fer	Gerardi2021b		This 244-item list was used in the analysis of sound correspondences in the 38 Tupian languages and in the development of a new internal classification of the Tupi-Guarani branch.	30-35
```
LinguList, line 361: There was not analysis of sound correspondences in this study. So I'd say, rather:
```
This 244 item list was the basis of a pilot study on the classification of 38 Tupían languages. It was later expanded to [414 concepts](:ref:Gerardi-2021-414) as part of the TULED database. 
```

concepticondata/conceptlists/Gerardi-2021-244.tsv
```diff
@@ -0,0 +1,245 @@
...
+Gerardi-2021-244-4	4	ASH	646	ASH
+Gerardi-2021-244-5	5	AXE	677	AXE
+Gerardi-2021-244-6	6	BACK	1291	BACK
+Gerardi-2021-244-7	7	BAD/EVIL	3764	BAD OR EVIL
+Gerardi-2021-244-8	8	BAMBOO	1927	BAMBOO
+Gerardi-2021-244-9	9	BANANA	868	BANANA
+Gerardi-2021-244-10	10	BASKET	1539	BASKET
+Gerardi-2021-244-11	11	BAT	1793	BAT
+Gerardi-2021-244-12	12	BE_WRONG	1390	WRONG
+Gerardi-2021-244-13	13	BEANS	2574	BEAN OR PEA
```
LinguList, line 14: 823 BEAN

LinguList: Sorry: or do they have BEAN OR PEA in the other database?

LinguList: No, I see now: it is in fact BEAN, I think BEAN or PEA makes less sense (also for the other list).

concepticondata/conceptlists/Gerardi-2021-244.tsv
```diff
@@ -0,0 +1,245 @@
...
+Gerardi-2021-244-176	176	SICK/ILL	1847	SICK
+Gerardi-2021-244-177	177	SISTER	1263	SISTER
+Gerardi-2021-244-178	178	SIT	1416	SIT
+Gerardi-2021-244-179	179	SLEEP	1585	SLEEP
+Gerardi-2021-244-180	180	SMALL	1246	SMALL
+Gerardi-2021-244-181	181	SMOKE	778	SMOKE (EXHAUST)
+Gerardi-2021-244-182	182	SNAKE	730	SNAKE
+Gerardi-2021-244-183	183	SNORE	1983	SNORE
+Gerardi-2021-244-184	184	SON	1620	SON
+Gerardi-2021-244-185	185	SOUR/ACID
```
LinguList, line 186: 1906 sour (even if it is something that can also mean ACID, I think sour is enough).

LinguList (approved):
Sorry, should've said that one earlier: I tried to make OR concept in such a way that we follow strict alphabetical order, so "bark or shell", "hurt or pain", etc. I think it is good to keep this in this way, even if it is not really "tested" by the code.

I'll go through the mappings now and double check them.

Okay, done, see my comments (BEANS should only link to BEAN, SOUR/ACID can link to SOUR, and that's it, I think). 

Sorry for not warning earlier about the conventions for the glosses... 

I already approve, so you can just merge once these things have been corrected. Many thanks!

Sorry, just saw: sour / acid / bitter: I think one can really link this to sour for now, but if they really merge sour and bitter, as two "fundamental" tastes, one may consider adding one more super-concept here. I leave that to you.

Kristina-Pianykh:
Thank you for the tips and the thorough review!

PR #1069, opened 2021-03-09
fix vulic

LinguList:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

fix concepticon concept list, add more languages to concepticon.json

@AnnikaTjuka and @chrzyki could you please check this? I added more languages to concepticon.json, adjusted the reference (the paper finally appeared, now post-dated to 2020), and also made sure that column names are consistent (Chinese instead of Mandarin, as this is what we use in concepticon.json). @AnnikaTjuka, please also check the concepticon.json, as this contains an internal re-mapping (adding glottocode and two-letter iso codes) of column names for languages. And you need to add the languaes we want to "re-use" later to the "source language" column. For this reason, you can just name a column ENGLISH, and we "know" the glottocode. The target language column has no real semantics, it is a free field, but SOURCE LANGUAGE in conceptlists.tsv indicates what we interpret as glosses and use for the internationalization of concepticon gloss mapping


AnnikaTjuka (approved):
Thanks for updating the list! I saw that you also changed the date back to 2020. I guess the confusion was that the PDF on the Computational Linguistics journal website was still the preprint version. They must have changed it in the last week.

chrzyki (approved):
Looks all good from me, thanks! Thank you also for updating `concepticon.json`, that's something we definitely need to keep in mind more often.

PR #1079, opened 2021-04-24
Adding Constenla-2005

Tarotis:
# Pull request checklist

- [x]  add new concept list
- [x] add new metadata
- [ ] retire data

## Additional information
This PR aims at adding the 110-item concept list offered by Constenla Umaña (2005). No new concepts have been introduced. As this is my first PR to Concepticon I hope I got everything right. If not I will try to adapt any necessary change asap.
...


Tarotis:
@LinguList I created the PR according to the guidelines in the blog, but for some reason the fest fails, claiming inconsistent data. I tried to pin it down by checking every step of the mapping process, but I can't seem to find it. What does this error message usually refer to?

LinguList:
fixed it, the metadata should give type "string" for the field "NUMBER", as the number can also contain "ABC", etc., and the check is sensitive to this :)

LinguList:
I'll moderate this PR and would ask @Schweikhard and @chrzyki to have a quick look during next week, if possible.

concepticondata/conceptlists/Constenla-2005-110.tsv
```diff
@@ -0,0 +1,111 @@
...
+Constenla-2005-110-90	90	SUN	1343	SUN	SOL
+Constenla-2005-110-81	81	TAIL	1220	TAIL	COLA
+Constenla-2005-110-12	12	TALK	2708	TALK	HABLAR
+Constenla-2005-110-98	98	THIS	1214	THIS	ESTE
+Constenla-2005-110-23	23	THOU	1215	THOU	TÚ
+Constenla-2005-110-30	30	THREE	492	THREE	TRES
+Constenla-2005-110-108	108	TONGUE	1205	TONGUE	LENGUA
+Constenla-2005-110-25	25	TOOTH	1380	TOOTH	DIENTE
+Constenla-2005-110-107	107	TREE	906	TREE	ÁRBOL
+Constenla-2005-110-27	27	TUMBLE (FALL DOWN)	2894	TUMBLE (FALL DOWN)	CAER
```
Schweikhard, line 97: Considering the English gloss, this is probably mapped correctly, but all other instances of caer in the database seem to be mapped to FALL instead.

concepticondata/conceptlists/Constenla-2005-110.tsv
```diff
@@ -0,0 +1,111 @@
...
+Constenla-2005-110-40	40	HAND	1277	HAND	MANO
+Constenla-2005-110-61	61	HE	1211	HE	ÉL
+Constenla-2005-110-104	104	HEAD	1256	HEAD	CABEZA
+Constenla-2005-110-75	75	HEAR	1408	HEAR	OÍR
+Constenla-2005-110-67	67	HEAVY	1210	HEAVY	PESADO
+Constenla-2005-110-63	63	HIT	1433	HIT	GOLPEAR
+Constenla-2005-110-79	79	HOT OR WARM	2272	HOT OR WARM	CALIENTE
+Constenla-2005-110-11	11	I	1209	I	YO
+Constenla-2005-110-68	68	KILL	1417	KILL	MATAR
+Constenla-2005-110-10	10	KNOW	3626	KNOW	SABER
```
Schweikhard, line 56: I think saber is more specific than know and only refers to KNOW (SOMETHING).

concepticondata/conceptlists/Constenla-2005-110.tsv
```diff
@@ -0,0 +1,111 @@
...
+Constenla-2005-110-2	2	BARK	1204	BARK	CORTEZA
+Constenla-2005-110-66	66	BIG	1202	BIG	GRANDE
+Constenla-2005-110-37	37	BIRD	937	BIRD	AVE
+Constenla-2005-110-54	54	BITE	1403	BITE	MORDER
+Constenla-2005-110-62	62	BLACK	163	BLACK	NEGRO
+Constenla-2005-110-86	86	BLOOD	946	BLOOD	SANGRE
+Constenla-2005-110-57	57	BONE	1394	BONE	HUESO
+Constenla-2005-110-6	6	BREAST	1402	BREAST	TETAS
+Constenla-2005-110-38	38	CHILD	2099	CHILD	NIÑO
+Constenla-2005-110-93	93	CLEAN	704	CLEAN	LAVAR
```
Schweikhard, line 14: CLEAN refers to the property. Depending on the exact meaning, this should be mapped to CLEAN (SOMETHING), WASH, or LAUNDER (WASH CLOTHES). We have thus far mapped lavar to WASH.

concepticondata/references/references.bib
```diff
@@ -3860,3 +3860,12 @@ @Book{Say2020b
   Address                             = {St. Petersburg},
   Publisher                           = {Institute for Linguistic Studies, RAS}
 }
+
+@Article{Constenla2005,
+  author       = {Constenla Umaña, Adolfo},
+  journaltitle = {Estudios de Lingüística Chibcha},
+  title        = {?`Existe relación genealógica entre las lenguas misumalpas y las chibchenses?},
```
Schweikhard, line 3867: I would suspect this should start with ¿ rather than ?`.

concepticondata/conceptlists.tsv
```diff
@@ -374,3 +374,4 @@ Thwaites-2020-98	Thwaites, Peter	2020		98	specific	English	English		Thwaites2020
...
 Say-2020-130	Say, Sergey	2020		130	questionnaire	English	Global		Say2020a		The given list underlies the BivalTyp database of 130 bivalent predicates in 86 languages, mainly spoken in Northern Eurasia. The data were gathered in the questionnaire, carried out in [Say et al. (2020)](:bib:Say2020b).		
+Constenla-2005-110	Constenla Umaña, Adolfo	2005		110	basic	Spanish	Chibchan, Misumalpam and Lencan languages		Constenla2005		The list is a reduced form from a 120-item concept list which the author used in a previous paper on the relationship on Lencan and Misumalpam languages. In this version, the information for Chibchan languages was added to analyze cognates of all three families and argue in favor of an underlying genealogical relationship between them.
```
Schweikhard, line 377: "a reduced form _of_" sounds more correct to me.

Schweikhard (commented):
Thank you for contributing this list! The mapping and everything looks good, just some small corrections:

chrzyki (approved):
Likewise, many thanks for adding this. I don't have anything to add except for what @Schweikhard already mentioned. Is there any reason for the order of the glosses/concepts in the list? Why not order it by ascending NUMBER/ID?

concepticondata/references/references.bib
Tarotis: As far as I know, both usually work, but I guess its more transparent if `¿` is used.

Tarotis:
Just to be sure on the workflow because this is the very first PR I am working on in git: I adapt and commit the suggested changes, and afterwards Mattis merges the PR?

As for the ordering, there is no special reason for this. Would it make more sense to order according to Concepticon_ID or is this not uniformly formatted across the other lists and Number (the original authors order in his paper) works just as well?

chrzyki:
> 
> 
> Thanks for the kind replies and suggestions. Just to be sure on the workflow because this is the very first PR I am working on in git: I adapt and commit the suggested changes, and afterwards Mattis merges the PR?
> 
> As for the ordering, there is no special reason for this. Would it make more sense to order according to Concepticon_ID or is this not uniformly formatted across the other lists and Number (the original authors order in his paper) works just as well?

Yes, you can incorporate the changes and then push the updated branch. Afterwards, they'll appear here in this PR and we can approve/merge.

For the ordering: Yes, generally speaking, we try to keep the order as in the original list and the resulting TSV should be in ascending order for the NUMBER/ID column. Makes for an easier reading.

Tarotis:
Thanks for the fast reply! I now updated the files with the corresponding changes.
In order to comply with the original author's ordering (Spanish alphabet), a rewrite of the number-column was necessary. Actually, you are not only totally correct in that it improves readability, but it also improves comparability if somebody wants to cross-check the lexibank-database with the original paper, so thanks a lot for that suggestion.

LinguList:
I'll merge later, once I have reviewed the whole discussion and see no need for further changes. BTW: is the original text with glosses in English and Spanish or only Spanish?

Tarotis:
The original glosses are only given in Spanish

LinguList:
So the column ENGLISH should be renamed to column GLOSS, as you made the 
translation, right? In that case, I'd even suggest to go lowercase, as 
you can give a translation that also reflects part of speech, etc., and 
is a bit more verbose. E.g., you can indicate if something is a verb, 
right? Otherwise, bailar is ambiguos as gloss in "dance".

LinguList:
Ah, just to be sure, could you send me the PDF via email, so I can have 
a look?

Tarotis:
I just sent you a download link per E-Mail (file was too big). Should I adapt the changes to ENGLISH/GLOSS already or wait for your confirmation? The translation is mine, so all of the suggestions can be adapted easily.

LinguList:
Thanks, I received the PDF now, this is important, as you will have to modify the Spanish glosses as well. You give them in upper case fully, but the source only capitalizes the first letter, so you should adhere to this. If it is not too much work, you could even add the page number in an additional column, which is helpful for those who want to look at the original resource. The English glosses are free, but I recommend to disambiguate things like "dance", so it is clear which POS we are dealing with.

Tarotis:
Thanks for the suggestions, I changed the respective columns and included a new column `PAGE_IN_SOURCE` to the list and the metadata.

LinguList (approved):

LinguList:
I just modified the glosses, will then merge.

PR #1107, opened 2021-05-27
Add Feleke 2021-150

chrzyki:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

As per https://github.com/lexibank/felekesemitic.


concepticondata/conceptlists.tsv
```diff
@@ -382,4 +382,5 @@ Hantgan-2018-302	Hantgan, Abbie and List, Johann-Mattis	2018		302	basic	English
...
+Feleke-2021-150	Feleke, Tekabe Legesse	2021		150	specific, areal	English	Ethiosemitic languages	https://www.sciencedirect.com/science/article/pii/S2215039021000023	Feleke2021		The list was used in a study for the areal classification of Ethiosemitic languages and to establish distances/relationships between the members of this group. The items for the list were randomly selected. Note that the article mentions that the list contains 147 items, whereas the actual supplementary materials consist of 150 items.
```
LinguList, line 386: the "randomly" selected sounds a bit strange, as this list seems ot be edited, so I'd add "according to the author"

LinguList (approved):
Thanks, @chrzyki! I have not checked for THOU, but my mapping was due to the fact that there's no "you plural", so I think this is fine here. Otherwise, one would have to check with Amharic or some other language in the original. Feel free to merge later (maybe after addressing my small comment).

concepticondata/conceptlists.tsv
chrzyki: Done.

Issue #955, opened 2020-10-08
Improvements to Galucio-2015-90


tresoldi:
We can improve the mapping of [Galucio-2015-90](https://concepticon.clld.org/contributions/Galucio-2015-90), based on work by @jtbirchall. 

- Concepts not linked but existing in Concepticon:

  - GLOWWORM (vagalume) -> FIREFLY 206

- Concepts wrongly mapped:

  - PEACH PALM (PUPUNHA) is mapped to PEACH TREE (but it is PEACH PALM, not in Concepticon)
  - LARGE ANT (TOCANDEIRA) is mapped to ANT, but it is not a general term ("tocandeira" or "tucandeira" are a specific ant species, culturally important in some groups)
  - TIMBÓ LIANA is mapped to LIANA, but it is a different group of plants
  - CARÁ ROOT is mapped to ROOT, but it is not a general term (the closest match would be MANIOC 927)
  - PACU FISH is mapped to FISH, but it is not a general term (it a family)

- A number of concepts do not exist as concept sets, but at least some of them should be included as they are common in wordlists from the Amazonian/South American area and/or common concepts for study language and culture relationships (including borrowings). In order of importance for inclusion (order given by me), they are:

- [ ] BRAZIL NUT, perhaps one of the most important products ([in TuLeD](https://tuled.org/parameters/49))
- [X]   - PEACH PALM, as in the item above ([in TuLeD](https://tuled.org/parameters/249))
- [ ]   - ANATTO, the orange-red condiment and food coloring derived from the achiote tree, an extremely important cultural concept  ([in TuLeD](https://tuled.org/parameters/7))
- [ ]   - TIMBÓ, as in the item above, considering its importance for study of linguistic and cultural relations ([in TuLeD](https://tuled.org/parameters/353))
- [ ]   - GNAT (PIUM), could be added as GNAT, found also in other geographic areas ([in TuLeD](https://tuled.org/parameters/148))
- [ ]   - TUCUMÃ PALM, the *Astrocaryum vulgare*, one of the most important palms of the Amazon ([in TuLeD](https://tuled.org/parameters/364))
- [ ]   - TRAHIRA FISH, fish of the *Erythrinidae* family, found from Costa Rica to Argentina  ([in TuLeD](https://tuled.org/parameters/362))
- [x]   - GENIPA, the *Genipa americana*, an important plant/fruit of the region ([in TuLeD](https://tuled.org/parameters/145))
[...]

PR #1115, opened 2021-05-28
Galucio improvements

Kristina-Pianykh:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information
Refers to #955. The first part of the problematic mappings was resolved in #989. In this PR, I've added new concept sets only in those cases when there is another list in concepticon with the same concept, which I think is reasonable, also regarding the previous practice of introducing new concepts given there are more than one lists with the concept.


LinguList (approved):

LinguList:
Very nice, many thanks!

PR #1123, opened 2021-06-03
adding tng.org concept list

SimonGreenhill:
# Pull request checklist

- [X] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

As requested by @LinguList. Note that the concept list has ~1200 items but 2525 ids as there are some missing items. I've named this -2525 to match the ids rather than the count, but perhaps the other way around is better. Let me know.


LinguList (approved):
Thanks, could you add one sentence regarding the concept mapping, like "Since the data was collected from a wider range of sources, the concept mapping was done in a very conservative manner and may be improved in the future." (along these lines). This would then make it clear, I hope, that with more concepts to be added, we may find more matches, etc. But as far as I checked, the current mapping is okay, I did not find completely mismapped items so far.

But this little caveat will allow us to circumvent the otherwise complex review protocol for now.

SimonGreenhill:
done!

LinguList (approved):

PR #1124, opened 2021-06-03
Add Tjuka-2019 body parts

AnnikaTjuka:
# Pull request checklist

- [X] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR adds a small list of body part terms across 15 languages. The data is also available as a CLDF dataset [here](https://github.com/cldf-datasets/tjukabodyparts).


AnnikaTjuka:
@chrzyki Would you mind reviewing this PR?

chrzyki:
Yep, I can have a look!

chrzyki (approved):
Looks all good, thanks! One small thing: I think we tend to mostly use empty cells to indicate `NA`s instead of directly putting `NA` (or similar) into the cell.

AnnikaTjuka:
I forgot to mention this: I received an error message "missing translation" when the cells were empty. That's why I used "NA".

chrzyki:
> 
> 
> I forgot to mention this: I received an error message "missing translation" when the cells were empty. That's why I used "NA".

Ahh, right! That was it - yeah, all good then! :)

AnnikaTjuka:
Thanks for the review! I'll merge now.

PR #1140, opened 2021-10-20
added the PDF and listed it in conceptlists.tsv

ilchec:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

...


chrzyki:
Thanks @ilchec. The contributing guidelines are outdated in on part where the handling of source PDFs is described. Basically, you just add the PDF name to `conceptlists`.tsv and send me the PDF directly (we want to avoid having large PDFs in the source tree). I'll then take care of the rest and upload it.

Issue #610, opened 2019-06-17
Nimuendaju-1939


tresoldi:
Curt Nimuendajú published a number of wordlist for Brazilian Native languages, with German and Portuguese glosses. One of the most stable ones seems to be the list used for Masakari in 1939, later reused for other languages:

http://etnolinguistica.wdfiles.com/local--files/artigo%3Aaraujo-1996/araujo_1996_masakari.pdf

It is worth adding it to Concepticon, as there is an going project to scan the lists of Nimuendajú after the fire at the Brazilian National Museum last year destroyed a number of them.

PR #1153, opened 2021-12-05
Added Nimuendaju1958 (from Araujo 1996, issue #610).

marthuis:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

...


chrzyki:
Hi @marthuis, thanks for preparing this list! Before we start the review process I'd kindly ask you to take care of the auto formattings that seem to have been applied to the different files you edited:

- capitalization and auto formatting in conceptlists.tsv
- large bulk of indentation (?) changes in references.bib

Thanks!

LinguList (commented):
Thanks a lot. Can I ask you to correct one crucial point already? Obviously, you use a spreadsheet editor, as you corrected the DATES in the conceptlists.tsv, which again were introduced by somebody unnoticed, as they do not make any sense, since cases like 07. Nov in line 173 should be page ranges. And they were apparently overwritten in the past! E.g. 173 Nagaraja-2013-200 is on page 7-11, so you see, what I think what happened was that the spreadsheet interprets page ranges as dates and converts them as such. So we really need to revert those page ranges back. 

Or @chrzyki, what do you think: we go back in the history until we restore the version without this automatic modification and then make sure to correct this in a separate issue? 

In any case, the upper-casing in the spreadsheet would have to be corrected or reverted (see the DIFF for conceptlists.tsv).

chrzyki:
> Thanks a lot. Can I ask you to correct one crucial point already? Obviously, you use a spreadsheet editor, as you corrected the DATES in the conceptlists.tsv, which again were introduced by somebody unnoticed, as they do not make any sense, since cases like 07. Nov in line 173 should be page ranges. And they were apparently overwritten in the past! E.g. 173 Nagaraja-2013-200 is on page 7-11, so you see, what I think what happened was that the spreadsheet interprets page ranges as dates and converts them as such. So we really need to revert those page ranges back.
> 
> Or @chrzyki, what do you think: we go back in the history until we restore the version without this automatic modification and then make sure to correct this in a separate issue?
> 
> In any case, the upper-casing in the spreadsheet would have to be corrected or reverted (see the DIFF for conceptlists.tsv).

Yes, I agree, we should handle that in a separate issue/PR. Also a good practice for @marthuis to see how merge conflicts can be resolved.

concepticondata/conceptlists.tsv
```diff
@@ -293,13 +293,13 @@ Schroeder-2012-824	Schroeder, Astrid and Gemballa, Teresa and Ruppin, Steffie an
...
+Maciejewski-2016-100	Maciejewski, Greg and Klepousniotou, Ekaterini	2016		100	ratings	English	English	" http://osf.io/7k3eh/"	Maciejewski2016		The list comprises relative meaning frequencies for 100 homonyms (British eDom Norms) and other lexical norms. For feasibility, each line of the table was duplicated to map the different meanings to one of the concepts found in Concepticon. The meaning definition was taken from the Wordsmyth dictionary [(Parks et al. 1998)](:bib:Parks1998). The dominance scores were included from [Armstrong et al. (2012)](:bib:Armstrong2012). Word frequencies and number of orthographic neighbours  were based on the British National Corpus [(BNC 2007)](:bib:BNC2007). The numbers for word senses were taken from [Parks et al. (1998)](:bib:Parks1998) and WordNet [(Fellbaum 1998)](:bib:Fellbaum1998). Semantic diversity values were included from [(Hoffman et al. 2013)](:bib:Hoffman2013). Ratings on familiarity, concreteness, and imageability were taken from [Stadthagen-Gonzalez and Davis (2006)](:bib:Stadthagen2006) and [Coltheart (1981)](:bib:Coltheart1981).
```
LinguList, line 302: There are some cases of `"` which should not be there in the file, you can see this from the DIFF here. Originally : 
```
http://osf.io/7k3eh/
```

now

```
" http://osf.io/7k3eh/"
```
This should be reverted.

LinguList (commented):
Okay, now we have problems with the quotes `"` as I mentioned in one comment. They should not be used here.

LinguList (changes requested):
So @marthuis if we proceed with this list now. What you can do is:

Actively download the most recent conceptlists.tsv from the github (open conceptlists.tsv in raw, then save). Introduce only that one line for hte concept list you add here, in the end, and do so in a text file, to not auto-format (until you have control over preventing auto-formating). 

Put it into your repository instead of the previous conceptlists.tsv

Then, you do a git pull to get the most recent version, and if there are no warnings, push again to this branch here.

We'll then make sure to review the mappings.

LinguList:
Let me know if running tests works on your side, as this means we can further proceed with sharing some of the basic linking data here.

marthuis:
@LinguList This is clear, the tab error should be fixed now, thanks!

LinguList:
Well, I am afraid you have re-introduced the autoformatting now, which I fixed some time ago.

marthuis:
@LinguList I'm on it. Also to be sure, as you can see the Nimuendaju-1958-298 file from this PR does not contain a GLOSS column, but was validated. Should I add it before committing to the new PR?

LinguList:
You need either an ENGLISH column or a GLOSS column. I will check with 
the new PR or your update.

PR #1162, opened 2022-02-28
Add Tjuka-2022-220

AnnikaTjuka:
# Pull request checklist

- [X] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR adds an extended version of Tjuka-2021-192.


concepticondata/references/references.bib
```diff
@@ -1,3 +1,11 @@
+@Misc{Tjuka2022,
```
LinguList, line 1: We quote our blog posts now as a journal. So you should write @article instead.

Like here:

```
@Article{Tjuka2020TBLOG1,
  author    = {Tjuka, Annika},
  title     = {Adding concept lists to Concepticon: A guide for beginners},
  journal   = {Computer-Assisted Language Comparison in Practice},
  year      = {2020},
  number    = {1},
  url       = {https://calc.hypotheses.org/2225},
  volume    = {3},
}
```

This year is volume 5, and page range is the range of the pages in your PDF.

LinguList: So I'd recommend to not quote Zenodo, which is referenced in the blogpost, but rather the article/blog post referencing this list. The URL / DOI can still be added from Zenodo.

LinguList (approved):
I already approve, but recommend to addjust the reference, as I suggested.

AnnikaTjuka:
I updated the reference for the current list and also for the previous version.

LinguList (approved):

PR #1176, opened 2022-05-07
update with new schema

LinguList:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

Check for csvw
...


LinguList:
@AnnikaTjuka, I foudn one solution, which works. We add "null" for undecided cases. We should add this to our "cheat sheet" for csvw, when we write up a blogpost on this topic.

LinguList:
Another trick for the cheat sheet:

```python
from pyconceepticon import Concepticon
con = Concepticon()
con.conceptlists["Epps-2021-843"].concepts
```

If this WORKS, the csvw does not have errors. So you can check without running all of `concepticon test`, but make sure, if you use this in python, to always initialize concepticon again, so you type:

```python
con = Concepticon()
con.conceptlists["Epps-2021-843"].concepts
```

Only the last line would use the old configuartion of csvw-metadata.

AnnikaTjuka:
I see! Yes, now it works.

AnnikaTjuka (approved):

PR #1203, opened 2022-08-03
Changed gloss KID to YOUNG GOAT (KID)

mathildavz:
I changed all glosses KID to YOUNG GOAT (KID)
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

...


LinguList (changes requested):
This is interesting, since it reveals individual problems in the concept lists. E.g. the list by Hill-2015-999 should be linked to "young human", as it is unlikely that this list refers to a young goat. Pepper's list is also probably rather referring to a young human, I think, we can ask Steve Pepper also in case of doubt. Snider needs double-checking, as needs Vergallito, Epps, and all Carling lists. Huang is more or less confirmed, but Anonby is not clear, and for Buck, I'd also suggest to compare with the original English gloss. 

I think, even if it takes time, it is best to go thorugh all 20 lists modified here and justify why they are linked to young goat or to young human now.

AnnikaTjuka:
I see. @mathildavz Would you be up to the challenge of reconstructing whether a gloss should be mapped to YOUNG GOAT or 1304 CHILD (YOUNG HUMAN)? The task includes checking the original source of the list and maybe checking if the list already has a mapping to CHILD (YOUNG HUMAN). You can also compare glosses that are linked to CHILD (YOUNG HUMAN) and see if you find a match. It's kind of detective work, but it would be a good way to get to know the lists. If you find an argument for mapping, you can post it as a comment in the respective line, see:
Screenshot 2022-08-03 at 11 50 54

mathildavz:
Of course, I'm up for that! It sounds like interesting work and will definitely help me get more familiar with the lists and the works of Concepticon as a whole. I'll start with this tomorrow.

mathildavz:
Here are my findings @LinguList and @AnnikaTjuka: 

- Anonby: YOUNG GOAT is meant, the previous glosses are male & female goat 
- Borin: YOUNG GOAT, see http://compling.hss.ntu.edu.sg/omw/cgi-bin/wn-gridx.cgi?synset=02416820-n
- Buck: YOUNG GOAT, similar to Anonby 
- all Carling lists: YOUNG GOAT, occurs in a list of farm animals 
- Epps: unsure, the Norwegian translation in the list (killing) actually translates to kitten (at least according to my quick search), so neither CHILD nor YOUNG GOAT would be correct 
- Hale: YOUNG GOAT, occurs in a list of farm animals 
- Haspelmath: YOUNG GOAT, occurs in a list of farm animals 
- Hill: I would argue that both instances refer to CHILD, but I wasn't able to find comprehensive evidence for this 
- Huang: YOUNG GOAT, occurs in a list of farm animals 
- Key: YOUNG GOAT, occurs in a list of farm animals 
- List 2014-2020: YOUNG GOAT, occurs in a list of farm animals 
- Lynott: I can’t find the supplemented material with the actual word list, I’ve tried to find the closed Issue but was only able to find the PR which gives no further information 
- Pepper: unclear, Gloss after KID is LAMB so that might give us a hint that the goat is meant, but it's not really good evidence 
- Snider: YOUNG GOAT, occurs in a list of farm animals 
- Vergallito: I would argue for CHILD as it occurs between boy and girl

mathildavz:
Based on this, I would propose changing Hill and Vergallito to 1304 CHILD (YOUNG HUMAN) and would wait for your opinion on Epps and Lynott.

LinguList:
I'd unlink it in Epps. Lynott is a list of 40000 norms, they are 
word-based, and they do not distinguish between the two readings of 
"kid". But the most frequent, it seems, nowadays is "kid = young human" 
(one might want to check with some dictionary like collins, what meaning 
they give first).

As to Pepper: we can write an email to Steve and ask.

AnnikaTjuka:
Thanks for providing this overview, @mathildavz ! Here are my comments on the shaky candidates:

- [ ] Epps: I found this Wikipedia page: https://nn.wikipedia.org/wiki/Killing so I think it should be mapped to YOUNG GOAT.
- [ ] Hill: That's a tricky case. The rating is based on word pairs. In this case, the pairs were boy-kid and daughter-kid. The ratings vary somewhat and the similarity seems to be a bit higher for the first pair. I would thus argue that both cases should be mapped to YOUNG HUMAN.
- [ ] Lynott: You can always find the original source by looking for the article either in `conceptlists.tsv` or online ([Lynott-2013-400](https://concepticon.clld.org/contributions/Lynott-2013-400)). Based on the ratings which are low for gustatory and olfactory, I would map to YOUNG HUMAN.
- [ ] Rzymski: Is correct since it occurs in the cluster of LAMB.

AnnikaTjuka:
> As to Pepper: we can write an email to Steve and ask.

I met Steve at a previous ALT conference and can write him a message.

AnnikaTjuka:
Just heard back from Steve Pepper: The gloss "kid" should be mapped to YOUNG GOAT.

mathildavz:
Thank you for the feedback! I will start integrating the changes now for Hill, Lynott an Vergallito.

AnnikaTjuka (approved):
Great detective work, @mathildavz! If @LinguList also approves the changes, you can merge.

LinguList (approved):
Very nice, many thanks, sorry for answering so late.

PR #1205, opened 2022-08-05
added more specific datatypes for variables relevant for norare

xrotwang:


LinguList (approved):
Very nice. Detailed data types should definitely be part of the paper, as this is one new huge step towards more consistent data in Concepticon and NoRaRe.

AnnikaTjuka (approved):
Many thanks for adding all the data types, @xrotwang!

PR #1207, opened 2022-08-09
Updated Bowern-2012-644

mathildavz:
I linked previously unlinked concepts in Bowern-2012-644. 

...


Tarotis:
Thanks for adding this PR, I will happily review.

I agree with `BILL`, `GRUB` and `LONG AGO`. The english forms have already been mapped to those concepts, so that shouldn't be problematic.

`later` should probably not be mapped to `late`, as the latter only represents adjectives and not the adverbial meaning comprised by `later`.

I am unsure about `DUSK`. Theoretically, there should be a difference between dusk and twilight as there is also morning twilight, but we did map twilight to `DUSK` before. How should this be handled in the best way?

AnnikaTjuka:
Thanks for the review, @Tarotis! I agree to unmap LATE and I would also unmap DUSK. There are two instances with a single gloss "twilight" and they are also not mapped. Most instances have either _dusk_ or _dawn_ as additional glosses to disambiguate.

mathildavz:
Thank you for the review @Tarotis! I've unmapped LATE and DUSK, as you've suggested.

PR #1210, opened 2022-08-11
Updated Bowern-2021-203c

mathildavz:
I've mapped Bowern-2021-203c-44 to COCA, as suggested in the google doc. However, I'm not sure about this as we already have a mapping of COCA in the list, where it is specified that a plant is meant. For Bowern-2021-203c-44 the list specifies it as "narcotics" rather than "flora-fauna", so I suppose they mean a substance won from the plant, instead of the plant itself. Should I unmap this again?


AnnikaTjuka:
Thanks, @mathildavz! Yes, I would unmap this one.

LinguList (approved):

PR #1212, opened 2022-08-12
Updated Bowern-2021-207a

mathildavz:
I've mapped two concepts. However, both to the same Concepticon Gloss. I do think both of this instances can be mapped to 1422 BE ALIVE, but I'm interested in your opinion on this.


LinguList (approved):

concepticondata/conceptlists/Bowern-2021-207a.tsv
```diff
@@ -8,7 +8,7 @@ Bowern-2021-207a-6	6	3pl	561	grammar	Basic Vocabulary	817	THEY
 Bowern-2021-207a-7	7	3sg	562	grammar	Basic Vocabulary	262	HE OR SHE OR IT
 Bowern-2021-207a-8	8	above	352	location	Basic Vocabulary	1741	ABOVE
 Bowern-2021-207a-9	9	again	353	grammar	Basic Vocabulary	204	AGAIN
-Bowern-2021-207a-10	10	alive	597	none	Basic Vocabulary		
+Bowern-2021-207a-10	10	alive	597	none	Basic Vocabulary	1422	BE ALIVE
```
AnnikaTjuka, line 11: Judging from the glosses mapped in other Bowern lists, I would unmap this one. It seems to be a more general meaning.

PR #1213, opened 2022-08-12
Updated Bowern-2021-342b

mathildavz:
I've linked some unlinked concepts in Bowern-2021-342b.


mathildavz:
I have two comments to make already: While I have mapped black palm to 3030 LONTAR PALM, according to the google sheet, I think it should be unmapped, as they are different species of palms to my knowledge.  
Same goes for elder being mapped to OLD (AGED). I believe the elder as a tree is meant, not the aged person.

LinguList:
Good arguments, I follow you on both!

LinguList:
So I'd say: unlink the lontar palm and the elder, and we can merge this.

AnnikaTjuka (approved):
Looks good! Feel free to merge.

Issue #1218, opened 2022-08-16
Remappings Bowern


AnnikaTjuka:
Based on the comments in PR https://github.com/concepticon/concepticon-data/pull/1212 and PR https://github.com/concepticon/concepticon-data/pull/1208, I would suggest the following remappings:

- Bowern-2017-348-9 Death adder 2654 -> remap to VIPER
- Bowern-2021-207a-10 alive 597 none Basic Vocabulary -> unmap 1422 BE ALIVE

PR #1221, opened 2022-09-08
Remappings of Bowern-2017-348 and Bowern-2021-207a

mathildavz:
As requested in Issue #1218, I've remapped two concepts in Bowern-2017-348 and Bowern-2021-207a.


AnnikaTjuka (approved):
Thanks! This can be merged as well.

PR #1224, opened 2022-09-09
Updated Chen-2012-888

mathildavz:
I linked some unlinked concepts in Chen-2012-888.


concepticondata/conceptlists/Chen-2012-888.tsv
```diff
@@ -335,7 +335,7 @@ Chen-2012-888-333	333	马鞍	saddle	677	3065	SADDLE		588	用品工具
 Chen-2012-888-334	334	鞭子	whip	678	2630	WHIP		597	用品工具
 Chen-2012-888-335	335	伞	umbrella	678	3105	UMBRELLA		583	用品工具
 Chen-2012-888-336	336	斧子	axe	678	677	AXE		605	用品工具
-Chen-2012-888-337	337	柴刀	chopper	679				633	用品工具
+Chen-2012-888-337	337	柴刀	chopper	679	2772	KITCHEN KNIFE		633	用品工具
```
AnnikaTjuka, line 338: The same gloss in [Huang-1992-1820-633](https://concepticon.clld.org/values/Huang-1992-1820-633) is mapped to CLEAVER. I'd propose to follow this mapping.

concepticondata/conceptlists/Chen-2012-888.tsv
```diff
@@ -686,7 +686,7 @@ Chen-2012-888-684	684	剁	chop into pieces	794	2149	CHOP (INTO PIECES)		1296	动
 Chen-2012-888-685	685	翻	turn over (on bed)	795	3537	TURN OVER		1310	动词
 Chen-2012-888-686	686	发烧	have a fever	795	2713	HAVE A FEVER		1305	动词
 Chen-2012-888-687	687	纺	spin (into yarn)	795	121	SPIN		1313	动词
-Chen-2012-888-688	688	放（牛）	put out to pasture	796				1316	动词
+Chen-2012-888-688	688	放（牛）	put out to pasture	796	2031	PUT OUT TO PASTURE		1316	动词
```
AnnikaTjuka, line 689: In [Castro-2015-608-383](https://concepticon.clld.org/values/Castro-2015-608-383), the same gloss is mapped and translated as GRAZE. @LinguList is it possible to disambiguate the two?

AnnikaTjuka (commented):
Thanks for adding the mappings! I have only two comments.

concepticondata/conceptlists/Chen-2012-888.tsv
LinguList: Yes, good point.

concepticondata/conceptlists/Chen-2012-888.tsv
```diff
@@ -388,7 +388,7 @@ Chen-2012-888-386	386	水缸	water tank	695				0	用品工具
 Chen-2012-888-387	387	水桶	bucket	695	2052	BUCKET		561	用品工具
 Chen-2012-888-388	388	罐子	can (jar)	696	3765	STORING POT		557	用品工具
 Chen-2012-888-389	389	水瓢	gourd ladle (wooden dipper)	696	2657	LADLE		0	用品工具
-Chen-2012-888-390	390	火钳	fire tongs	696				566	用品工具
+Chen-2012-888-390	390	火钳	fire tongs	696	272	TONGS		566	用品工具
```
LinguList, line 391: The "fire" is part of the compound, so it is a Zange that is used to get things from the fire or similar (don't know for sure). But it is not a generic tong, I'd say (tong is "Zange" in German, right?).

concepticondata/conceptlists/Chen-2012-888.tsv
LinguList: In this case, it is like "put (a cow)", so it means probably something like "put out to pasture". GRAZE is not causative in our definitoin, but this "put to pasture" is like "grasen lassen", so I'd make the disambiguation along this line. This is definitely causative here.

LinguList (approved):
See my two comments, otherwise, I already approve now.

concepticondata/conceptlists/Chen-2012-888.tsv
mathildavz: So, @LinguList, should I leave it as 2031 PUT OUT TO PASTURE? Also, if the same gloss is mapped to GRAZE in other cases, should I open a new issue to discuss this?

LinguList: Leave it for now, I think that it is close enough.

PR #1237, opened 2022-10-05
Added new concept sets from West Old Turkic to the master list

martino-vic:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information
May need some help with `concepticon notlinked --gloss "NEW_GLOSS"`
...

### DIFF
--- a/concepticondata/concepticon.tsv
+++ b/concepticondata/concepticon.tsv
@@ -3959,3 +3959,87 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3973	DEPRESSED	Emotions and values	A mental state of despondency, unhappinness, or hopelessness.	Property	
 3974	UNHAPPY	Emotions and values	Feeling not cheerful or glad, sometimes due to a misfortune.	Property	
 3975	NERVOUS	Emotions and values	A mental state of nervousness or agitation. Also, being easily excited or irritated.	Property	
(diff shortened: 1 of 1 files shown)


LinguList:
Thanks for the PR. I am sorry to only answer now, as I did not see it before. These kinds of PRs with numerous new concepts must normally come with a new concept list, and we usually do not allow too many concepts that are new to be added at once. So we need to find a use-basis for them in at least two or three conceept lists,and from eye-balling the ones you provide, I do not see them. So our procedure would rather be:

1. leave concepts unmapped
2. check sporadically, if they are central for you, if the occur in other lists
3. propose some new concepts (< 20) in a github issue and discuss with concepticon editors

LinguList:
That means, we'd close this PR and ask you to leave the corresponding entries unmapped. You can still add the definitions ina dictionary or in a specific gloss field different from Concepticon, but we cannot handle to add too many concepticon conceepts at once. This would break our workflow and also make it difficult to avoid concept duplicates, if we grow too fast at once.

LinguList:
I am sorry if it happened that you asked me about this and I gave an answer that did not warn you, as I understand it is a lot of work to make all these definitions. They are not lost, if you store them, we could even think of storing definitions in a pool-file that we then check if we encounter so far unmapped concepts that we see multiple times?

martino-vic:
No worries

The idea with the pool-file sounds good! Does it already exist somewhere, or should it rather be launched from scratch?

I'll open a new issue for the < 20 concepts now

PR #1188, opened 2022-07-12
Add Bastin-1999-92 and Zanchi-2022-170

mathildavz:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] add new Concepticon concept sets
  * [ ] checked whether the new concept(s) can be applied to existing lists with
    `concepticon notlinked --gloss "NEW_GLOSS"`
- [ ] add new Concepticon concept relations
- [ ] refine existing Concepticon concept set mappings
- [ ] refine Concepticon glosses
- [ ] refine Concepticon concept relations
- [ ] refine Concepticon concept definitions
- [ ] retire data

## Additional information

...

PR #1241, opened 2022-10-07
Kaufman 1973

Tarotis:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR will prepare the addition of Kaufman's 1973 concept list. For the digitization, there are some open problems that I want to solve before starting the mapping. @AnnikaTjuka , maybe you could provide the answers to my questions?

- In the original PDF, the author includes his own semantic domains. However, we cannot simply add a column to the file, as they are often nested, for example right in the beginning:
  - NATURE, NATURAL ENVIRONMENT (4-76)
  - EARTH (4-19) 
  - SUBSTANCES (4-9)
  - Should we ignore this semantic domains?
- The author also provides his own ID. However, those are sometimes inconsistent (e.g. typos like "TL" instead of "TK"), or an overarching term like `DRY-ADAPTED PLANTS` receives an ID, and many sub-entries which will actually be mapped to glosses do not (in this case, `YUCCA`, `CACTUS`, and others). Should we include the ID whenever possible, and remove the overarching terms?
- In some cases, the author groups multiple concepts as one. Examples are `EAST, SUNRISE`, `WHISKERS, FACE HAIR, BEARD, MUSTACHE` and `VULVA, VAGINA, FEMALE GENITALS`.  Some of those would indeed correspond to different concepticon glosses, while others wouldn't. Should we go through those cases manually and see whether they should be separated?


AnnikaTjuka:
@Tarotis I would propose adding the concept list to a Google Sheet, opening a group chat in Mattermost, and sending us a link to the file as well as the questions above in the group chat. We use GitHub only to add concept lists in their final stage when all files are prepared in the same way as in this example PR https://github.com/concepticon/concepticon-data/pull/1188

Tarotis:
Closed until finalizing the mapping.

Issue #1204, opened 2022-08-03
Fix definition for HAY


martino-vic:
Currently the defintion of HAY at https://github.com/concepticon/concepticon-data/blob/master/concepticondata/concepticon.tsv#L343 reads `Grass or cut and dried for use as animal fodder.`, I guess it should be changed to `Cut and dried grass for use as animal fodder.`?

LinguList:
Sounds right to me. Do you want to change this and make a PR? This does 
not need any additional changes, just the definition.

AnnikaTjuka:
I asked @mathildavz to create a PR for this change.

PR #1248, opened 2022-10-12
Fixed definition of HAY

mathildavz:
As suggested in Issue #1204, I changed the definition for 342 HAY.


AnnikaTjuka (approved):
This can be merged as well.

PR #1273, opened 2022-11-04
fix Gerardi-2022

xrotwang:
removed 4 duplicate concepts, and 2 more mapped to replaced conceptsets, where replacement was mapped as well.

I kept the original numbers in place - could be moved to `LEGACY_NUMBER`, though, if you think that would be better.


concepticondata/conceptlists/Gerardi-2022-705.tsv
```diff
@@ -0,0 +1,706 @@
...
+Gerardi-2022-705-413	417	PUMA	ANIMALS	1891	PUMA
+Gerardi-2022-705-414	418	PUMPKIN	AGRICULTURE_AND_VEGETATION	864	PUMPKIN
+Gerardi-2022-705-415	419	PURPLE BANANA	NA		
+Gerardi-2022-705-416	420	PUSH	MOTION	1452	PUSH
+Gerardi-2022-705-417	421	PUT	SPATIAL_RELATIONS	998	PUT
+Gerardi-2022-705-418	422	RAIN (PRECIPITATION)	THE_PHYSICAL_WORLD	658	RAIN (PRECIPITATION)
+Gerardi-2022-705-419	423	RAINBOW	THE_PHYSICAL_WORLD	1733	RAINBOW
+Gerardi-2022-705-420	424	RAT	ANIMALS	1490	RAT
+Gerardi-2022-705-421	425	RED	SENSE_PERCEPTION	156	RED
+Gerardi-2022-705-422	426	RED-AND-GREEN MACAW (Ara chloropterus)	ANIMALS	3938	RED-AND-GREEN MACAW (Ara chloropterus)
```
LinguList, line 423: My last PR, did it not remove the (Ara chloperus)?

LinguList: Seems I did not push it, but it would be nice to quickly also change this Concepticon gloss, as we have removed all other glosses with lower case now.

LinguList (approved):

LinguList:
Sorry for the trouble; last thing that is missing is the adjustment of the gloss for Macaw in concepticon.tsv

PR #1274, opened 2022-11-04
V3.0

xrotwang:


xrotwang:
@LinguList @AnnikaTjuka I added a bit of documentation here and there about what has changed and how concepticon-cldf is going to become the main way to access concepticon data. So it may be worth reading through
https://github.com/concepticon/concepticon-data/pull/1274/files

LinguList (approved):
Thanks a lot, looks good to me! One short question: Do we also create the app data, or should I do that later? I'd then also make a new release with pysem...

xrotwang:
Creating the JS app data isn't described in the release procedure and it isn't part of the release. So, I'd say you run `concepticon make_app` once the release is done.

LinguList:
Alright!

AnnikaTjuka (approved):
Great!

PR #1288, opened 2022-12-12
Add Chacon-2022-94

MuffinLinwist:
# Pull request checklist

- [x] add new concept list
- [x] add new metadata
- [ ] retire data

## Additional information

All integrity test passed and also `concepticon check`.


concepticondata/conceptlists.tsv
```diff
@@ -1,4 +1,5 @@
 ID	AUTHOR	YEAR	LIST_SUFFIX	ITEMS	TAGS	SOURCE_LANGUAGE	TARGET_LANGUAGE	URL	REFS	PDF	NOTE	PAGES	ALIAS
+Chacon-2022-94	Chacon, Thiago C.	2022		94	basic	English,Portuguese	Northwest Arawakan Languages		Chacon2022		This concept list is taken from a dataset on Northwest Arawakan Languages. The list is based on [Swadesh's list of 100](:ref:Swadesh-1955-100).
```
LinguList, line 2: "[Swadesh's list of 100 items]"

LinguList (approved):
Thanks, this looks fine for me. Maybe one more person can check quickly (as this has already been mapped before)?

Tarotis (approved):
One general curiosity I have: Should new lists and bibliography files be added at the beginning or at the end of the files? Or is that really not that important?

I have problem with some Portuguese translations (serra, pessoa civilizado, de pe, sentado) and the corresponding mapping, which do not seem to fit 100%. However, the list is based on the Swadesh list and all translations were probably based on that, so I assume that everything is correct. They are also all mapped the same as another list by Chacon.

MuffinLinwist:
> One general curiosity I have: Should new lists and bibliography files be added at the beginning or at the end of the files? Or is that really not that important?
> 
> I have problem with some Portuguese translations (serra, pessoa civilizado, de pe, sentado) and the corresponding mapping, which do not seem to fit 100%. However, the list is based on the Swadesh list and all translations were probably based on that, so I assume that everything is correct. They are also all mapped the same as another list by Chacon.

Should I put it at the end? If not, I can merge this PR.

AnnikaTjuka:
> One general curiosity I have: Should new lists and bibliography files be added at the beginning or at the end of the files? Or is that really not that important?

@Tarotis, @MuffinLinwis, @mathildavz  I'm pinging all of you because this is relevant to everyone: New lists are usually added at the end of the `conceptlists.tsv` file. It is not super important that this is done every time, but it gets confusing when we add several lists at once because you have to check if the conflict occurs at the beginning or at the end of the list. For the bibliography, I usually add the bibtex at the beginning, but again this is a recommendation.

PR #1292, opened 2023-01-13
add new concepts

### LINKED ISSUE #1282: New concepts: GUAVA and BREEZE
Based on existing entries and a new list that I will add soon, I propose to add the following entries to Concepticon:

- GUAVA
  - fruit from the myrtle family, native to Latin America
  - unmapped in three existing lists, +1 that I will add
- BREEZE
  - a gentle to moderate wind
  - possibly conflicting concepts: WIND; however, I would interpret BREEZE as a sub-entry to WIND
  - unmapped in 4 existing lists, +1 that I will add

### LINKED ISSUE #1290: New concepts: GET DRUNK and TADPOLE
Part of a list that will be added soon ([Lionnet-1985-362](https://github.com/concepticon/concepticon-data/pull/1289)), I have found that both meanings are already well represented within Concepticon, but without a Gloss.

- GET DRUNK
  - Activity of drinking alcoholic beverages that leads to DRUNK
  - already present in three lists
  - once mapped (erroneously, I would argue), to DRUNK, which, however, is only the STATE, not the Activity
  - Link to wrong mapping: https://concepticon.clld.org/values/Hale-1973-1798-1448
- Tadpole
  - Larval stage of amphibians
  - exists in five different lists already
  - no related concepts

If you @LinguList and @AnnikaTjuka both agree (especially to the re-mapping of one case of DRUNK), I would open a PR.

Tarotis:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR is intended to close #1282 and to close #1290 . Four new concepts are added, which have all at least three occurrences across different conceptlists. In one case (GET DRUNK), an existing mapping gets modified, the others were unmapped before. In the case of GET DRUNK, it was previously mapped to DRUNK, which, however, has its own entry within the same list.

@AnnikaTjuka Could I ask you to moderate or review this PR?


concepticondata/conceptrelations.tsv
```diff
@@ -793,3 +793,4 @@ SOURCE	SOURCE_GLOSS	RELATION	TARGET	TARGET_GLOSS
 619	ANIMAL	narrower	3923	FEMALE ANIMAL
 1553	MALE	narrower	3924	MALE ANIMAL
 619	ANIMAL	narrower	3924	MALE ANIMAL
+960	WIND	narrower	3982	BREEZE
```
LinguList, line 796: Careful with these relations here: We could also say "storm" is narrower than "wind", but we haven't done so so far: https://concepticon.clld.org/parameters/858 I would say, that in these cases it is better to leave concepts as separate entities. We use the narrower-broader relations only when we have clear-cut cases, such as kinship, where we have clear over-under-specification, etc., but avoid to add too many relations on concepts where it is less clear. 

The alternative is to check for wind / storm / tornado and add non-existing relations here. 

Maybe we need a discussion on this?

LinguList (approved):
Please see my comment on the very last point, the relation between wind and breeze, where I ask the others to share their opinions, as I myself am hesitant here...

concepticondata/conceptrelations.tsv
Tarotis: I'd be fine to remove this relation is well, as this seems to be what has done so far.

LinguList:
Fine :)

Tarotis:
I removed the concept relation. 

@AnnikaTjuka Do you also approve of the changes made?

concepticondata/conceptrelations.tsv
AnnikaTjuka: I would also argue against adding the relation. We have too many weather-related concepts, which then also need a thorough check, and I'm not sure that we will find many clear cases.

AnnikaTjuka (approved):
Looks good! Many thanks for adding the concepts.

tuffnatty:
@Tarotis In this commit, at least concepticon.tsv entries are broken - they have 5 columns each instead of 6.

Tarotis:
Thanks for the note, we've fixed this in #1334

PR #1320, opened 2023-06-08
add Auderset et al. 2023

SimonGreenhill:
# Pull request checklist

- [X] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

- no page numbers, the word list is in the supplementary material.


SimonGreenhill:
PDF is available [here](https://zenodo.org/record/6513506) as "sm1"

LinguList (commented):
I am very confused now: this concept list is mentioned in the paper by Auderset et al. 2023 in JOLE, but the concept list listed in the CLDF dataset has 240 entries, and seems to be entirely different, with capital letters, written, etc.? 

https://github.com/lexibank/mixtecansubgrouping/blob/main/etc/concepts.tsv

I mean, this does not really make sense: there's a full publication mentioning a list of 209 items, and the CLDF data then has 240 but the paper says it is based on the CLDF data?

LinguList:
In these cases, we typically fix the errors of the data creators / authors by combining both lists. This results, however, also in a re-structuring of the CLDF. Or is the CLDF supposed to be for another paper?

xrotwang:
Btw.: `etc/concepts.tsv` has 279 items. It's https://github.com/lexibank/mixtecansubgrouping/blob/main/cldf/parameters.csv which has 240 - presumably because there's no lexical items for the other 39?

SAuderset:
The 209 number refers to the custom word list I used to collect the data that forms the basis of the JOLE paper. This is the list provided in "sm1" (linked by Simon above). I remember that this list contained quite a few items that were not in Concepticon (at least not two years ago), so so some of the items would not be linked to any concept.
The concept list extracted from the data set has more concepts, because the glosses (or concepts) listed sometimes had to be adjusted slightly to stay faithful to the source material.

LinguList:
Well, in the paper, you mention 209 concepts, but the cldf dataset has 279 now, so how do we explain this discrepancy? Is the lexibank/mixtecansubgrouping now a completely different dataset or is it the same data, and you use a subset that you did not indicate when creating the CLDF representation of your data? Or is it indicated anywhere?

LinguList:
I'd like to ask you, @SAuderset, to please prepare a new concepts.tsv for us, in which you make sure to merge the list that your reference in paper and supplementary material in an extra column (or more), so we have ONE representation for the concepts. Based on this, we could then proceed and unify this and make sure that the discrepancy is handled properly.

SAuderset:
I'm not sure I understand what exactly the problem is. 
The 209-item list contains only English and Spanish search terms, which I used to gather the Mixtecan entries. So it's like a Swadesh or Leipzig-Jakarta list, just customized to fit better with the Mesoamerican cultural area.
The concept list derived from the CLDF-version of the data is extracted from the full cognate-annotated data base. There are more concepts there than the original 209 search terms, because there were a few adjustments necessary I didn't foresee when creating the list of search terms (e.g. sibling terms in Mixtecan are usually differentiated by gender of ego, but when collecting data I ran into some languages that do not make this distinction, so I had to add 'sister' in addition to 'sister of woman' and 'cross-sex-sibling').
Maybe only the 209-item search term list should be added to Concepticon then?

LinguList:
The problem is that you propose a paper with data restricted to 209 concepts, but the data in CLDF does not provide this restriction. So it needs to be added post-hoc, as it is not replicable in this way, obviously, if we do not have any link between the 209 concepts in the paper and the data that is on github/zenodo and in CLDF. So we need this combined list from you, to update the CLDF, and we'll based on this then decide if the CLDF also has to be restricted to the 209 concepts, of if a flag is enough to make sure that the commands producing nexus and the like work properly.

SimonGreenhill:
I think the best way to treat this as
1. Auderset et al. 2023 has a concept list (provided in this PR) as the full dataset was pruned for phylogenetic analysis in this paper.
2. We keep the other dataset (with its concept list) as the 'full' dataset, and when other papers come out from Sandra's paper we make new datasets for those.

LinguList:
But if the concept list has no relation whatsoever to the original CLDF dataset, the original idea of concepticon to link between the real data and the "talking about data in publications" is lost. So I think there is no way around re-publishing CLDF data with this 209 item concept list. Whatever new publications with more data come, they would be a new dataset, since the data is anyway also moving, as @SAuderset did not stop working on the data, as far as I understand.

LinguList:
This is also the way we have proceeded in the past with many other cases.

LinguList:
If @SAuderset cannot provide the combined list, we'll ask our student assistants to help.

xrotwang:
I'd agree with @LinguList here. So if the 209 item list (attached above) is the one that describes the data used in the analyses of Auderset et al. 2023, then the dataset at
https://github.com/lexibank/mixtecansubgrouping should match that list (i.e. just include the data for the 209 concepts). Datasets that have a specific source or are attached to a specific paper shouldn't be managed like standalone datasets where each release may add more data (such as ABVD or Transnewguinea).

LinguList:
@SAuderset, thanks, we'll add all of these as explanations to the concepticon version. However, for the future, it seems better that you pay attention to also describe your data clearly in the paper itself. After all, I assume you agree that scientific research should be replicable without these larger inquiries post-publication...

LinguList:
So my question is now, @SAuderset: what is the role of these 209 concepts, are these really the ones you used for the analysis? And what about the data in the CLDF dataset? I really think it is important to clarify these issues quickly, since if we cannot clarify it, it means that we have a huge issue with wrong data descriptions in a published study here.

xrotwang:
So, overall, I'd say that while the 209 item list may have some historical value, the concept list with the practical value (giving access to actual lexical data) is the 240 item list. Thus, I think the latter list should be
- a candidate for inclusion in Concepticon and
- the concept list referenced in the CLDF dataset.

LinguList:
In addition, I made a manual comparison of the 209 list and the CLDF  list, with 278 items. It turns out, I can match almost all cases but 2, maybe, @Simon
Greenhill and @SAuderset as the ones who used the data, you can find these?

[concepts.xlsx](https://github.com/concepticon/concepticon-data/files/11733715/concepts.xlsx)

LinguList:
With these mappings, we can add two columns of glosses in concepticon and make sure to make the connection between the 209 item list and the 240 list explicit.

SAuderset:
I think @xrotwang 's proposal is the best and easiest way to go, i.e. to only link the 240-list (the one extracted from the data) to Concepticon. If needed, a note could be added that the 209-item list was used as a "guide" so to speak, but forms with very similar meanings were also included in the final data set.

xrotwang:
I can take a look at that. I'd also update the CLDF dataset to work from the latest release of the supplemental material on Zenodo, ok @SAuderset ?

LinguList:
Since I already did the mapping, with two cases remaining, where I hope 
you can help me, let us use the combined list. Otherwise, I did this 
work also for nothing. I'd just need a quick check, where I missed to 
find the corresponding element for the two concepts in the 209 list with 
the 278 list.

SAuderset:
I don't know why these two concepts are missing, they are in the data (sm3_mixtecan_cognates.tsv) as BLAME and EAR OF CORN (FRESH).

xrotwang:
> I don't know why these two concepts are missing, they are in the data (sm3_mixtecan_cognates.tsv) as BLAME and EAR OF CORN (FRESH).

The matching didn't work for these concepts, because they had no spanish gloss listed - instead, the english gloss was repeated in the `Spanish` column.

LinguList:
So `sm3_mixtecan_cognates.tsv` is the file we refer to now also in CLDF conversion, right, @xrotwang ?

xrotwang:
@LinguList yes, more specifically, [sm3_mixtecan_cognates.tsv](https://zenodo.org/record/7940497/files/sm3_mixtecan_cognates.tsv?download=1) here https://zenodo.org/record/7940497

SAuderset:
Ah sorry, my bad for overlooking that. The Spanish Glosses should be 'culpa' for BLAME and 'elote' for EAR OF CORN (FRESH).

SimonGreenhill:
Sorry for slow reply, been tied up with other things. The reason I wanted the 209 list in concepticon is so Phlorest can point to it (the trees and nexus data published there are the 209 variant). Can we add both and mark the 209 list as a subset of the bigger one?

SimonGreenhill:
btw: in the 240 list, item 20 ("CROSS-SEX SIBLING") should not be mapped to `1263 SISTER`

LinguList:
then my original plan of having a merged list is best.

LinguList:
This then also links to the publication in the best way.

LinguList:
Dunno when I can look into this, but it requires to re-create CLDF and teh list here, so it is some work.

xrotwang:
> The reason I wanted the 209 list in concepticon is so Phlorest can point to it (the trees and nexus data published there are the 209 variant). Can we add both and mark the 209 list as a subset of the bigger one?

But that's not the case. AFAICT, the phlorest data is based on the 240 concept list as well (as it should, since that's what the paper links as supplemental material):

```python
>>> from csvw.dsv import reader
>>> names = [r['Name'] for r in reader('phlorest/auderset_et_al2023/cldf/parameters.csv', dicts=True)]
>>> names[:2]
['0ascertainment', 'ALLDAY_1']
>>> len(names)
1183
>>> concepts = ['_'.join(n.split('_')[:-1]) for n in names]
>>> concepts[:2]
['', 'ALLDAY']
>>> len(set(concepts))
241
```

LinguList:
I did not have time to check this recently, but my suggestion now is to go for the 240 concept list. I'd update that now, try to fix this soon.

LinguList (approved):

Issue #1285, opened 2022-12-06
Move mappings from 2309 LIVE to 1422 BE ALIVE


AnnikaTjuka:
The concept LIVE is a duplicate so the following lists need to be remapped:

- [Greenhill-2015-2525-148](https://concepticon.clld.org/values/Greenhill-2015-2525-148)
- [Starostin-2017-401-192](https://concepticon.clld.org/values/Starostin-2017-401-192)

PR #1326, opened 2023-07-13
Remapped 2309 LIVE to 1422 BE ALIVE

mathildavz:
As discussed in #1285, I've remapped the duplicates in two lists from 2309 LIVE to 1422 BE ALIVE.


AnnikaTjuka (approved):
Looks good! Could you link the issue under "Development" so that it is closed automatically when the PR is merged?

PR #1334, opened 2023-10-27
fix rows for new concepts

Tarotis:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

Not sure which category this fits in. As commented in the last PR (#1292 ) which is already merged, the new concepts miss the empty sixth row. This PR fixes this. Here is the comment: https://github.com/concepticon/concepticon-data/pull/1292#issuecomment-1782795158

This is not captured by the `concepticon test`, but I also don't know if this actually breaks anything. Probably still safer to fix this. @xrotwang or @AnnikaTjuka Could you briefly check if everything is okay?


AnnikaTjuka (approved):
Ah yes, it depends on the software that was used to edit the concept list. It doesn't break anything, but the rendering in GitHub into a table doesn't work if the empty rows are missing. Feel free to merge.

PR #1346, opened 2023-12-20
Winter 2022

LinguList:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

I detected a bug which is annoying, in my logic of parsing the data. This results in undirected graphs being inherently still directed, since JSON is only assigned to a part of it. So I try to fix these issues now in this PR for the List-2023-1308 list.


concepticondata/conceptlists/List-2023-1308/convert.py
```diff
@@ -39,6 +39,7 @@
     ])
 
 valid_edges = collections.defaultdict(list)
+selected_edges = collections.defaultdict(list)
```
xrotwang, line 42: Where does this come from now? Like the `common_edges` in a previous version, this variable doesn't seem to be used anywhere.

xrotwang (changes requested):

PR #1381, opened 2024-03-18
fix seifart 410 to 408

### LINKED ISSUE #1380: Duplicates in Seifart-2015-410
We missed two duplicated while adding the above mentioned concept list: BLOOD (26 + 205) and FULL (232 + 346). This also means that the concept list only has 408 entries, and not 410. I will soon open a PR to fix this.

FredericBlum:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

This PR fixes #1380 , two duplicates that we had previously missed. Are there precedents for renaming existing conceptlists? Since there is only one dataset where this is used (seifartecheverriboran) and this is only by ourselves, it will probably be not be a problem for other users.


chrzyki:
Sorry, missed that PR. Yes, you'd do a

```bash
$ concepticon rename Seifart-2015-410 Seifart-2015-408
```

and then fix the list/items accordingly. The only other thing that `rename` does in addition to the things you've done manually here (fix IDs etc.) is that it populates `retired.json` with mappings from old to new IDs (in order to keep references stable, since that list has already been published). This includes mapping the old to the new concept list ID in `retired.json`:

```diff
@@ -30091,6 +32141,11 @@
       "id": "Yakhontov-1991-35",
       "comment": "renaming",
       "replacement": "Yakhontov-1956-35"
+    },
+    {
+      "id": "Seifart-2015-410",
+      "comment": "renaming",
+      "replacement": "Seifart-2015-408"
     }
   ]
 }
```

And individual mappings for all entries in a concept list:

```diff
diff --git a/concepticondata/retired.json b/concepticondata/retired.json
index 3b40dc8e..899d13ea 100644
--- a/concepticondata/retired.json
+++ b/concepticondata/retired.json
@@ -29949,6 +29949,2056 @@
       "id": "Yakhontov-1991-35-35",
       "comment": "renaming",
       "replacement": "Yakhontov-1956-35-35"
+    },
+    {
+      "id": "Seifart-2015-410-1",
+      "comment": "renaming",
+      "replacement": "Seifart-2015-408-1"
+    },
+    {
+      "id": "Seifart-2015-410-2",
+      "comment": "renaming",
+      "replacement": "Seifart-2015-408-2"
+    },
...
```

FredericBlum:
Thanks for the tip, I will re-do the changes then

FredericBlum:
I have updated the PR using the `renaming' command. Following strictly the described workflow, the list gets updated, but the individual entries do not. This means that the last entry has the following ID:

`Seifart-2015-408-410 410 conjure 2222 CONJURE 562`

I assume that this is working as intended, so that the ID of the entry does not get changed for backwards compatibility?

chrzyki:
> I have updated the PR using the `renaming' command. Following strictly the described workflow, the list gets updated, but the individual entries do not. This means that the last entry has the following ID:
> 
> `Seifart-2015-408-410 410 conjure 2222 CONJURE 562`
> 
> I assume that this is working as intended, so that the ID of the entry does not get changed for backwards compatibility?

Yes, here the manual intervention mentioned above is required: Since the number of items in the list was reduced, you need to manually fix those two cases (and make sure that the `retired.json` mapping still make sense, given the change in order and depending where in the list that has happened).

FredericBlum:
I am not sure I can follow you. From what I've read above, I should run the changes in the following order:

1. Run `rename' - the `retired.json' file gets populated
2. Fix the ID's

But then the retired.json-file does not have the updated ID's, and I'd have to make 200+ manual changes to retired.json. Is there another option to re-populate the retired.json?

If I first change the ID's and remove the duplicates, the remaining ID's are correctly displayed, but there is no information about the two deleted lines. What happens if any project had one of the duplicates as ID?

Example: I deleted Seifart-2015-205

But since I had to do the changes BEFORE running `rename', the entry in retired.json refers to the NEW 205, not the old one. So any reference to the old Seifart-2015-205 is now wrongly referred to as the new Seifart-2015.

Also, there is another problem: Running rename seems to overwrite some previous changes that have been made to retired.json, namely referring to a conceptlist from Yakhontov - how can I avoid this?

FredericBlum:
@chrzyki Thanks for the detailed lead-through. I have now fixed the list according to your guidelines, and it seems like everything worked. Can you give it a final check?

concepticondata/retired.json
```diff
@@ -29949,7 +29949,2057 @@
...
+	},
+	{
+	  "id": "Seifart-2015-410-26",
+	  "comment": "renaming",
+	  "replacement": "Seifart-2015-408-26"
+	},
+	{
+	  "id": "Seifart-2015-410-27",
+	  "comment": "renaming",
+	  "replacement": "Seifart-2015-408-26"
```
chrzyki, line 30086: The remapping is unfortunately wrong here (did you generate IDs before removing the doubles?): this maps the old ID for `HEART` to the new ID for `BLOOD` and propagates the wrong ID throughout the following entries.

concepticondata/retired.json
```diff
@@ -29949,7 +29949,2057 @@
...
+	  "comment": "renaming",
+	  "replacement": "Seifart-2015-408-230"
+	},
+	{
+	  "id": "Seifart-2015-410-232",
+	  "comment": "renaming",
+	  "replacement": "Seifart-2015-408-230"
+	},
+	{
+	  "id": "Seifart-2015-410-233",
```
chrzyki, line 31114: Likewise further down below.

FredericBlum:
Ah, I see the error now... I have removed the second occurrence of each concept, but your script assumes that I remove the first. I will change the code and update `retired.json` accordingly

FredericBlum:
@chrzyki The latest commit should contain the correct replacements

chrzyki (approved):
Looks good to me, thank you very much for taking care of this!

PR #1394, opened 2024-07-02
EDIT: Gänsefüßchen weg! 

alzkuc:
Hi Mattis, 
as you can see now, I have edited the quotation marks out. You just have to accept the pull request now. :-) @LinguList


LinguList:
The problem is now: if I accept. I merge directly, without review.

LinguList:
So, @alzkuc, what I suggest I do is that I manually adjust this in the first PR and delete this one, okay?

alzkuc:
> So, @alzkuc, what I suggest I do is that I manually adjust this in the first PR and delete this one, okay?

okay, sure, but that is additional work for you! thank you, @LinguList

PR #1399, opened 2024-08-08
mapping concepts

MuffinLinwist:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

...

Here is an update of this concept list increasing significantly the number of mapped entries.


chrzyki (commented):
Sorry, I'm a bit confused about the layout of this study/its concepts: the appendix of the thesis contains the 275 concepts that have been mapped previously. This PR updates the concepts to everything that was used in the cognate list, correct? What's the relation between the previous (amount) of concepts and the updated list? Apologies if this is made clear in the thesis, haven't had time to check that thoroughly.

LinguList:
The thesis mentions 275 concepts, the DATA however, consist of 1800 concepts or more. So we have a discrepancy that is somehow problematic. My suggestion would be to stick with the 275, see my [NOTE](https://github.com/lexibank/lundgrenomagoa/blob/master/NOTES.md). 

But @MuffinLinwist, the mappings you provide in fact are useful but should then go into a NEW concept list that is published alternatively to the 275 meanings and used. We would then start from this new concept list in the lexibank script?

chrzyki:
> The thesis mentions 275 concepts, the DATA however, consist of 1800 concepts or more. So we have a discrepancy that is somehow problematic. My suggestion would be to stick with the 275, see my [NOTE](https://github.com/lexibank/lundgrenomagoa/blob/master/NOTES.md).
> 
> But @MuffinLinwist, the mappings you provide in fact are useful but should then go into a NEW concept list that is published alternatively to the 275 meanings and used. We would then start from this new concept list in the lexibank script?

Thanks for the clarification. That sounds like a good plan!

PR #1411, opened 2024-09-30
Added two new concepts (piano (instrument), top (toy))

alzkuc:
two new concepts: PIANO (INSTRUMENT) and TOP (TOY) 

# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [X] add new Concepticon concept sets
- [ ] retire data

## Additional information

...

### DIFF
--- a/concepticondata/concepticon.tsv
+++ b/concepticondata/concepticon.tsv
@@ -3977,3 +3977,5 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 3991	TORTILLA	Food and drink	A thin, flat bread made from maize flour, eaten hot or cold, typically with a savoury filling.	Person/Thing	
 3992	SARDINE	Animals	A pilchard or other fish in the herring family.	Person/Thing	
 3993	JINGLE (INSTRUMENT)	Speech and language	One of a cluster of small bells, shaken or tapped on the palm of the player's hand. The small metal discs arranged around the frame of a tambourine are also called "jingles".	Person/Thing	
+3994	PIANO (INSTRUMENT)	Speech and language	A large musical instrument with a row of black and white keys that are pressed to play notes	Person/Thing
+3995	TOP (TOY)	The physical world	A toy with rounded sides, a flat top and a point at the bottom that turns round and round on the point.	Person/Thing


LinguList:
@alzkuc, can you elaborate why you suggest to add these? Are they already often used? Are they useful for many lists that you intend to map, etc?

LinguList:
We'd then discuss if it is worth adding the new concepts here.

alzkuc:
Hallo @LinguList! Ja genau, danke für die schnelle Reaktion. Diese zwei Konzepte kommen in sehr vielen Listen vor, deshalb wollte ich sie hinzufügen, wenn du damit einverstanden bist. :-)

LinguList:
Are they also in existing lists? This makes their integration easier. Could you share, since we will discuss this with other editors, like @AnnikaTjuka and @FredericBlum, some arguments and some summary on the concept definitions?

alzkuc:
> Are they also in existing lists? This makes their integration easier. Could you share, since we will discuss this with other editors, like @AnnikaTjuka and @FredericBlum, some arguments and some summary on the concept definitions?

Hi @AnnikaTjuka and @FredericBlum: whenever you have time, would you please be so kind and let me know your thoughts? I would like to add two new concepts to concepticon.tsv. Thanks a lot in advance! :-) 

Here's a summary:

PIANO (INSTRUMENT): A large musical instrument with a row of black and white keys that are pressed to play notes.
TOP (TOY): A toy with rounded sides, a flat top and a point at the bottom that turns round and round on the point.

Sure! I think it would be useful to add these two new concepts—PIANO (INSTRUMENT) and TOP (TOY)—to the concepticon.tsv, as they appear in many of the lists I have recently mapped. I did a simple search for the word _piano_ in all the lists, and it comes up in 10 of the existing ones as well as 13 of my own, totaling 23. The word top (toy) appears in 11 of my lists, indicating that it is a common object to include in picture-naming tasks. Additionally, the top is not a modern world item, it is considered one of the oldest toys in the world, therefore more likely to appear in other (of my) wordlists as well.

AnnikaTjuka:
@LinguList @alzkuc The concept PIANO also exists in other lists, but we have not added it yet because it is not necessarily relevant for the language comparison, since most languages use the original word ‘piano’. I could not find the concept TOP (TOY) in other lists. I would therefore only add PIANO and not TOP (TOY). 

@alzkuc A more general note: It would be great if we could initially discuss adding new concepts in an issue so that we can decide on possible mappings. The PR should be the last step and would include mappings to all existing Concepticon lists.

alzkuc:
> @LinguList @alzkuc The concept PIANO also exists in other lists, but we have not added it yet because it is not necessarily relevant for the language comparison, since most languages use the original word ‘piano’. I could not find the concept TOP (TOY) in other lists. I would therefore only add PIANO and not TOP (TOY).
> 
> @alzkuc A more general note: It would be great if we could initially discuss adding new concepts in an issue so that we can decide on possible mappings. The PR should be the last step and would include mappings to all existing Concepticon lists.

Alright, thanks for your reply! @AnnikaTjuka Next time, I will create an issue with these first. Thank you for the note on which word is usually used for the concept. “Top,” when it appears, is typically expressed by a unique word in most languages rather than a borrowing. This is another reason I thought it might be useful to add it.

AnnikaTjuka:
I'm closing this PR for now.

PR #1419, opened 2024-11-08
Added concep list Needham-1897-263

patkaiist:
# Pull request checklist

- [x] add new concept list
- [x] add new metadata
- [ ] retire data

## Additional information

Concepts from Needham 1897 mapped to Concepticon concepts based on the author's intended targets.


patkaiist:
@LinguList Could you review this for me? Thank you.

LinguList (changes requested):
Thanks, a couple of things that must be modifiied please:

- [ ] Can you please reinsert the Lexibank_Gloss? We need it, as it links the concept list to your data.
- [ ] Also: we need an explanation for ditto-cases, the asterisk. 
- [ ] Also: you are not giving us the literal form: you write `Bury, v` but it should be `Bury, v.`.  This is indeed very important for Concepticon to stick to the true value we find in the source. (Chillic has quotes, etc.)

If we have these covered, I make a thorough review afterwards.

patkaiist:
I will check again on asterisks and dittos; I thought I caught them all. I will also make sure the transcription of the glosses is exact, e.g. chillic.

For `Lexibank_Gloss`, I removed it since this is Needham's attempted concepts, not the ones the speakers actually returned. The concepticon mappings here are what Needham intended. I had planned to submit my corrected version of his data as a modification of the original, as the Swadesh modifications currently exist in Concepticon. For that, the Concepticon mappings will be different, since they will match the actual concepts given. That is what I expected the Lexebank data to mater for. That's what I was trying to clarify on Mattermost yesterday.

I had understood that we were adding Neeham'd concept list first. Then my corrections of his from van Dam & Mossang 2025 as a second, which in turn (the latter) would be submitted to Lexibank as the language data. That way the historical data would still be correctly linked to it's meaning.

LinguList:
There are two things to be considered here, but what is really important 
is that we avoid making things more complicated.

Can you please provide me with a list of cases where you think that a 
concept A was elicited, but a word meaning b was returned? Here, please? 
The number of cases is important. I decide how to proceed from there.

LinguList:
I will close this PR now and move to [this one](https://github.com/concepticon/concepticon-data/pull/1420) where I typed off the data another time to have a version that is also linked to the data that would be added to Lexibank. This also addresses the discussion we had here.

PR #1420, opened 2024-11-08
adding needham

LinguList:
# Pull request checklist

- [x] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information

@patkaiist, please check my PR here, where I add Needham in a version compatible with Lexibank, solving your problem. We must review the mappings now, but that is what I meant.


patkaiist:
@LinguList This seems to mostly solve it. I've only skimmed it now, but will look at it more fully Monday morning and then request the changes for mappings that may be needed. Thank you.

patkaiist:
@LinguList Thank you, I have gone through this in more detail now. I want to clarify the process slightly. Let me know if this is correct:

1. First, the concept mapping represents the actual data which Needham collected, not his intended concepts. This is because the data itself goes into Lexibank and we only have the one concept list for Concepticon. 
2. Then, if we add my modern data, as well as the two "dialects" Needham recorded, that still links to the same `LEXIBANK_GLOSS` values, thus tying it all together.

In this case the only thing I need to do is add a couple more glosses. _Petticoat_ is actually simply `SKIRT`, _Woman's “chudder“_ is BLANKET, et cetera.

If the numbered points above are correct, then I will do that, and afterward we can finalise the PR.

LinguList:
@patkaiist, so if I understand properly you'd say that Needham writes down he asked for "petticoat" but he received the word for "skirt" and you deduce this from the modern fieldwork?

LinguList:
I think in this way, we are true to the source, true to the interpretation, and open for including this in Lexibank with the more recent data.

patkaiist:
I think it may be fine to merge. I wanted to check on the tones in the orthography profile but I looked earlier today and everything is accounted for.

LinguList:
Okay, then if you could please resolve conflicts, and merge yourself? I'd then check next week the lexibank integration, so this is good for your publication!

LinguList:
Just resovled the conflicts on my side.

PR #1480, opened 2025-08-27
New Concept Set 4074 DELPHINOID + refinement of 114 PORPOISE and 1479 DOLPHIN definitions and relations

alzkuc:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [x] refine Concepticon concept relations
- [ ] retire data

## Additional information

- Adding a new broader concept set for the superfamily of DELPHINOID, into which both dolphins and porpoises belong.
- Refining the relations accordingly (DOLPHIN and PORPOISE as instances of DELPHINOID)
- Refining definitions to be biologically more accurate


alzkuc:
Hi @LinguList, this is the discussed PR. Would you please be so kind as to either moderate or review this really quick? Thank you!

LinguList (changes requested):
Is *delphinoid* a term that is scientifically used as such? If not, you should drop it. Delphinoid also sounds utterly German, should it not be DOLPHINOID in English? We'd then have DOLPHIN OR PORPOISE. If you find the scientific term for the group (Wikipedia has good information on this), use that term, if it is not pure Latin, and put `DOLPHIN OR PORPOISE` in brackets after it in the gloss (we order terms alphabetically in these OR glosses).

alzkuc:
> Is _delphinoid_ a term that is scientifically used as such? If not, you should drop it. Delphinoid also sounds utterly German, should it not be DOLPHINOID in English? We'd then have DOLPHIN OR PORPOISE. If you find the scientific term for the group (Wikipedia has good information on this), use that term, if it is not pure Latin, and put `DOLPHIN OR PORPOISE` in brackets after it in the gloss (we order terms alphabetically in these OR glosses).

I can definitely add the gloss (DOLPHIN OR PORPOISE)!

PR #1495, opened 2025-10-30
Update Dunabeitia-2022-500 (MultiPic)

alzkuc:
This is the most recent version of the MultiPic Dataset. It now includes all new language varieties as well as corrected values. Relevant files were updated accordingly.

- [x] update existing concept list
- [x] update metadata


LinguList (commented):
@alzkuc, didn't we want to just use the normal concept list here and add the individual data on NoRaRe? We can merge it here now, but I'd later revert it to the form that seems more apt then...

PR #1515, opened 2026-03-23
add Matter-2021-96

FredericBlum:
# Pull request checklist

- [x] add new concept list
- [x] add new metadata
- [ ] retire data

## Additional information

Adding a new conceptlist that is used in a Comparative Cariban Database. Its mostly the Swadesh-1996-100a list, minus four concepts, and remapped four of them, so this should be fairly quick and non-polemic.

@AnnikaTjuka Can you do moderate the PR? @RebeccaLethgo Can do the review if necessary.


FredericBlum:
Thanks for the moderation @AnnikaTjuka and for the review @MiraAhmedovic

Issue #1418, opened 2024-10-30
15 Object Naming Lists: A Collaborative Review


alzkuc:
Hi @AnnikaTjuka, @FredericBlum, @xrotwang, @MuffinLinwist!

@LinguList and I have decided that the best approach is to extract all concepts from the 15 newly mapped object naming datasets and review them in batches collaboratively. We would be very grateful if each of you could help us review a portion of the final list. Please let us know by giving a thumbs-up here, whether you can help, and I will assign you a manageable section of the prepared Google Sheet.

Thank you so much in advance! Your feedback and expertise are greatly appreciated.

If you have any additional comments or questions, please leave them here or contact me via GitHub or Mattermost.

FredericBlum:
What is the timeline for this? I could probably only help in the week of November 25th, because I have some obligatory classes to attend in the next few weeks which will leave me with few time for other things.

alzkuc:
Hi all! Thank you for the first reviews. Here are some points that came up:

**Suggestion 1:** Both @LinguList and @AnnikaTjuka agree (and so do I) that scarf in the lists refers specifically to a SCARF and not SCARF OR TURBAN. However, we currently do not have a concept set SCARF. Would you like to create a new one - SCARF - or should we just delete the mapping for now? Scarf appears in 5/15 lists and refers to the scarf worn in winter.

**Suggestion 2:** Then we have the example of bottle vs small bottle. I would personally stick to mapping bottle (boca) to BOTTLE and leaving small bottle unmapped, as it refers to a very specific type of bottle—the author there means a _baby bottle_ with small bottle.

**Suggestion 3:** SWORD vs BROAD SWORD: Both these terms were provided by the author as English translations in a Mandarin Chinese naming list. The authors explicitly distinguish between 刀 (translated as broad sword) and 剑 (sword). Mattis suggested mapping BROAD SWORD to SWORD as well. What do the others think? (pictures attached for reference)

**Question:** What about the !NEWCONCEPTs, should we add some of them or not for now? :-) 

Thank you for your thoughts and time! 

<img width="867" alt="sword_broadsword" src="https://github.com/user-attachments/assets/1f505f28-cd9e-44c6-99a6-033f1d3654e5">

alzkuc:
Two new points raised:
**Suggestion 4:** FOOTBALL (a specific type of ball, not the sport) comes up in many of the lists. Should we add it as well?
[...]

PR #1522, opened 2026-04-23
New concepts

MiraAhmedovic:
# Pull request checklist

- [ ] add new concept list
- [ ] add new metadata
- [ ] retire data

## Additional information
Adding 18 new concepts:

- APARTMENT
- BAGPIPE
- BAT (BASEBALL)
- BEETROOT
- CABIN
- CATHEDRAL
- CHAPEL
- CLAMS
- CLARINET
- CORK
- CORKSCREW
- GOLDFISH
- HOSE
- SKYSCRAPER
- SLIPPERS
- TUNA
- YACHT
- FOOTBALL

Further, the data format of some entries in Janschewitz-2008-460 was refined since it was faulty.
Regarding concept 4161 HOSE: Hill-2015-999-1280 also mapped because in Vulic-2020-2244-1507 we get HOSE for car and garden. Because this means that the term is already quite ambiguous, I decided to map it both times in Hill as well.
Regarding concept 4163 SLIPPER: I did not map Schroeder-2012-824-326 because the term might refer to an outside shoe and that's not what we usually consider a slipper.
Regarding 4166 FOOTBALL: I mapped both Schroeder-1012-824-573 and Schroeder-1012-824-574 because the description of this new concept does not differentiate American vs. European Football.


MiraAhmedovic:
@AnnikaTjuka, could you moderate this PR?

AnnikaTjuka:
Sure. @alzkuc and @FredericBlum I think you were involved in some of the new concept proposals. Could you take a look at the PR?

alzkuc:
> Sure. @alzkuc and @FredericBlum I think you were involved in some of the new concept proposals. Could you take a look at the PR?

Sure, no problem!

concepticondata/concepticon.tsv
```diff
@@ -4130,4 +4130,22 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
 4145	WRENCH	Basic actions and technology	A hand tool with a shaped jaw or adjustable head used for gripping, turning, or tightening nuts, bolts, and other fasteners.	Person/Thing
 4146	SLINGSHOT	Warfare and hunting	A handheld weapon consisting of a forked frame with elastic bands used to propel small projectiles, such as stones or pellets, by tension and release.	Person/Thing
 4147	CEDAR	Agriculture and vegetation	A large evergreen coniferous tree of the genus Cedrus or related genera, valued for its aromatic, durable wood and needle-like leaves.	Person/Thing
-4148	TENSE	Emotions and values	Characterized by mental or muscular strain, stress, or heightened alertness; can describe emotions, physical states, or situations.	Property
\ No newline at end of file
+4148	TENSE	Emotions and values	Characterized by mental or muscular strain, stress, or heightened alertness; can describe emotions, physical states, or situations.	Property
```
FredericBlum, line 4133: How do we disentangle this from TIGHT? Should we separate the physical from the emotional?

concepticondata/concepticon.tsv
```diff
@@ -4130,4 +4130,22 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+4149	APARTMENT	The house	A set of rooms forming a self-contained housing unit within a larger building, used as a residence.	Person/Thing
+4150	BAGPIPE	Basic actions and technology	A musical instrument consisting of a bag that supplies air to one or more pipes, producing continuous tones when played.	Person/Thing
+4151	BAT (BASEBALL)	Modern world	A solid, typically wooden or metal club used by a player to hit the ball in the sport of baseball.	Person/Thing
+4152	BEETROOT	Agriculture and vegetation	A cultivated root vegetable of the variety Beta vulgaris with a round, typically dark red bulbous root, used as food.	Person/Thing
+4153	CABIN	The house	A small, simple, and often roughly built shelter, typically made of wood and located in a remote or rural area.	Person/Thing
+4154	CATHEDRAL	Religion and belief	A large and principal church that serves as the official seat of a bishop within a Christian diocese.	Person/Thing
+4155	CHAPEL	Religion and belief	A small place of Christian worship, typically serving a private community or institution	Person/Thing
```
FredericBlum: Missing the point after the description.

concepticondata/concepticon.tsv
```diff
@@ -4130,4 +4130,22 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+4160	GOLDFISH	Animals	A small freshwater fish of the carp family, commonly kept as an ornamental pet in aquariums and ponds.	Person/Thing
+4161	HOSE	Basic actions and technology	A flexible tube used to convey liquids or gases, typically water, from one place to another.	Person/Thing
+4162	SKYSCRAPER	Modern world	A very tall, continuously habitable building with many floors, designed for residential, commercial, or mixed use in urban areas.	Person/Thing
+4163	SLIPPER	Clothing and grooming	A soft, lightweight indoor shoe typically worn for comfort and warmth at home.	Person/Thing
+4164	TUNA	Animals	A large marine fish of the genus Thunnus, known for its streamlined body and importance as a commercial food species.	Person/Thing
+4165	YACHT	Modern world	A recreational boat or vessel, typically medium to large in size, used for leisure cruising or racing on water.	Person/Thing
+4166	FOOTBALL	Modern world	A sport played by two teams in which players move a ball toward a goal to score points, mainly by kicking or carrying it OR the ball used in that sport.	Person/Thing
```
FredericBlum: Are we fine with this meaning both the sport and the ball itself? I fear that this could create problems in the long run.

FredericBlum (commented):
I have three comments on the proposed concepts which could be worth discussing. But perhaps I am missing previous discussions on those questions?

concepticondata/concepticon.tsv
FredericBlum: > description of this new concept does not differentiate American vs. European Football.

We should. But we don't right now, correct?

FredericBlum: https://github.com/concepticon/concepticon-data/issues/1418

I just found this discussion. @LinguList this involves your proposal to have a single concept that refers to both object and game. I have to say, I am not convinced. Was there some general agreement to such cases before?

concepticondata/concepticon.tsv
MiraAhmedovic: That was the idea here, yes. Should it be made more concrete?

concepticondata/concepticon.tsv
AnnikaTjuka: @FredericBlum Thanks for the review! @MiraAhmedovic Could you ping @LinguList on Mattermost regarding this discussion? We also wondered how we should handle this, but we couldn’t find a previous concept set to use as a guide.

MiraAhmedovic: I wasn't aware of that discussion, thank you for bringing it up! Since we don't have a lot of sports in Concepticon, there was no case to compare this mapping to. There are 6 unmapped lists for "basketball" but we run into the same issue, where it is only clear for one of those lists that it's the ball and for one that it is the game and the other four remain ambiguous. I think having two separate concepts and then one shared one for the ambiguous cases seems reasonable overall. The only concern I have is that this could scale quite quickly, since we would need to apply the same split to other polysemous items, which would increase overall granularity.
As for distinguishing American vs. European Football: We don't right now and I don't think there's a good and clean way to do this, since the meaning of this word depends on the speaker's variety. Even for lists that are not English but contain an English gloss, I think we could never be quite certain which of the two is meant. But I'm happy to be convinced otherwise! Did you have a strategy in mind?

MiraAhmedovic: @AnnikaTjuka I'll ping him! :)

LinguList: > Judging from the matches for football, we can say: we have an overarching concept FOOTBALL (GAME OR OBJECT) plus then the FOOTBALL (BALL) and FOOTBALL (GAME) ? It's a bit tedious to map like this, but would avoid the problems.

That's what I said back then.

LinguList: Is there a problem with this suggestion, given that we discussed the ambiguity?

LinguList: We have similar situations in "burn", where we cannot determine often, if it is something that burns or somebody burning something, or with MUROID (MOUSE OR RAT).

MiraAhmedovic: I’ll split them. Depending on the outcome of the American vs. European football question, we should probably also discuss how "soccer" should be mapped (3 lists).

alzkuc: Just to add my two cents: In the datasets that I have previously added, FOOTBALL always links to the ball, not the sport. @MiraAhmedovic

concepticondata/conceptlists/Liu-2011-435.tsv
```diff
@@ -149,7 +149,7 @@ Liu-2011-435-147	147	花	flower	239	FLOWER	1	1	4.12	3.61	4.69	2.49	3.77	0.60	1.5
 Liu-2011-435-148	148	笛子	flute	777	FLUTE	1	2	2.79	3.66	2.94	2.95	2.14	0.25	2.52	0.63	3.81	5.77	0.20	1357
 Liu-2011-435-149	149	苍蝇	fly	1504	FLY (INSECT)	1	2	2.81	3.69	4.03	3.32	2.83	0.63	1.86	0.70	2.44	10.45	0.06	1127
 Liu-2011-435-150	150	脚	foot	1301	FOOT	1	1	3.12	4.20	5.00	2.19	4.32	0.68	1.56	1.00	1.29	2.25	0.00	791
-Liu-2011-435-151	151	橄榄球	football			1	3	2.10	4.00	3.11	2.41	1.97	0.45	0.72	0.50	5.75	10.50	0.31	1374
+Liu-2011-435-151	151	橄榄球	football	4166	FOOTBALL	1	3	2.10	4.00	3.11	2.41	1.97	0.45	0.72	0.50	5.75	10.50	0.31	1374
```
alzkuc: This is, for example, the ball, not the sport. And it refers to the European one.

concepticondata/conceptlists/Nishimoto-2005-359.tsv
```diff
@@ -336,7 +336,7 @@ Nishimoto-2005-359-334	334	Lily			ユリ	YURI	80.0	5.2	4.5	2
 Nishimoto-2005-359-335	335	Pear	922	PEAR	ヨウナシ	YOU-NASHI	87.5	5.5	5.0	4
 Nishimoto-2005-359-336	336	Yo-yo			ヨーヨー	YÔYÔ	80.0	4.0	3.4	4
 Nishimoto-2005-359-337	337	Sailboat	4026	SAILBOAT	ヨット	YOTTO	85.0	4.1	2.9	3
-Nishimoto-2005-359-338	338	Yacht			ヨット	YOTTO	97.5	4.3	3.8	3
+Nishimoto-2005-359-338	338	Yacht	4165	YACHT	ヨット	YOTTO	97.5	4.3	3.8	3
 Nishimoto-2005-359-339	339	Lion	1386	LION	ライオン	RAION	100.0	3.1	3.2	4
 Nishimoto-2005-359-340	340	Camel	1871	CAMEL	ラクダ	RAKUDA	97.5	4.1	2.9	3
 Nishimoto-2005-359-341	341	Football	1845	BALL	ラグビーボール	RAGUBÎ BÔRU	90.0	6.7	3.3	7
```
alzkuc: This is the ball as well, but here it is the rugby one/american football one. (There is a difference, but the dataset lists it as _football_)

concepticondata/conceptlists/Snodgrass-1980-260.tsv
```diff
@@ -93,7 +93,7 @@ Snodgrass-1980-260-91	91	239	FLOWER	Flower
 Snodgrass-1980-260-92	92	777	FLUTE	Flute
 Snodgrass-1980-260-93	93	1504	FLY (INSECT)	Fly
 Snodgrass-1980-260-94	94	1301	FOOT	Foot
-Snodgrass-1980-260-95	95			Football
+Snodgrass-1980-260-95	95	4166	FOOTBALL	Football
```
alzkuc: This is the American football _ball_.

concepticondata/concepticon.tsv
alzkuc: @MiraAhmedovic I have now added comments at the specific datasets where I am sure that they are either listing the ball vs. game and then also american football _ball_ vs. european football _ball_.

alzkuc (changes requested):
A few comments including the discussed FOOTBALL

concepticondata/conceptlists/Dunabeitia-2022-500.tsv
```diff
@@ -275,7 +275,7 @@ Dunabeitia-2022-500-273	273	404	dragonfly	2352	DRAGONFLY
 Dunabeitia-2022-500-274	274	405	anchor	1870	ANCHOR
 Dunabeitia-2022-500-275	275	407	mountain	639	MOUNTAIN
 Dunabeitia-2022-500-276	276	408	finger	1303	FINGER
-Dunabeitia-2022-500-277	277	409	cork		
+Dunabeitia-2022-500-277	277	409	cork	4158	CORK
```
alzkuc, line 278: the new concept CORK relates to agriculture and vegetation. This right here, however, refers to the actual cork in a bottle. I think this should either be part of the description, or disambiguated, if we are to map it as such.

concepticondata/conceptlists/Tsaparina-2011-260.tsv
```diff
@@ -93,7 +93,7 @@ Tsaparina-2011-260-91	91	flower	239	FLOWER	цветок	цветок 	89	4.32	4.
 Tsaparina-2011-260-92	92	flute	777	FLUTE	флейта	флейта	57	4.15	2.50	2.90	3.50
 Tsaparina-2011-260-93	93	fly	1504	FLY (INSECT)	муха	муха	100	4.53	4.33	1.42	4.65
 Tsaparina-2011-260-94	94	foot	1301	FOOT	стопа	стопа	35	4.41	4.61	2.52	4.32
-Tsaparina-2011-260-95	95	football			мяч для регби	мяч для регби 	46	4.71	1.64	3.81	2.88
+Tsaparina-2011-260-95	95	football	4166	FOOTBALL	мяч для регби	мяч для регби 	46	4.71	1.64	3.81	2.88
```
alzkuc: this is again the american (rugby) one. and a ball as an object, not the sport.

alzkuc (changes requested):

concepticondata/conceptlists/Dunabeitia-2022-500.tsv
MiraAhmedovic: I'll refine the description to capture the substance and object meaning. Thank you for pointing this out!

MiraAhmedovic:
I've added the FOOTBALL split and implemented the other changes. The Eurpoean vs. American football debate has been relocated to issue #1525 for the time being. For BASKETBALL, I decided to use the (BALL) concept for object naming studies since that made the most sense to me.
@alzkuc and @FredericBlum, could you have another look at this?

alzkuc:
> I've added the FOOTBALL split and implemented the other changes. The Eurpoean vs. American football debate has been relocated to issue #1525 for the time being. For BASKETBALL, I decided to use the (BALL) concept for object naming studies since that made the most sense to me. @alzkuc and @FredericBlum, could you have another look at this?

Sure, I'll take a look!

concepticondata/concepticon.tsv
```diff
@@ -4130,4 +4130,27 @@ ID	GLOSS	SEMANTICFIELD	DEFINITION	ONTOLOGICAL_CATEGORY	REPLACEMENT_ID
...
+4161	HOSE	Basic actions and technology	A flexible tube used to convey liquids or gases, typically water, from one place to another.	Person/Thing
+4162	SKYSCRAPER	Modern world	A very tall, continuously habitable building with many floors, designed for residential, commercial, or mixed use in urban areas.	Person/Thing
+4163	SLIPPER	Clothing and grooming	A soft, lightweight indoor shoe typically worn for comfort and warmth at home.	Person/Thing
+4164	TUNA	Animals	A large marine fish of the genus Thunnus, known for its streamlined body and importance as a commercial food species.	Person/Thing
+4165	YACHT	Modern world	A recreational boat or vessel, typically medium to large in size, used for leisure cruising or racing on water.	Person/Thing
+4166	FOOTBALL (GAME OR OBJECT)	Modern world	A sport played by two teams in which players move a ball toward a goal to score points, mainly by kicking or carrying it OR the ball used in that sport.	Person/Thing
+4167	FOOTBALL (GAME)	Modern world	A sport played by two teams in which players move a ball toward a goal to score points, mainly by kicking or carrying it depending on the variant.	Person/Thing
```
alzkuc: toward the goal _post_ perhaps?

alzkuc: So, I have taken a look and it seems good to go! Just left one small comment :-) Also, it would be great if you could also include the new concepts in concepticondata/conceptrelations.tsv. E.g. FOOTBALL (OBJECT) as an instance of BALL

alzkuc: (where applicable)

alzkuc (approved):

concepticondata/concepticon.tsv
MiraAhmedovic: I've avoided adding to the conceptrelations file lately since the directionality of `instanceof` or `narrower/wider` is not 100% clear to me. I think ideally this would be handled by somebody who undertands the setup of that file better than I do.
As for the goal post comment, I can do that, it's a good idea. I'll also map any occurences of soccer to 4167 as well.

MiraAhmedovic:
@AnnikaTjuka, would you have a final look at everything?

AnnikaTjuka (approved):
Looks good. Thanks, everyone!

